PR #9, opened 2017-03-05
Add Device cmd_push_constants()

mikedilger:
Double check if the casting is how you want it.  Push constants are untyped (`*const c_void`) so I thought a &T where T is anything made sense.  Since `size` isn't really the size of T (but just the size of the range in T starting at offset) I don't think we can hide those details.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -316,6 +316,21 @@ pub trait DeviceV1_0 {
                                         first_instance);
     }
 
+    unsafe fn cmd_push_constants<T>(&self,
+                                    command_buffer: vk::CommandBuffer,
+                                    layout: vk::PipelineLayout,
+                                    stage_flags: vk::ShaderStageFlags,
+                                    offset: vk::uint32_t,
+                                    size: vk::uint32_t,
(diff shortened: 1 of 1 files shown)


MaikKlein:
Thanks for the PR, I'll have a deeper look at it tomorrow but this should probably use a slice instead of a ptr, offset and size.

```
    unsafe fn cmd_push_constants<T>(&self,
                                    layout: vk::PipelineLayout,
                                    stage_flags: vk::ShaderStageFlags,
                                    slice : &[T]) {
...
}
```

There are a few considerations for size and offset because they need to be a multiple of 4. https://www.khronos.org/registry/vulkan/specs/1.0/man/html/vkCmdPushConstants.html

MaikKlein:
I don't think we can check for the size yet, that would probably require pi types + const fn. I think we should just insert a debug assert.

`debug_assert!(size_of::<T>() % 4 == 0);`

mikedilger:
I'm closing this pull request because mikedilger/master is not necessarily the branch to pull, it's what I use in my project.  I should have forked a branch just for this PR.

PR #11, opened 2017-04-05
Add some commands and fix graphics pipeline creation

### LINKED ISSUE #10: Wrong offsets in bind_vertex_buffers
https://github.com/MaikKlein/ash/blob/master/src/device.rs#L347

Offsets should be a slice not a reference

msiglreith:
- Add `clear_color_image` command
- Add `dispatch` command
- Fix vertex buffer binding, fixes #10 
- Fix graphics pipeline creation for multiple pipelines. (Return pipelines *and* error on failure to recover the correctly built pipelines)

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -286,6 +286,20 @@ pub trait DeviceV1_0 {
         self.fp_v1_0().cmd_bind_index_buffer(command_buffer, buffer, offset, index_type);
     }
 
+    unsafe fn cmd_clear_color_image(&self,
+                                    command_buffer: vk::CommandBuffer,
+                                    image: vk::Image,
+                                    image_layout: vk::ImageLayout,
+                                    clear_color_value: &vk::ClearColorValue,
+                                    ranges: &[vk::ImageSubresourceRange]) {
+        self.fp_v1_0().cmd_clear_color_image(command_buffer,
+                                             image,
+                                             image_layout,
+                                             clear_color_value,
+                                             ranges.len() as vk::uint32_t,
+                                             ranges.as_ptr());
+    }
+
     unsafe fn cmd_draw_indexed(&self,
                                command_buffer: vk::CommandBuffer,
                                index_count: vk::uint32_t,
@@ -343,13 +357,15 @@ pub trait DeviceV1_0 {
 
     unsafe fn cmd_bind_vertex_buffers(&self,
                                       command_buffer: vk::CommandBuffer,
+                                      first_binding: vk::uint32_t,
                                       buffers: &[vk::Buffer],
-                                      offsets: &vk::DeviceSize) {
+                                      offsets: &[vk::DeviceSize]) {
+        debug_assert_eq!(buffers.len(), offsets.len());
         self.fp_v1_0().cmd_bind_vertex_buffers(command_buffer,
-                                               0,
+                                               first_binding,
                                                buffers.len() as vk::uint32_t,
                                                buffers.as_ptr(),
-                                               offsets);
+                                               offsets.as_ptr());
     }
 
     unsafe fn cmd_end_render_pass(&self, command_buffer: vk::CommandBuffer) {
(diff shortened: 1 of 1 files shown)


MaikKlein:
Looks good, thanks for the PR. It seems that the examples need to be updated. I can do that later today.

src/device.rs
```diff
@@ -343,13 +357,15 @@ pub trait DeviceV1_0 {
...
     unsafe fn cmd_bind_vertex_buffers(&self,
                                       command_buffer: vk::CommandBuffer,
+                                      first_binding: vk::uint32_t,
                                       buffers: &[vk::Buffer],
-                                      offsets: &vk::DeviceSize) {
+                                      offsets: &[vk::DeviceSize]) {
+        debug_assert_eq!(buffers.len(), offsets.len());
         self.fp_v1_0().cmd_bind_vertex_buffers(command_buffer,
-                                               0,
+                                               first_binding,
```
MaikKlein: Is this necessary? We already use a slice here which already has the information where it should start. For example

`device.cmd_bind_vertex_buffers(cb, &buffers[5 .. 10], &offsets);`

src/device.rs
```diff
@@ -409,8 +436,8 @@ pub trait DeviceV1_0 {
                                        pipelines.as_mut_ptr());
         pipelines.set_len(create_infos.len());
         match err_code {
-            vk::Result::Success => Ok(pipelines),
-            _ => Err(err_code),
+            vk::Result::Success => (pipelines, Ok(())),
+            _ => (pipelines, Err(err_code)),
```
MaikKlein: I am a bit confused after reading https://www.khronos.org/registry/vulkan/specs/1.0/man/html/vkCreateGraphicsPipelines.html

It seems that the function can only fail if the host or the device is oom. How can you recover pipelines that failed to build? I assumed that no pipeline would be returned if an error occurred.

src/device.rs
msiglreith: This should be necessary imo if you only want to update a part of the vertex input bindings,
`first_binding` states the first vertex input binding offset, else we would have to set it to 0 and have to update all input bindings.

Corresponding specs: (https://www.khronos.org/registry/vulkan/specs/1.0/man/html/vkCmdBindVertexBuffers.html)

> The values taken from elements i of pBuffers and pOffsets replace the current state for the vertex input binding firstBinding + i, for i in [0, bindingCount).

src/device.rs
msiglreith: Good point there. After reading the linked page I would agree with you.
I mostly followed the `Multiple Pipeline Creation` section.
> If any pipeline fails creation (for example, due
to out of memory errors), the vkCreate*Pipelines commands will return an error code. The
implementation will attempt to create all pipelines, and only return VK_NULL_HANDLE values for
those that actually failed.

The specification isn't really clear here I think.

Therefore, I tried a bit around to get one of these mysterious `VK_NULL_HANDLES`. Setting `s_type` in the `GraphicsPipelineCreateInfo` to `vk::StructureType::ComputePipelineCreateInfo` results in a `Err(ErrorInitializationFailed)` on my AMD system, which isn't listed as a return code!
Creating two pipelines (one of them failed) gave me this as return code:
`([0x0, 0x1edbeea5830], Err(ErrorInitializationFailed))`
So one null handle, a valid pipeline handle and an error.

I'm not sure if this is correct behavior from the driver or the specs are missing something, but imo the proposed code change handles this scenario more gracefully.

msiglreith:
Sorry, for missing to update the examples. I can also take care of it if you want (:

MaikKlein:
I already did it, but I didn't push it yet. I'll push it tomorrow.

src/device.rs
MaikKlein: I see, I guess that makes sense.

src/device.rs
MaikKlein: You are right, I actually misunderstood the purpose of firstBinding.

src/device.rs
MaikKlein: I think i might slightly adjust the return type to `Result<Vec<vk::Pipeline>, (Vec<vk::Pipeline>, vk::Result)>`.  This guarantees that on success that there will not be any null handles inside the vec, and you can still recover if you need too. Any objections?

msiglreith: Sounds good! 👍

PR #17, opened 2017-04-25
Expose InstanceError

cjhowe:
Shouldn't this be exposed publicly, since it's used in the public `Entry::create_instance`?


MaikKlein:
It should, thanks for the PR.

PR #31, opened 2017-09-29
[WIP] Copy implementations

kvark:
This would help us with https://github.com/gfx-rs/gfx/issues/1226
If you find this too intrusive, we can make a `copy` feature for those extra derives.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/31)

### DIFF
--- a/ash/Cargo.toml
+++ b/ash/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ash"
-version = "0.18.5"
+version = "0.18.6"
 authors = ["maik klein <maikklein@googlemail.com>"]
 description = "Vulkan bindings for Rust"
 license = "MIT"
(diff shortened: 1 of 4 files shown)


kvark:
@MaikKlein if that's indeed so heavy, we'll just box the device then.
`Copy` on the extents would still be useful, but it's not critical.

kvark:
Actually, the problem is not just for storing `VkDevice` in the portability layer. The problem is that we have multiple things actually using a device:
  - `Device`
  - `CommandBuffer`
  - `CommandQueue`
  - `DescriptorPool`
  - `CommandPool`

So, naively, making a device non-copyable means I can't make those copyable.
A workaround could be to store the device by a raw mutable pointer in those dependent objects... eww

kvark:
> A workaround could be to store the device by a raw mutable pointer in those dependent objects... eww

I don't see another way. This is implemented now, closing.

PR #36, opened 2017-10-14
Add MoltenVK support on macOS/iOS

davll:
It will enable Vulkan availability on macOS/iOS with MoltenVK installed

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/36)


MaikKlein:
Thanks, this is great :).

PR #40, opened 2017-12-02
Fix push constant size (length of bytes instead of DWORDs)

msiglreith:
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/40)

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -667,14 +667,14 @@ pub trait DeviceV1_0 {
         layout: vk::PipelineLayout,
         stage_flags: vk::ShaderStageFlags,
         offset: vk::uint32_t,
-        constants: &[u32]
+        constants: &[u8]
     ) {
         self.fp_v1_0().cmd_push_constants(
             command_buffer,
             layout,
             stage_flags,
             offset,
-            constants.len() as vk::uint32_t,
+            constants.len() as _,
             constants.as_ptr() as _,
         );
     }


kocsis1david:
It would be more consistent to always use `&[u8]` where binary data is expected. I would use u8 when the parameter name is size and if it's something else, like count or number_of, then maybe you could use something else.

msiglreith:
Sorry, I forgot that this PR is still open!
The motivation for `u32` vs `u8` is that the number of bytes pushed **must** be a multiple of 4bytes. Therefore, we can directly enforce this requirement at API boundary via `u32`.

MaikKlein:
@msiglreith Sorry for taking so long for a simple change.

Okay I just had a look at it and I think that `u8` is probably still better.

```
use std::mem::{size_of, zeroed};
use std::slice::from_raw_parts;
type Matrix = [f32; 16];

fn main() {
    let matrix: Matrix = unsafe { zeroed() };

    let u32_slice: &[u32] = unsafe {
        from_raw_parts::<u32>(
            matrix.as_ptr() as *const u32,
            size_of::<Matrix>() / size_of::<u32>(),
        )
    };
    
    let u8_slice: &[u8] = unsafe {
        from_raw_parts::<u8>(
            matrix.as_ptr() as *const u8,
            size_of::<Matrix>(),
        )
    };
    
    println!("{:?}", u32_slice.len());
    println!("{:?}", u8_slice.len());
}
```

This is probably how it ends up being used. In the end it probably doesn't matter but the `u8` variant  might be a bit safer as Vulkan will throw a debug error if the size is not a multiple of 4, where the `u32` variant would result in a more subtle bug. Of course this is probably checked with `assert!(size_of::<T> % 4 == 0)` or something similar.

I would also think that in this case it might be better to not use a slice at all, and pass the pointer + size directly, but again it probably doesn't matter. I assume DX12 also uses a pointer + offset?

msiglreith:
I'm fine with changing it to `u8`. Atm it's a bit awkward as `offset` is offset in bytes.

kvark:
> the u8 variant might be a bit safer as Vulkan will throw a debug error if the size is not a multiple of 4

This contradicts one of the basics principles we've been doing Rust ecosystem development: if a precondition is not costy to lift to a type-system level, it definitely should be lifted. `&[u32]` is a clear API on the type level.

> I would also think that in this case it might be better to not use a slice at all, and pass the pointer + size directly, but again it probably doesn't matter. I assume DX12 also uses a pointer + size?

This seems like going even farther from the type-level enforcement.

MaikKlein:
>This contradicts one of the basics principles we've been doing Rust ecosystem development: if a precondition is not costy to lift to a type-system level, it definitely should be lifted. &[u32] is a clear API on the type level.

I completely understand what you are saying, but in the end you still have to check for the correct alignment in all cases and with `&[u8]` you at least get the chance of triggering the debug layer.

I just have to be consistent, if I would use `&[u32]`, then I guess the offset shouldn't be in bytes anymore, or completely remove the offset and retrieve the offset from the slice itself, otherwise it would be inconsistent.

kvark:
@MaikKlein ok, I suppose it does bring more trouble than the benefit then.

msiglreith:
The consensus is pass data as `&[u8]` and offset in bytes, isn't it?

MaikKlein:
@msiglreith Yeah that should be fine. Although do you think it would be nicer to compute the offset directly from the slice? Actually I think I have done that before, https://github.com/MaikKlein/ash/blob/master/ash/src/device.rs#L808-L819.

msiglreith:
@MaikKlein I don't see how this would work. In the linked example, the offset is hardcoded to 0, limiting the configuration possbilities actually.

MaikKlein:
@msiglreith The idea was to use slicing, like `cmd_set_viewport(cb, &viewports[10..])`, Then the base address of the slice is essentially the base address + offset. I wasn't sure if that would be better than passing the offset explicitly.

msiglreith:
@MaikKlein But this doesn't work. Offset specifies the beginning in the array of viewports stored inside the driver. The current implementation doesn't allow to modify a slice in the driver array with an offset != 0.

For example
```
driver side: [Viewport0, Viewport1, Viewport2, Viewport3]
user side: update range: 1..3
   -> cmd_set_viewport(cb, 1, &[Viewport1, Viewport2]) // without offset we would modify 0 and 1
```

PR #43, opened 2018-01-30
Nicer Entry implementation

kvark:
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/43)


MaikKlein:
Wow that code was terrible, thanks for cleaning it up.

PR #77, opened 2018-07-10
[WIP] Adds support for using custom ways to load the vulkan dynamic libraries.

ZeGentzy:
These changes are backwards compatible with older versions.
Signed-off-by: Hal Gentz <zegentzy@protonmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/77)

### DIFF
--- a/ash/Cargo.toml
+++ b/ash/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ash"
-version = "0.24.3"
+version = "0.24.4"
 authors = ["maik klein <maikklein@googlemail.com>"]
 description = "Vulkan bindings for Rust"
 license = "MIT"
@@ -10,7 +10,7 @@ keywords = ["vulkan", "graphic"]
 documentation = "https://docs.rs/ash"
 
 [dependencies]
-shared_library = "0.1.5"
+shared_library = "0.1.9"
 lazy_static = "0.2.1"
 libc = "0.2.26"
 
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -9,6 +9,8 @@ use std::fmt;
 use std::path::Path;
 use RawPtr;
 use version::{EntryLoader, FunctionPointers, InstanceLoader, V1_0};
+use std::sync::Arc;
+use libc;
 
 #[cfg(windows)]
 const LIB_PATH: &'static str = "vulkan-1.dll";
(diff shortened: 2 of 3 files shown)


ash/Cargo.toml
```diff
@@ -10,7 +10,8 @@ keywords = ["vulkan", "graphic"]
 documentation = "https://docs.rs/ash"
 
 [dependencies]
-shared_library = "0.1.5"
+#shared_library = "0.1.9"
```
ZeGentzy: Will fix when 0.1.9 is out.

MaikKlein: What is the plan here? Because this library has been yanked from crates.io https://crates.io/crates/shared_library Will this be published again?

MaikKlein: Ah I see the "newer" versions have a lower version which messes up crates.io.

ZeGentzy:
@MaikKlein 0.1.9 is out now, should be good for review/merging.

ash/src/entry.rs
```diff
@@ -153,24 +153,40 @@ impl EntryV1_0 for Entry<V1_0> {
...
-            .map_err(|err| LoadingError::LibraryLoadError(err.clone()))?;
-
-        let static_fn = vk::StaticFn::load(|name| unsafe {
-            lib.symbol(&*name.to_string_lossy())
-                .unwrap_or(ptr::null_mut())
-        }).map_err(LoadingError::StaticLoadError)?;
+impl<V: FunctionPointers, L> EntryCustom<V, L> {
+    pub fn new_custom<OF, LF>(o: OF, mut l: LF) -> Result<Self, LoadingError>
+    where
+        OF: FnOnce() -> Result<L, LoadingError>,
```
MaikKlein: Could you give the generics more explicit names? Maybe something like `Open` and `Load` instead of `OF`and `LF`

MaikKlein:
Everything else looks good, thanks for the PR. Could you also quickly explain the use case for the portability lib? I assume you are going to create a custom entry?

ZeGentzy:
@MaikKlein One example is using `RTLD_NEXT` or `RTLD_DEFAULT`: https://github.com/gfx-rs/gfx/pull/2221/files#diff-44198a1b71ecb41e695db8b6967cd689R90

For why someone might want to do this, I refer you to this wonderful gem I found: https://github.com/amonakov/on-wrapping/blob/master/interposers-discussion.asciidoc

ZeGentzy:
@MaikKlein Can you release 0.24.4 on crates.io?

PR #81, opened 2018-07-28
Implementation of Default trait

MatusT:
This pull request implements Default trait as discussed in https://github.com/MaikKlein/ash/issues/27. It mimics the C++ wrapper behavior in that the s_type field is always filled correctly and remaining fields are either nulled or properly defaulted (for example: unions are set to their first option).

Further, the generate_struct function is reorganized such that the Rusts derive macro is used if possible for any trait.
 
Currently, this PR will not compile until the full support for extensions lands. All 187 errors are related to missing values in StructureType.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/81)

### DIFF
--- a/ash/src/version.rs
+++ b/ash/src/version.rs
@@ -1,6 +1,6 @@
-pub use device::DeviceV1_0;
+pub use device::{DeviceV1_0, DeviceV1_1};
 pub use entry::EntryV1_0;
-pub use instance::InstanceV1_0;
+pub use instance::{InstanceV1_0, InstanceV1_1};
 use std::mem;
 use vk;
 pub trait FunctionPointers {


MaikKlein:
> It mimics the C++ wrapper behavior in that the s_type field is always filled correctly and remaining fields are either nulled or properly defaulted (for example: unions are set to their first option).

Are you sure unions are filled with the first value? That seems a bit strange, I would have expected it to be `0 ` all the time. Which C++ wrapper are you talking about, VulkanHpp?

I currently put this PR on hold because I want to implement all the extension constants first. Then I'll have a look if we can automatically detect which structs should implement copy, clone, debug, default etc and how this PR can be integrated.

Cyres:
No, Unions are **not** filled with the first value, they are zeroed.

MatusT:
> No, Unions are not filled with the first value, they are zeroed.

Sorry, my bad. Anyway, There seems to be only 2 unions and both are zeroed when their first option is chosen as Default.

```rust
#[repr(C)]
#[derive(Copy, Clone)]
pub union ClearColorValue {
    pub float32: [c_float; 4],
    pub int32: [int32_t; 4],
    pub uint32: [uint32_t; 4],
}
impl Default for ClearColorValue {
    fn default() -> ClearColorValue {
        ClearColorValue { float32: [0.0; 4] }
    }
}
#[repr(C)]
#[derive(Copy, Clone, Default)]
pub struct ClearDepthStencilValue {
    pub depth: c_float,
    pub stencil: uint32_t,
}
#[repr(C)]
#[derive(Copy, Clone)]
pub union ClearValue {
    pub color: ClearColorValue,
    pub depth_stencil: ClearDepthStencilValue,
}
impl Default for ClearValue {
    fn default() -> ClearValue {
        ClearValue {
            color: ClearColorValue::default(),
        }
    }
}
```

I guess, for now "choosing the first option" <=> "zeroed". However, maybe Vulkan may add something in future that could break it? I am open to suggestions what to do here as I am not sure.

Another thing I was thinking about that you might see as an issue: currently, Default trait is aggressively derived for every struct, even those that are only a return value from Vulkan' s functions, such as *PhysicalDeviceProperties*, and so there is no reason for them to have the trait. What do you think should be done? 
1. Let It be
2. Build a set of structs returned by Vulkan's functions and omit them in generation

MaikKlein:
Oh sorry I haven't look to closely at the union implementation. I actually don't know what the best possible implementation would be. I think we should use [zeroed](https://doc.rust-lang.org/std/mem/fn.zeroed.html).

I have no problem with extra derives for structs. It is only compiled once and doesn't effect incremental builds and it will also be stripped by the collector anyway.

I still would like to avoid manually specifying those types, that just seems like a maintenance nightmare. I have now mostly finished the essential part of the generator and I can now have a look at the derives.

MaikKlein:
@MatusT https://github.com/MaikKlein/ash/issues/52#issuecomment-408965884 I went ahead and quickly implemented it myself. The changes are already live if you want to have a look at the output, maybe you can spot some mistakes.

MatusT:
@MaikKlein Amazing! The only thing I would like to retain from my original code is the one else if branch to capture pointers and use ::std::ptr::null() instead of zeroed. This is now the only diff in my PR. I think that now you can either merge or close this PR.

Thank you :)

MaikKlein:
Okay nice, I'll look at it tomorrow :).

MaikKlein:
Can you squash the commits and regenerate the vk.rs to make sure everything still works?

PR #87, opened 2018-08-01
Add new functions from 1.1

Cyres:
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/87)


Cyres:
@MaikKlein This should be ready to be merged. However you may take a look, if `unsafe` is used correctly. All Device functions are unsafe. Also, all Instance function, where a non-exhaustive type is passed (like indices or info-structs), are marked unsafe.

MaikKlein:
Because of the recent implementation of the default trait, I am inclined to believe that everything might need to become unsafe, but I'll open a separate issue for that and merge this PR now.

PR #88, opened 2018-08-02
Handle trait, VK_EXT_debug_utils

Ralith:
This implements the abstractions discussed in https://github.com/MaikKlein/ash/issues/70 to allow conversion of handles to/from `u64` and easy access of handle types.

I ran into a few interesting questions:
- Should the names of methods on extension objects (and even fields on their pointer tables) include the extension suffix? They're already disambiguated by being on the extension object, and a similar argument was used to remove suffixes from constants.
- When a rustier but less trivial interface can be written for a function (compare `set_object_name` vs `debug_utils_set_object_name_ext` here), should it exist alongside the naive wrapping, or replace it outright? If they both exist, what naming convention should be used?

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/88)

### DIFF
--- a/ash/src/extensions/debug_utils.rs
+++ b/ash/src/extensions/debug_utils.rs
@@ -1,0 +1,145 @@
+#![allow(dead_code)]
+use prelude::*;
+use std::ffi::CStr;
+use std::mem;
+use version::{DeviceV1_0, InstanceV1_0};
+use {vk, RawPtr};
+
+#[derive(Clone)]
+pub struct DebugUtils {
+    debug_utils_fn: vk::ExtDebugUtilsFn,
+}
+
+impl DebugUtils {
+    pub fn new<I: InstanceV1_0, D: DeviceV1_0>(
+        instance: &I,
+        device: &D,
+    ) -> Result<DebugUtils, Vec<&'static str>> {
+        let debug_utils_fn = vk::ExtDebugUtilsFn::load(|name| unsafe {
+            mem::transmute(instance.get_device_proc_addr(device.handle(), name.as_ptr()))
+        })?;
+        Ok(DebugUtils {
+            debug_utils_fn: debug_utils_fn,
+        })
+    }
+
+    pub fn name() -> &'static CStr {
+        CStr::from_bytes_with_nul(b"VK_EXT_debug_utils\0").expect("Wrong extension string")
+    }
+
+    pub unsafe fn debug_utils_set_object_name_ext(
+        &self,
+        device: vk::Device,
+        name_info: &vk::DebugUtilsObjectNameInfoEXT,
+    ) -> VkResult<()> {
+        let err_code = self.debug_utils_fn.set_debug_utils_object_name_ext(device, name_info);
+        match err_code {
+            vk::Result::SUCCESS => Ok(()),
+            _ => Err(err_code),
+        }
+    }
+
+    pub unsafe fn debug_utils_set_object_tag_ext(
+        &self,
+        device: vk::Device,
+        tag_info: &vk::DebugUtilsObjectTagInfoEXT,
+    ) -> VkResult<()> {
+        let err_code = self.debug_utils_fn.set_debug_utils_object_tag_ext(device, tag_info);
+        match err_code {
+            vk::Result::SUCCESS => Ok(()),
+            _ => Err(err_code),
+        }
+    }
+
+    pub unsafe fn cmd_begin_debug_utils_label_ext(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        label: &vk::DebugUtilsLabelEXT,
+    ) {
+        self.debug_utils_fn
+            .cmd_begin_debug_utils_label_ext(command_buffer, label);
+    }
+
+    pub unsafe fn cmd_end_debug_utils_label_ext(&self, command_buffer: vk::CommandBuffer) {
+        self.debug_utils_fn
+            .cmd_end_debug_utils_label_ext(command_buffer);
+    }
+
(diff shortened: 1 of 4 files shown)


MaikKlein:
I'll update rust, we need the `transparent` feature anyway.

Ralith:
I'm actually tempted to take the high-level spirit a further still, and replace all occurrences of `&vk::DebugUtilsLabelEXT` with `name: &CStr, color: [f32; 4]`. However, this has the downside of preventing seamless future idiomatic use of the `p_next` member of `DebugUtilsLabelEXT`. Do we care about that? If so, the generic name/tag API should be removed for the same reason, and perhaps relegated to a helper function somewhere.

My personal feeling is that convenient zero-cost interfaces are great to have even if may not account for every possible future use case; we can always add new functions, though we might need to get creative with naming them. If the goal is to eventually expose totally un-wrapped function pointers alongside, perhaps that's enough?

ash/src/extensions/debug_utils.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    pub fn name() -> &'static CStr {
+        CStr::from_bytes_with_nul(b"VK_EXT_debug_utils\0").expect("Wrong extension string")
+    }
+
+    pub unsafe fn debug_utils_set_object_name_ext<H: vk::Handle>(
+        &self,
+        device: vk::Device,
+        object: H,
+        name: &CStr,
```
MaikKlein: I think it might be more ergonomic if this would be a `name: &str` and we create the `CStr` inside the function.

Cyres: I'd go with `name: &str`. The function is not performance critical.

Ralith: That would be inconsistent with other parts of the API (e.g. extension names, layers) and require allocation, making the interface no longer strictly zero-cost, though the cost is probably not important. Are we sure that's desired?

I use a macro like this to make `CStr`s convenient to come up with:
```rust
#[macro_export]
macro_rules! cstr {
    ($str:expr) => { #[allow(unused_unsafe)] unsafe { ::std::ffi::CStr::from_bytes_with_nul_unchecked(concat!($str, "\0").as_bytes()) } }
}
```

MaikKlein: Ah right I didn't think about the null terminator, yeah `CStr` is probably fine.

MaikKlein:
>I'm actually tempted to take the high-level spirit a further still, and replace all occurrences of &vk::DebugUtilsLabelEXT with name: &CStr, color: [f32; 4]. However, this has the downside of preventing seamless future idiomatic use of the p_next member of DebugUtilsLabelEXT. Do we care about that? If so, the generic name/tag API should be removed for the same reason, and perhaps relegated to a helper function somewhere.

But does it add much? It seems easy to write something on top of it, which you probably need to do anyway.

Everything looks great, I also now finally understand why you wanted the `Handle` trait.

Ralith:
> But does it add much?

It means you could write

```rust
debug.queue_insert_debug_utils_label_ext(queue, cstr!("foo"), [0.0; 4]);
```

instead of

```rust
debug.queue_insert_debug_utils_label_ext(queue, &vk::DebugUtilsLabelEXT {
    p_label_name: cstr!("foo"),
    ..Default::default()
});
```

I'm ambivalent about whether that's worth it.

Ralith:
Rebased.

So, the real thing that gives me pause isn't the relatively trivial dichotomy I posed above, but rather the inconsistency between the concise abstract interface exposed by `fn debug_utils_set_object_name_ext<H: vk::Handle>` versus the more strictly authentic version used elsewhere. It feels weird to maximize ergonomics for one function but err on the side of authenticity in the next. If the simplified label interface is best left to downstream code, can't the same be said of the simplified object naming interface?

MaikKlein:
I always chose the highest level of abstraction without sacrificing flexibility before. `Handle` seems to be a great addition.

Leaving pnext out would force the user to needlessly resort to the raw function, and I haven't done this before. You could always implement something like `From<(CStr, [f32;4])>`. But I think this is best left to the user.

I would merge the PR as it is.

Ralith:
> Do you think there can be multiple `ObjectType` per handle in the future?

No, this should always be 1:1.

> Leaving pnext out would force the user to needlessly resort to the raw function, and I haven't done this before

The current implementation of `debug_utils_set_object_name_ext` and `debug_utils_set_object_tag_ext` hides `p_next` in a similar way. Should I back that out? The trait makes it trivial for people to write the wrapper themselves, though it seems a little silly to have everyone reimplementing the same code. Maybe we could explore a common module or crate of helper functions in the future.

MaikKlein:
Ah I didn't look to closely at those structs. That is a bit unfortunate. I am honestly not sure what we should do, I kind of like the way it automatically sets the `ObjectType`, seems less error prone that way. It is also pretty intuitive. 

But it would be unfortunate if you write code today and then needing pnext in the future and you would need to refactor everything to raw function pointers. 

What do you think we should do?

> Maybe we could explore a common module or crate of helper functions in the future.

I think that definitely would be helpful.

Ralith:
I think I'm most comfortable being strictly authentic (i.e. preserve maximum expressiveness) with the wrappers. The trait's still useful, both as a building block and for access to the inner u64. 

In my project I have my own layer of wrapping on top of the debug marker extension anyway, since it's only available with the layers enabled and hence calls for some magic conditional stuff, so it comes out about the same for me anyway.

I think this is ready now. We can explore ways to gracefully expose lossy helpers in future work.

PR #93, opened 2018-08-03
Move `vk.rs` to new vulkan-sys crate

### LINKED ISSUE #91: Split into ash and ash-sys
It would be very much more clean to split up ash into ash and ash-sys, where ash-sys only contains the generated files from the generator: https://github.com/MaikKlein/ash/issues/52#issuecomment-402036590

I'd do a PR when other PRs which affects *vk.rs* has been merged.

TimDiekmann:
Closes #91

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/93)


MaikKlein:
Do we want `vk-sys` to start from version `0.24.3`. Shouldn't the version be separate from ash?

Cyres:
What's about the current header version, which is used for generating *vk.rs*?

Ralith:
Changes to the generator will inevitably force the version number to diverge from upstream's sooner or later.

On another subject, I wonder if it would be appropriate to grab a more generic name like `vulkan-sys`?

Cyres:
`vulkan-sys` isn't used atm. For users who don't use ash but raw vulkan, it doesn't lead to confusion.

MaikKlein:
Sorry I meant `ash-sys` above, I think `ash-sys` is fine but I really don't have any strong opinions about the name. 

I still think the version should be separate and that means the workspace dependency should also require an explicit version.

Ralith:
I really don't think using the vulkan.h version makes sense, unless we want to commit to never, ever making another breaking change to the structure of the output.

Cyres:
> I really don't think using the vulkan.h version makes sense, unless we want to commit to never, ever making another breaking change to the structure of the output.

I think @MaikKlein has to decide this.

# Pro:
- It's very transparent, which specs can be used
- breaking change will be rarely when generator will be merged to master, `vulkan-sys` won't be published until then (I guess?)

# Con:
- If the time span between two spec updates is long, a critical update may take longer

Ralith:
> It's very transparent, which specs can be used

It's perfectly legal to use a newer binding with an older vulkan implementation, and it will be immediately obvious (and easily corrected) if the binding doesn't have something you need. This also breaks down permanently the first time a breaking generator change happens.

> breaking change will be rarely when generator will be merged to master

However rare breaking changes to the behavior of the generator are, they will almost certainly happen inevitably, rendering the whole endeavor futile.

Ralith:
I'm a bit fuzzy on the motivation too. It *would* be aesthetically pleasing for the various vulkan crates to agree on a common -sys layer (and occasionally useful for type-safe interop, e.g. in windowing and VR libraries), but perhaps we should get buy-in from the other maintainers before investing effort in it.

Cyres:
Rerenamed to `ash-sys` and changed version to `0.1`. Now it's up to @MaikKlein to merge or close it.

kvark:
In general, gfx-backend-vulkan would like to use a lower level crate to do things.
We'd also want to ask @tomaka if they'd consider moving Vulkano to ash-sys, if it's superior to the current vk-sys.

MaikKlein:
`ash-sys` is definitely a bit opinionated and is just an implementation detail for ash. `vk-sys` is very similar to `ash-sys`. Ash just creates separate function pointer structs for the various extensions. Also the naming scheme is different.

The biggest benefit is that `ash-sys` is generated and therefore will include all the new stuff. `ash-sys` also has real bitflags, more derives (also not complete), most things also have a default implementation.

const names are currently still missing but will be implemented very soon and the function pointer loading will become similar to vulkano, where null pointers will resolve into a function that always panics. At the moment ash just panics if anything fails to load.

I'll merge this PR before the next version of `ash` is released.

tomaka:
> also has real bitflags, more derives (also not complete)

That's really opinionated, but I don't really like that.
The way I see FFI bindings is that they should just be function pointers and constants, and that's it. Any trait implementation for example makes the bindings "unpure", as it can possibly lead to a disagreements over how the trait implementation should behave.

However that's not a strong enough argument to *not* switch vulkano to ash-sys, I guess.

MaikKlein:
> @tomaka's complaints could be addressed by generating two separate files: one for the sys crate which contains nothing but type and constant definitions, and another that implements the idiomatic rust wrappers, strong types, loading, etc. and goes in the ash crate.

I am okay with this. We could generally try to make everything more configurable, for example allow different naming schemes etc

Although this probably will require some refactoring. 

My only concern is that we might end up with too many versions of ash-sys, which might not be compatible with each other.

Thomspoon:
Has there been any progress on this? As a new customer of Vulkan, it's hard to jump into an API that abstracts out so much. It'd be nice to have a raw Vulkan API that doesn't require reading complex macros or building a bunch of wrapper structs.

Ralith:
I don't think ash's API requires you to use macros or construct wrapper structs at any point.

Thomspoon:
I'm talking about using raw vulkan generator. If you want to use raw vk calls, you can follow suit to what vulkano and ash are doing, but I think it's a little complicated than just loading the library, constructing the pointer and being good to go. I don't really want to start with Ash or Vulkano, I'd rather see the footguns for myself.

Ralith:
To be clear, ash is nothing like vulkano; there is practically no abstraction, just trivial (but convenient) syntax sugar. It's almost identical to what you'd write in C.

Thomspoon:
Gotcha, maybe I'm worrying about nothing. I get antsy when I see issues like: #100, where it looks like an abstraction that won't let you use the raw functionality of Vulkan. Thank you for the responses.

Ralith:
Yeah, issues like that are rare, and are considered to be bugs.

PR #108, opened 2018-08-22
Make `enumerate_instance_version` an `Entry` function

### LINKED ISSUE #103: enumerate_instance_version should be available before creating an instance
According to the [docs](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkEnumerateInstanceVersion.html), `EnumerateInstanceVersion` is meant to

> Query instance-level version **before** instance creation

(emphasis mine).

However, it is currently listed on the `InstanceFnV1_1` struct, and it requires the creation of an instance before it can be called. It should be available in the `EntryFn` struct.

GabrielMajeri:
`enumerate_instance_version` is a function meant to be called before instance creation. This pull request moves its function pointer to `Entry`.

~~Also adds the `EntryV1_1` trait. I have not added any other functions to the trait yet, so as per [this comment](https://github.com/MaikKlein/ash/issues/52#issuecomment-412281719) it is not currently possible to create a 1.1 instance safely.~~
Nevermind, I tested this and it works. The issue mentioned in the comment is fixed now. Ash now basically has Vulkan 1.1 support, although there still are some issues remaining.

**Note**: ~~do we want this function to be marked as `unsafe`? It seems pretty safe to me.~~
It's been made safe.

Fixes #103.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/108)

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -170,3 +170,17 @@ impl<V: FunctionPointers> Entry<V> {
         })
     }
 }
+
+#[allow(non_camel_case_types)]
+pub trait EntryV1_1: EntryV1_0 {
+    fn fp_v1_1(&self) -> &vk::EntryFnV1_1;
+
+    unsafe fn enumerate_instance_version(&self) -> VkResult<vk::uint32_t> {
+        let mut api_version = mem::uninitialized();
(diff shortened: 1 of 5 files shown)


MaikKlein:
This function should be safe.

Issue #109, opened 2018-08-22
Vulkan 1.1 getter functions don't work


GabrielMajeri:
The function `get_physical_device_properties2` (and possibly others) don't work. To call this function, we have to initialize the return structure's `s_type` and `p_next`, leave the rest uninitialized, then call the function. Currently, the [code uses mem::uninitialized()](https://github.com/MaikKlein/ash/blob/3e812896ff0aa8c170feb07ba07c928bb3393c96/ash/src/instance.rs#L125) for filling the whole structure.

<hr/>

This brings up a new issue: how to support extensions in these getters? We have to initialize every extension's structure in the chain, then call the function. If I want to get, for example, `PhysicalDeviceIDProperties`, I have to initialize the first structure (`VkPhysicalDeviceProperties2`), then the `PhysicalDeviceIDProperties` structure.

It seems there is no way for us to safely wrap Vulkan getter functions. We need to allow the user to create the chain of structures themselves, and we should only take in a `&mut` to the first structure in the chain.

Ralith:
I would expect the affected functions are all which accept pointers to structs as output arguments where the structs contain `pNext` members. "returned" here almost certainly does not mean that `pNext` will never be used.

omni-viral:
Or if it contains `sType` member.

MaikKlein:
Fixed by https://github.com/MaikKlein/ash/pull/133 feel free to reopen the issue if there are more special cases

PR #111, opened 2018-08-23
Fix instance getters to support extensions

GabrielMajeri:
As discussed in #109, the various getter functions for physical device don't work, because the user isn't able to pass in a chain of mutable structures for the Vulkan driver to fill in.

- All getters will probably have to be marked `unsafe`, since we cannot validate the user passed in a structure with a valid `p_next` pointer.

- I have not touched the getters which don't have any extensions yet. For example, I haven't modified `get_physical_device_external_fence_properties` since there are no published extensions which affect it (yet).

- I don't know how to solve the issue with functions which need to work with vectors, specifically `get_physical_device_queue_family_properties2`, which does have an extension structure for it (`QueueFamilyCheckpointPropertiesNV`).
  - We would have to provide a function to get the length of the returned array.
The user would then allocate an array of structures, create an array of extension structures, then properly set the `p_next` pointers, and call the function to fill in the array of structure chains.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/111)

### DIFF
--- a/ash/src/instance.rs
+++ b/ash/src/instance.rs
@@ -151,10 +151,10 @@ pub trait InstanceV1_1: InstanceV1_0 {
         }
     }
 
-    fn get_physical_device_queue_family_properties2(
+    fn get_physical_device_queue_family_properties2_len(
         &self,
         physical_device: vk::PhysicalDevice,
-    ) -> Vec<vk::QueueFamilyProperties2> {
+    ) -> usize {
         unsafe {
             let mut queue_count = 0;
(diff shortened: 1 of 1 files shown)


GabrielMajeri:
@Ralith ok, I've separated those functions into two, one with the `_len` prefix which retrieves the length of the array, and the other performs the actual call.

Alternatively, we could have a single function, which takes in the return parameter as an `Option`.

Ralith:
I was imagining the latter option, but on review this seems nicer anyway.

PR #114, opened 2018-08-25
Use Rust integral types

GabrielMajeri:
I tried using the `vk_version_{*}` macros, but they didn't work: 
```
cannot find type `uint32_t` in this scope`
```

This pull request fixes them.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/114)

### DIFF
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -19,19 +19,19 @@ macro_rules! vk_make_version {
 #[macro_export]
 macro_rules! vk_version_major {
     ($major:expr) => {
-        ($major as uint32_t) >> 22
+        ($major as u32) >> 22
     };
 }
 #[macro_export]
 macro_rules! vk_version_minor {
     ($minor:expr) => {
-        (($minor as uint32_t) >> 12) & 0x3ff
+        (($minor as u32) >> 12) & 0x3ff
     };
 }
 #[macro_export]
 macro_rules! vk_version_patch {
     ($minor:expr) => {
-        ($minor as uint32_t) & 0xfff
+        ($minor as u32) & 0xfff
     };
 }
 pub type RROutput = c_ulong;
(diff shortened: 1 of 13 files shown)


ash/src/vk.rs
```diff
@@ -19,19 +19,19 @@ macro_rules! vk_make_version {
 #[macro_export]
 macro_rules! vk_version_major {
     ($major:expr) => {
-        ($major as uint32_t) >> 22
+        ($major as u32) >> 22
```
MaikKlein: I think we should use something like `crate::ash::vk::uint32_t`, but I don't think that quite works if the crate renames ash. I think there was a correct way but I forgot how to do it as I don't write macros that often.

GabrielMajeri: I think will work `$crate::vk::uint32_t`. But before I make this change, any reason why we don't use Rust's own `u32`? It's statically guaranteed to match Vulkan's 32-bit C unsigned integers.

MaikKlein: I mean it doesn't really matter, but ash uses libc everywhere else atm, so this would be the most consistent. Maybe we should switch away from libc and use rust std lib. I am honestly not sure.

GabrielMajeri:
As discussed, I've changed the pull request to use Rust's integer types everywhere, instead of using `libc` types.

GabrielMajeri:
@MaikKlein Is there anything left to, or can we get this PR merged?

MaikKlein:
I think we probably should remove libc altogether? Or is there another use case? 

I think rusts ffi types misses things like wchar, but that just seems to be a typedef for `i32`.

GabrielMajeri:
@MaikKlein `wchar_t` is `u32` on most platforms, but on Windows (the only place where it is even used...) is `u16`.

GabrielMajeri:
Done. `libc` is gone. We now use Rust's integer types wherever possible.

PR #117, opened 2018-08-27
WIP: Implement builder-like setters

MatusT:
This PR implements builder-like setters in order to reduce the noise created by manual filling of Vulkan's structs. It is supposed to work nicely with the implementation of Default trait and further improve it. The main contribution is in more rustier versions of p_ and pp_ fields, now taking Rust's slices and filling appropriate _count or _size field automatically.

Example of code before:
```rust
    let device_create_info = vk::DeviceCreateInfo {
        p_queue_create_infos: &graphics_queue_create_info,
        queue_create_info_count: 1,

        enabled_layer_count: instance_layers_names.len() as u32,
        pp_enabled_layer_names: instance_layers_names_raw.as_ptr(),

        enabled_extension_count: device_extension_names_raw.len() as u32,
        pp_enabled_extension_names: device_extension_names_raw.as_ptr(),

        p_enabled_features: &device_features,
        ..Default::default()
    };

    let pipeline_vertex_input_state_create_info = vk::PipelineVertexInputStateCreateInfo {
        vertex_binding_description_count: Vertex::binding_descriptions().len() as u32,
        p_vertex_binding_descriptions: (&Vertex::binding_descriptions()).as_ptr(),
        vertex_attribute_description_count: Vertex::attribute_descriptions().len() as u32,
        p_vertex_attribute_descriptions: (&Vertex::attribute_descriptions()).as_ptr(),
        ..Default::default()
    };
```

And example of code after applying builder-like pattern.
```rust
    let device_create_info = vk::DeviceCreateInfo::default()
        .p_queue_create_infos(&[graphics_queue_create_info])
        .pp_enabled_layer_names(&instance_layers_names_raw)
        .pp_enabled_extension_names(device_extension_names_raw)
        .p_enabled_features(&device_features);

    let pipeline_vertex_input_state_create_info = vk::PipelineVertexInputStateCreateInfo::default()
        .p_vertex_binding_descriptions(&Vertex::binding_descriptions())
        .p_vertex_attribute_descriptions(&Vertex::attribute_descriptions());
```

Please let me know your opinion on this in the comments.

EDIT:
Implementation based on the PhantomData and Deref is in the PR. Things to resolve:

- [x] should p_ and pp_ prefixes be removed? yes
- [x] what should be taken instead of *const char? &str, CStr? &CStr for p_ &[*const c_char] for pp_
- [x] does Ash even want this pattern or should this be delegated to higher-level libraries? yes

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/117)

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1283,6 +1283,153 @@ pub fn derive_debug(_struct: &vkxml::Struct, union_types: &HashSet<&str>) -> Opt
     Some(q)
 }
 
+pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
+    let name = name_to_tokens(&_struct.name);
+    let name_builder = name_to_tokens(&(_struct.name.to_string() + "Builder"));
+
+    let members = _struct.elements.iter().filter_map(|elem| match *elem {
+        vkxml::StructElement::Member(ref field) => Some(field),
+        _ => None,
+    });
+
+    let setters = members.clone().filter_map(|field| {
+        let param_ident = field.param_ident();
+        let param_ty_tokens = field.type_tokens();
+
+        let param_ident_string = param_ident.to_string();
+        if param_ident_string == "s_type" || param_ident_string == "p_next" {
+            return None;
+        }
+
+        let mut param_ident_short = param_ident_string.as_str();
+        if param_ident_string.starts_with("p_") {
+            param_ident_short = &param_ident_string[2..];
+        };
+        if param_ident_string.starts_with("pp_") {
+            param_ident_short = &param_ident_string[3..];
+        };
+        let param_ident_short = Term::intern(&param_ident_short);
+
+        // TODO: Improve in future when https://github.com/rust-lang/rust/issues/53667 is merged
+        if param_ident.to_string().starts_with("p_") || param_ident.to_string().starts_with("pp_") {
+            let param_ty_string = param_ty_tokens.to_string();
+
+            if param_ty_string == "*const c_char" {
+                return Some(quote!{
+                        pub fn #param_ident_short(mut self, #param_ident_short: &'a ::std::ffi::CStr) -> #name_builder<'a> {
+                            self.inner.#param_ident = #param_ident_short.as_ptr();
+                            self
+                        }
+                }); 
+            }
+
+            if let Some(ref array_type) = field.array {
+                if let Some(ref array_size) = field.size {
+                    if !array_size.starts_with("latexmath") {
+                        let length_type;
+                        let array_size_ident = Ident::from(array_size.to_snake_case().as_str());  
+                        if array_size_ident.to_string().contains("_count") {
+                            length_type = Term::intern("u32");
+                        } else {
+                            length_type = Term::intern("usize");
+                        }                        
+
+                        if param_ty_string == "*const *const c_char" {
+                            return Some(quote!{
+                                    pub fn #param_ident_short(mut self, #param_ident_short: &'a [*const c_char]) -> #name_builder<'a> {
+                                        self.inner.#param_ident = #param_ident_short.as_ptr();
+                                        self.inner.#array_size_ident = #param_ident_short.len() as #length_type;
+                                        self
+                                    }
+                            }); 
+                        }                      
+
+                        let slice_param_ty_tokens;
+                        let ptr_mutability;
+                        if param_ty_string.starts_with("*const ") {
+                            slice_param_ty_tokens = "&'a [".to_string() + &param_ty_string[7..] + "]";
+                            ptr_mutability = ".as_ptr()";
+                        } else {
+                            // *mut
+                            slice_param_ty_tokens =
+                                "&'a mut [".to_string() + &param_ty_string[5..] + "]";
+                            ptr_mutability = ".as_mut_ptr()";
+                        }
+                        let slice_param_ty_tokens = Term::intern(&slice_param_ty_tokens);
+                        let ptr_mutability = Term::intern(ptr_mutability);
+
+
+                        match array_type {
+                            vkxml::ArrayType::Dynamic => {
+                                return Some(quote!{
+                                    pub fn #param_ident_short(mut self, #param_ident_short: #slice_param_ty_tokens) -> #name_builder<'a> {
+                                        self.inner.#array_size_ident = #param_ident_short.len() as #length_type;
+                                        self.inner.#param_ident = #param_ident_short#ptr_mutability;
+                                        self
+                                    }
+                                });
+                            }
+                            _ => {}
+                        }
+                    }
+                }
+            }
+
+            if param_ty_string.starts_with("*const ") && param_ty_string.ends_with("Info") {
+                let slice_param_ty_tokens = "&'a ".to_string() + &param_ty_string[7..];
+                let slice_param_ty_tokens = Term::intern(&slice_param_ty_tokens);
+                return Some(quote!{
+                        pub fn #param_ident_short(mut self, #param_ident_short: #slice_param_ty_tokens) -> #name_builder<'a> {
+                            self.inner.#param_ident = #param_ident_short;
+                            self
+                        }
+                }); 
+            }
+        }
+
+        Some(quote!{
+            pub fn #param_ident_short(mut self, #param_ident_short: #param_ty_tokens) -> #name_builder<'a> {
+                self.inner.#param_ident = #param_ident_short;
+                self
+            }
+        })
+    });
+
+    let q = quote!{
+        impl #name {
+             pub fn builder<'a>() -> #name_builder<'a> {
+                #name_builder {
+                    inner: #name::default(),
+                    marker: ::std::marker::PhantomData,
+                }
+             }
+         }
+
+        pub struct #name_builder<'a> {
+            inner: #name,
+            marker: ::std::marker::PhantomData<&'a ()>,
+        }
+
+        impl<'a> ::std::ops::Deref for #name_builder<'a> {
+            type Target = #name;
+
+            fn deref(&self) -> &Self::Target {
+                &self.inner
+            }
+        }
+
+        impl<'a> #name_builder<'a> {
+            #(#setters)*
+
+            pub fn build(self) -> #name {
+                self.inner
+            }
+        }
+    };
+
+    Some(q)
+}
+
 /// At the moment `Ash` doesn't properly derive all the necessary drives
 /// like Eq, Hash etc.
 /// To Address some cases, you can add the name of the struct that you
@@ -1308,6 +1455,7 @@ pub fn generate_struct(_struct: &vkxml::Struct, union_types: &HashSet<&str>) -> 
 
     let debug_tokens = derive_debug(_struct, union_types);
     let default_tokens = derive_default(_struct);
+    let setter_tokens = derive_setters(_struct);
     let manual_derive_tokens = manual_derives(_struct);
     let dbg_str = if debug_tokens.is_none() {
         quote!(Debug,)
@@ -1327,6 +1475,7 @@ pub fn generate_struct(_struct: &vkxml::Struct, union_types: &HashSet<&str>) -> 
         }
         #debug_tokens
         #default_tokens
+        #setter_tokens
     }
 }
 
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -463,18 +463,18 @@ impl ::std::clone::Clone for InstanceFnV1_0 {
             enumerate_physical_devices: self.enumerate_physical_devices,
             get_physical_device_features: self.get_physical_device_features,
             get_physical_device_format_properties: self.get_physical_device_format_properties,
-            get_physical_device_image_format_properties:
-                self.get_physical_device_image_format_properties,
+            get_physical_device_image_format_properties: self
+                .get_physical_device_image_format_properties,
             get_physical_device_properties: self.get_physical_device_properties,
-            get_physical_device_queue_family_properties:
-                self.get_physical_device_queue_family_properties,
+            get_physical_device_queue_family_properties: self
+                .get_physical_device_queue_family_properties,
             get_physical_device_memory_properties: self.get_physical_device_memory_properties,
             get_device_proc_addr: self.get_device_proc_addr,
             create_device: self.create_device,
             enumerate_device_extension_properties: self.enumerate_device_extension_properties,
             enumerate_device_layer_properties: self.enumerate_device_layer_properties,
-            get_physical_device_sparse_image_format_properties:
-                self.get_physical_device_sparse_image_format_properties,
+            get_physical_device_sparse_image_format_properties: self
+                .get_physical_device_sparse_image_format_properties,
         }
     }
 }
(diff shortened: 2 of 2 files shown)


Ralith:
Ooh, great idea! Not being able to ergonomically supply slices to structs was always disappointing.

As written, however, I am worried that this is a bit of a landmine. In your example code, I believe the array literals are short-lived--their lifetime ends after the `DeviceCreateInfo` is constructed but before it's used, resulting in UB when the vulkan impl dereferences the pointer. I wonder if there's a way we can finagle the lifetime annotations to require that the slices outlive `self`?

I'd also suggest stripping the `p_`/`pp_` prefixes, since slices aren't pointers, a unified setter doesn't need the disambiguation so much, and rust has strong types.

MaikKlein:
I assume that something like this should work
```
struct Foo(*const u33);
impl Foo 
 {
    fn set<'a>(&'a mut self, data: &'a u32) -> &'a mut Foo{
        self.0 = data as _;
        self
    }
}
fn get_value() -> u32 {
    42
}
fn main() {
    /// Doesn't compile
    //let mut foo = Foo(std::ptr::null()).set(&get_value());
    // Note that something like `&3` gets an extended lifetime because
    // it is an rvalue reference of a constant
    let value = get_value();
    let mut foo = Foo(std::ptr::null());
    foo.set(&value);
}
```

This should then look like 
```
let bindings = Vertex::binding_descriptions();
let desc = Vertex::attribute_descriptions();
let mut pipeline_vertex_input_state_create_info = vk::PipelineVertexInputStateCreateInfo::default();
pipeline_vertex_input_state_create_info
    .vertex_binding_descriptions(&bindings)
    .vertex_attribute_descriptions(&desc);

``` 

Thoughts?

MaikKlein:
```
use std::marker::PhantomData;
struct Foo<'a>(*const u32, *const f32, PhantomData<&'a ()>);
impl<'a> Foo<'a> {
    fn set_u32(mut self, data: &'a u32) -> Foo<'a> {
        self.0 = data as _;
        self
    }
    fn set_f32(mut self, data: &'a f32) -> Foo<'a> {
        self.1 = data as _;
        self
    }
}

fn get_value() -> u32 {
    42
}
fn main() {
    let v = 3;
    let f = 3.0;
    let foo = Foo(std::ptr::null(), std::ptr::null(), PhantomData)
        .set_u32(&v)
        .set_f32(&f);
}
```
https://play.rust-lang.org/?gist=d14a77c92f295ccdebe2b4dd90ec9afa&version=stable&mode=debug&edition=2015

This should work and users can still store ii with `Foo<'static>` if they wanted to, although then they can't use the builder.
But this requires some significant work to the generator because now the `impl` requires a lifetime. I also don't want to make the generator too complex.

Alternatively we could just accept it as it is, deref happens inside Vulkan and those functions are already marked as unsafe.

omni-viral:
> deref happens inside Vulkan and those functions are already marked as unsafe.

But this hides raw pointers behind references.
Yet with `PhantomData` (any ZST in general) the type become not ffi-safe.

GabrielMajeri:
> Yet with `PhantomData` (any ZST in general) the type become not ffi-safe.

Can't [`#[repr(transparent)]`](https://github.com/rust-lang/rust/issues/43036) help?

MaikKlein:
>Can't #[repr(transparent)] help?

This only works for structs with one variant. But I am not sure why PhantomData would be unsafe in FFI? Where is this documented? I always assumed that it is just ignored in codegen?

GabrielMajeri:
> This only works for structs with one variant.

I assume you mean only one field. But that's not the case. Example rom the RFC:

```rust
// Transparent struct wrapper with a marker.
#[repr(transparent)]
struct TransparentWrapper<T> {
    only_non_zero_sized_field: f64,
    marker: PhantomData<T>,
}
```

The non-zero-sized field could be the actual structure we need to pass the Vulkan. The builder wraps that structure and adds a PhantomData marker.

MaikKlein:
> The non-zero-sized field could be the actual structure we need to pass to Vulkan. The builder wraps that structure and adds a PhantomData marker.

That would be another possibility 
```
use std::marker::PhantomData;
struct Builder<'a, T> {
    build: T,
    _m: PhantomData<&'a ()>,
}

struct Foo(*const u32);
impl Foo {
    fn builder<'a>() -> Builder<'a, Foo> {
        Builder {
            build: Foo(std::ptr::null()), // Use default here or sth
            _m: PhantomData,
        }
    }
}

impl<'a, T> Builder<'a, T> {
    pub fn build(self) -> T {
        self.build
    }
}
impl<'a> Builder<'a, Foo> {
    fn set(mut self, data: &'a u32) -> Builder<'a, Foo> {
        self.build.0 = data as _;
        self
    }
}

fn get_value() -> u32 {
    42
}
fn main() {
    let v = 3;
    let foo = Foo::builder().set(&v).build();
}
```
https://play.rust-lang.org/?gist=d14a77c92f295ccdebe2b4dd90ec9afa&version=stable&mode=debug&edition=2015

Ralith:
That also avoids needing to deal with impact on other traits, not to mention complicating every vulkan function with the addition of lifetime arguments; I like it.

> I guess we could also just implement Deref<Target=T> for the builder, then we don't even need the build method.

I think this is preferable, because it doesn't interfere with lifetime checking.

MaikKlein:
@omni-viral I think this is best suited for layers / validation layers.

Ralith:
> should p_ and pp_ prefixes be removed?

I'd say yes; after all, these methods may set multiple fields (e.g. a length field too), and it leaves the door open for other higher-level settlers in the future.

> what should be taken instead of *const char? &str, CStr?

`&CStr` is the correct type for "borrowed C string".

> does Ash even want this pattern or should this be delegated to higher-level libraries?

I think this is exactly the sort of zero-cost helper Ash should have. There's discussion of exposing a bindings-only sys crate at https://github.com/MaikKlein/ash/pull/93 for when that's not desired.

MaikKlein:
> should p_ and pp_ prefixes be removed?

Yes I think that would make sense.

> what should be taken instead of *const char? &str, CStr?

`CStr`
 
> does Ash even want this pattern or should this be delegated to higher-level libraries?

Yes this fits Ash perfectly.

kocsis1david:
It doesn't seem useful to me to have `enabled_layer_count` after the `enabled_layer_names` setter.

MaikKlein:
I only do a partial review for the generated code for now. I'll look at it more closely tomorrow.

Some thoughts:

All the count setters are kind of useless. You might be able to filter them out. I am not sure if there are any edge cases. For example a field with the count name that is not used for an array. I think you can check for the substring in the names

For example

```
            enabled_layer_count: u32::default(),
            pp_enabled_layer_names: ::std::ptr::null(),
```
`enabled_layer` appears as a substring more than one time, and then we should remove the count method.

```
    pub fn sample_mask(
        mut self,
        sample_mask: *const SampleMask,
    ) -> PipelineMultisampleStateCreateInfoBuilder<'a> {
        self.inner.p_sample_mask = sample_mask;
        self
    }
```
I think this could become
```
    pub fn sample_mask(
        mut self,
        sample_mask: &'a SampleMask,
    ) -> PipelineMultisampleStateCreateInfoBuilder<'a> {
        self.inner.p_sample_mask = sample_mask;
        self
    }
```
I haven't yet thought about all the implications yet. I guess it could protect you from very stupid things but the lifetime is only tied to the builder and so it might appear more safe than it actually is. But the builder also using borrows for slices so this would make it more consistent.

kocsis1david:
While using this branch, I've noticed that inside `SubmitInfoBuilder`, the `wait_semaphore_count` is set by both `wait_semaphores` and `wait_dst_stage_mask` functions, but maybe it's not that important. It's nice to use the setters.

MaikKlein:
> While using this branch, I've noticed that inside SubmitInfoBuilder, the wait_semaphore_count is set by both wait_semaphores and wait_dst_stage_mask functions, but maybe it's not that important. It's nice to use the setters.

Hm that does seem quite tricky to properly implement. I guess we could add escape hatches for specific types where we could manually implement some of these setters.

Ralith:
> All the count setters are kind of useless. You might be able to filter them out.

It should be straightforward to filter these out by never generating setters for anything referenced by a `len` attribute in the same struct.

> I guess we could add escape hatches for specific types where we could manually implement some of these setters.

This isn't absolutely necessary. For example, you could detect when a single length field is used by multiple pointers, and generate whatever preferred implementation in that case. I'm not sure what the preferred implementation for this case would be, though...

kocsis1david:
Maybe just an `assert!` or `debug_assert!` is enough to check if the slice length matches the previous count value. But currently there's no harm by assigning the count twice, it will be optimized away. It's not safe, but  I'm not sure if the builder is supposed to be super safe.

Ralith:
A `debug_assert!` is a nice thought, but is redundant to the validation layers, and wouldn't be able to catch all the cases anyway (since you can't distinguish uninitialized from assigned-to-zero, or easily ensure that other related slices get assigned at all). I think the practical approach is just to accept the unsafety.

generator/src/lib.rs
```diff
@@ -1292,7 +1292,24 @@ pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
         _ => None,
     });
 
+    let count_members: Vec<String> = members.clone().filter_map(|field| {
+        if field.array.is_some() {
+            if let Some(ref array_size) = field.size {
+                if !array_size.starts_with("latexmath") && array_size.ends_with("Count") {
```
Ralith: What invalid cases are excluded by the `ends_with("Count")` constraint?

MatusT: That's a failure in my brain. Thanks for catching that.

MatusT:
(Hopefully) last thing to resolve: what about pCode and CodeSize in VkShaderModuleCreateInfo? 
- codeSize must be divisible by 4
- pCode can either be a glsl string or binary depending on the presence of the glsl extension

Questions:
- Should the pCode be padded in the builder to be divisible by 4?
- Should the pCode be &[u8], &Cstr, something else?

MatusT:
pCode in ShaderModuleCreateInfo is now handled properly

Ralith:
On second thought, alignment might be an issue here; unless someone knows better, `&[u32]` would certainly be safest, and if people want to risk punning then they can do so themselves, as when using the structs directly.

omni-viral:
`DeviceCreateInfoBuilder::enabled_features` takes raw pointer instead of reference.

omni-viral:
IIRC. Convering `&T` to `&[u32]` can trigger UB.

MaikKlein:
Also I used `&[u8]` in a couple of other places. The reasoning here is that the validation layer will trigger when the alignment is incorrect, and it wouldn't with `&[u32]`.

> IIRC. Convering &T to &[u32] can trigger UB.

I think that should be fine with `slice::from_raw_parts`.

omni-viral:
I don't know where to find precise memory model for rust.
But from what I know from C you can't cast pointers to/from arbitrary types.

MatusT:
It seems in C99 type punning is fine if done through unions[1]. 
Rust (thanks to LLVM IR?) doesn't seem to have strict aliasing restrictions[2] and it is only necessary to keep alignment in order.

In any case, is there any user-friendly way to create shader module with SPIR-V obtained from include_bytes aside from converting it to vector of proper alignment? Since one cannot specify alignment in that case.

[1] https://stackoverflow.com/questions/25664848/unions-and-type-punning
[2] https://stackoverflow.com/questions/26163272/strict-aliasing-in-rust

MaikKlein:
`&[u8]`, its the most natural. `&[u32]` enforces alignment but that doesn't really help if the data is invalid in the first place. Just converting it to `&[u32]` doesn't really make it more safe. But the vulkan api uses `&[u32]` so that is what we probably should use.

> In any case, is there any user-friendly way to create shader module with SPIR-V obtained from include_bytes aside from converting it to vector of proper alignment?

You don't need to worry about alignment, either it is valid in the first place or it is not. You can use `from_raw_parts` to convert it. If it is a ptr (which you can get with `as_ptr()` on a slice) you can just cast `*const u8` to `*const u32`. The length for the shader module is still in **bytes**, so be careful about that. 

I am not sure what the best API for the builder would be.

omni-viral:
> You don't need to worry about alignment, either it is valid in the first place or it is not

That's why you *should* worry.

MaikKlein:
> That's why you should worry.

That is not the job of ash. As a user you need to check this yourself and make sure the SPIR-V that you load is valid.

> No, you can't.

That is not really a helpful comment at all. Why not? 

> A SPIR-V module is a single linear stream of **words**.

Surely you can read it as bytes and cast it to u32? Why do you think that would be illegal?

kocsis1david:
`&[u32]` is not convenient to use, because if you read it from a file or use `include_bytes!`, it will be `&[u8]`. Correct me if I'm wrong, but to use it safely you have to copy it to a different buffer to make it `&[u32]`. People don't bother with this and just cast the pointer which is not safer than taking `&[u8]`. But I understand that you want show with the API that it needs 4 byte alignment.

I think Rust doesn't have a perfect way of describing a type that is a binary blob with a certain alignment. It doesn't have to consist of only `u32` integers, it may also contain floats and anything else mixed. A common convention is to always use byte arrays in other languages, but yes, that doesn't represent alignment.

Ralith:
> People don't bother with this and just cast the pointer which is not safer than taking &[u8].

It's better to make the unsafety obvious than to silently perform such a conversion internally.

> I think Rust doesn't have a perfect way of describing a type that is a binary blob with a certain alignment. It doesn't have to consist of only u32 integers, it may also contain floats and anything else mixed. A common convention is to always use byte arrays in other languages, but yes, that doesn't represent alignment requirements.

The convention established by Vulkan here is to treat it as a buffer of `u32`, per the wrapped structure. If any convention takes precedence I imagine that's the one that should.

kocsis1david:
> The convention established by Vulkan here is to treat it as a buffer of u32, per the wrapped structure. If any convention takes precedence I imagine that's the one that should.

But even in Vulkan, `codeSize` is in bytes, only the pointer points to `uint32_t`, so it's an exception for the generator.

Maybe the correct way is to use `&[u32]`.
But then the problem is: One does not simply load a SPIR-V module.

Ralith:
It's definitely weird and awkward, and if there was authoritative information that the alignment implications of the pointer type could be ignored then I'd go for that in a heartbeat.

MaikKlein:
I think `&[u8]` would be better but because Vulkan uses `&[u32]`I think that is what we should use. Honestly it doesn't really matter that much.

The PR looks fine to me, if there are no objections I am going to merge it tomorrow. And @MatusT thank you for all the hard work.

kocsis1david:
`VkPipelineMultisampleStateCreateInfo.pSampleMask` is an array of `rasterizationSamples` bits, so it shouldn't be a reference I guess, but I don't understand why Vulkan has `rasterizationSamples` as an enum rather than integer.

kocsis1david:
One problem with `VkPipelineViewportStateCreateInfo`: If you want to use dynamic state, the `viewportCount` has to be filled, but the `pViewports` should be `NULL`.

MatusT:
@kocsis1david 
VkPipelineMultisampleStateCreateInfo.pSampleMask: good catch. It seems that this is because I was filtering out members which I shouldn't have done. When you look at documentation and/or vk.xml there is: ```<member optional="true" len="latexmath:[\lceil{\mathit{rasterizationSamples} \over 32}\rceil]" altlen="(rasterizationSamples + 31) / 32">```. It is one of two members using latexmath, the other being codeSize in ShaderModule, and I don't think there is any general solution except ignoring the latexmath and special casing some members as needed.

VkPipelineViewportStateCreateInfo.viewportCount: now this is quite a problem. Because it was agreed that we filter out ```count``` members that are associated with array members and as such can be removed. If there are no more of these cases except pViewports(and pScissors) a special case could be added.

TL;DR: if there are no more of these cases => treat them as special cases and let's add manual overrides similar to pCode and pCodeSize. @MaikKlein @omni-viral Sounds ok?

kocsis1david:
One more thing is `dynamic_state` inside the pipeline state is also nullable, a simple reference setter doesn't allows setting it to null. But as a workaround, you can leave it out from the builder calls and its default will be null.

MaikKlein:
>if there are no more of these cases => treat them as special cases and let's add manual overrides similar to pCode and pCodeSize.

I think you meant ~no~ more cases? Yes definitively, I kinda of expect a few more special cases to pop up. I think it would probably make more sense to catch those `setters` by name, don't generate any code for them, and then implement everything manually.

Ralith:
Automatically generating everything according to general rules is never *quite* going to work out right; a small proportion of special cases is expected and absolutely fine.

kocsis1david:
`DescriptorSetLayoutBinding` has no setter for `descriptor_count`

BTW, I hope you don't mind if I write my findings here or is there a better place?

MaikKlein:
@kocsis1david It is okay, at least everything is still in one place.

MatusT:
So far I fixed:
- counts: viewportCount, scissorCount, descriptorCount (there is an array in the code where you can just add any additional counts you find)
- sampleMask as an additional special case

Should I open additional PRs for builder fixes?

MatusT:
The Bool32 -> bool is implemented as well.

Ralith:
That's a pretty cool idea, but I don't think it should block getting the badly needed next release out the door.

MaikKlein:
@omni-viral Seems doable 

I looked at the vk.xml
```
        <type category="struct" name="VkDebugReportCallbackCreateInfoEXT" structextends="VkInstanceCreateInfo">
            <member values="VK_STRUCTURE_TYPE_DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT"><type>VkStructureType</type> <name>sType</name></member>
            <member>const <type>void</type>*                      <name>pNext</name></member>
            <member optional="true"><type>VkDebugReportFlagsEXT</type>            <name>flags</name><comment>Indicates which events call this callback</comment></member>
            <member><type>PFN_vkDebugReportCallbackEXT</type>     <name>pfnCallback</name><comment>Function pointer of a callback function</comment></member>
            <member optional="true"><type>void</type>*            <name>pUserData</name><comment>User data provided to callback function</comment></member>
        </type>
```
I think we can rely on `structextends` to indicate the `p_next`.

omni-viral:
@Ralith thanks. It's just an idea. It isn't blocking. I just figured it is good place to write it down :smile:

MaikKlein:
@MatusT Yes PRs would be appreciated.

Lets open separate issues, Github starts to get slow.

PR #118, opened 2018-08-29
Fix Windows surface creation in the example

GabrielMajeri:
The surface creation code expects `hinstance` to be the handle of the process with created the window. For some reason, in the example we were passing in a handle to the top level window.

The `HINSTANCE` parameter isn't really used for anything on modern Windows, and even the validation layers will gladly accept invalid values, but the example should be technically correct.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/118)


MaikKlein:
Interesting, I didn't know that. I rarely use windows myself.

PR #122, opened 2018-09-16
More consistent names in flags API

Ralith:
Improves consistency with other newtypes and with bitflags and allows construction of statically unknown flags.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/122)

### DIFF
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -11,25 +11,25 @@ pub trait Handle {
 }
 #[macro_export]
 macro_rules! vk_make_version {
-    ($major:expr, $minor:expr, $patch:expr) => {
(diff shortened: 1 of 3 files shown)


GabrielMajeri:
+1. With `a.subset(b)`, it's not clear that I mean that _`b` is a subset of `a`_, instead of the opposite. `a` `contains` `b` is a lot clearer.

MaikKlein:
I agree with `contains`. I think we should try to be as similar to bitflags as possible. Maybe `as_raw` should become `bits`?

Ralith:
I'm not sure. We definitely need `from_raw`, and the symmetry and consistency with handles is nice.

PR #124, opened 2018-09-29
Flatten vk.rs module structure and use f32

Ralith:
This clarifies generated documentation and removes one more unnecessary C type.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/124)

### DIFF
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -1,7 +1,3 @@
-#[doc(hidden)]
-pub use self::bitflags::*;
-#[doc(hidden)]
-pub use self::extensions::*;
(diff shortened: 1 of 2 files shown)


MaikKlein:
> This clarifies generated documentation

Can you explain the reasoning? I wrote it this way because then all the flags are in one place inside the docs.

Ralith:
It's confusing for the documentation to claim a type is defined in a submodule when in fact it is normally accessed as a member of the root module

> I wrote it this way because then all the flags are in one place inside the docs.

This should no longer be an issue due to the adoption of associated constants.

PR #130, opened 2018-10-07
Generate bindings in a build script

Ralith:
This reduces commit churn, repository size, and incidence of merge conflicts.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/130)


MaikKlein:
I must say I am not sure what the best action here would be. It definitely would be nicer that way, especially with the PRs that have to be rebased for every change to the generator.

But it is also nice to see a diff of the changes for the `vk.rs` file in the PR. Also we would lose all the auto completion related to the `vk.rs` file.

Ralith:
> it is also nice to see a diff of the changes for the vk.rs file in the PR

In my experience, these diffs are usually flooded with garbage to the point of uselessness due to trivial formatting changes.

> Also we would lose all the auto completion related to the vk.rs file.

That sounds like a bug in a specific autocompletion tool. It should be able to respect `include!` just like the rest of the toolchain. We do actually generate a vk.rs still, after all, it's just not in version control.

Ralith:
This could be reprised as a procedural macro, which might make tooling happier.

MaikKlein:
I don't think the current tools beside intellij can do any of this. But RLS 2.0 is on the rust roadmap, so hopefully by then this will be fixed.

I still like that we have autocompletion right now, and as an author I am willing to endure a little bit of rebasing pain for it.

But I definitely agree with the PR, this should really be automated.

Ralith:
Closing this as no longer desirable for two reasons:
> In my experience, these diffs are usually flooded with garbage to the point of uselessness due to trivial formatting changes.

This was fixed by #193.

More importantly, though, I'm coming around to the perspective that build scripts should be avoided whenever possible as they're very difficult to reason about. Publishing generated code is easier to audit and package.

Issue #41, opened 2017-12-29
[RFC] Error handling


msiglreith:
There are function which convert a successful return code into an error.
For example [`acquire_next_image`](https://github.com/MaikKlein/ash/blob/master/ash/src/extensions/swapchain.rs#L49) may return `VK_SUBOPTIMAL_KHR` among other codes, indicating that the return index could still be used but the swapchain doesn't match the surface anymore. In the current API, this would return an error.

My suggestion would be to return a tuple of `(ReturnType, SuccessCode)` in these cases.

kvark:
Seems a bit inconsistent to return a tuple in some cases and `Result` in others. And clearly, the latter is preferred if possible. Maybe we can have enumerations for those specific success results, making the return types `Result<SpecificSuccessEnum, GenericError>` for those cases?

MaikKlein:
That seems reasonable, I am also leaning more towards custom success enums. Alternatively I think I could also just keep it as it currently is, but return all success error codes in the `Ok` variant.

    pub unsafe fn acquire_next_image_khr(
        &self,
        swapchain: vk::SwapchainKHR,
        timeout: vk::uint64_t,
        semaphore: vk::Semaphore,
        fence: vk::Fence,
    ) -> VkResult<vk::uint32_t> {
        let mut index = mem::uninitialized();
        let err_code = self.swapchain_fn.acquire_next_image_khr(
            self.handle,
            swapchain,
            timeout,
            semaphore,
            fence,
            &mut index,
        );
        match err_code {
            vk::Result::Success | vk::Resut::SubObtimalKHR | vk::Result::Timeout | .. => Ok(index),
            _ => Err(err_code),
        }
    }

Thoughts?

kvark:
I like that

msiglreith:
> Seems a bit inconsistent to return a tuple in some cases and Result in others. And clearly, the latter is preferred if possible.

I agree. I meant more like `Result<(Index, SuccessCode), ErrorCode>`.
The current form is `Result<T, VkResult>` for all function. I would suggest here to replace `VkResult` with `ErrorCode` containing only negative values (errors) of the `VkResult` type. If a function returns any other value than `Success` on success it should be in the `Ok` variant along the actual return value.

LPGhatguy:
I like the idea of changing all of the functions to return `Result<(T, SuccessCode), ErrorCode>` -- I think we should sure not to lose the status code in the case of success.

[...]

PR #132, opened 2018-10-07
Expose suboptimal results from swapchain operations

Ralith:
Conservative approach to #41. Are there any other cases where information is currently being lost?

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/132)

### DIFF
--- a/ash/src/extensions/swapchain.rs
+++ b/ash/src/extensions/swapchain.rs
@@ -43,13 +43,14 @@ impl Swapchain {
         );
     }
 
+    /// On success, returns the next image's index and whether the swapchain is suboptimal for the surface.
     pub unsafe fn acquire_next_image_khr(
         &self,
         swapchain: vk::SwapchainKHR,
         timeout: u64,
         semaphore: vk::Semaphore,
         fence: vk::Fence,
-    ) -> VkResult<u32> {
+    ) -> VkResult<(u32, bool)> {
         let mut index = mem::uninitialized();
         let err_code = self.swapchain_fn.acquire_next_image_khr(
             self.handle,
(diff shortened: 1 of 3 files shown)


MaikKlein:
There are a few places probably that are not quite obvious. For example

```
unsafe fn create_graphics_pipelines(
    &self, 
    pipeline_cache: PipelineCache, 
    create_infos: &[GraphicsPipelineCreateInfo], 
    allocation_callbacks: Option<&AllocationCallbacks>
) -> Result<Vec<Pipeline>, (Vec<Pipeline>, Result)>
```

The proposed change was `Result<(Index, SuccessCode), ErrorCode>` but I am leaning more towards

```
    pub unsafe fn acquire_next_image_khr(
        &self,
        swapchain: vk::SwapchainKHR,
        timeout: u64,
        semaphore: vk::Semaphore,
        fence: vk::Fence,
    ) -> Result<u32, (u32, vk::Result)> {
```

I think this would be more consistent with the rest of the api.



```
let index = match acquire_next_image_khr(..) {
    Some(idx) => idx,
    Err((idx, err)) => {
        // Handle the warnings, recreate the swapchain
        panic!(err == vk::Result::SUBOPTIMAL_KHR);
        idx
    }
};
```

Ralith:
> The proposed change was `Result<(Index, SuccessCode), ErrorCode>`

Note that all "success" codes other than `SUBOPTIMAL` still leave the index uninitialized in this example. That's why I reduced it to a single boolean.

> `Result<u32, (u32, vk::Result)>`

We don't want to return uninitialized data in the error case, since it is UB to access. We could return an arbitrary invalid index instead, but this still strikes me as unnecessarily error-prone. Why return an index when we can trivially determine that there is no correct index to return?

PR #133, opened 2018-10-08
Fix Vulkan 1.1 getters

### LINKED ISSUE #109: Vulkan 1.1 getter functions don't work

Ralith:
Fixes #109

I checked every vk.xml struct marked `returnedonly`, and made a pass to sanity-check wrapped 1.1 getters in general. Might've missed some, but probably not many.

### DIFF
--- a/ash/src/instance.rs
+++ b/ash/src/instance.rs
@@ -87,25 +87,30 @@ impl<V: FunctionPointers> Instance<V> {
 pub trait InstanceV1_1: InstanceV1_0 {
     fn fp_v1_1(&self) -> &vk::InstanceFnV1_1;
 
-    fn enumerate_physical_device_groups(&self) -> VkResult<Vec<vk::PhysicalDeviceGroupProperties>> {
+    unsafe fn enumerate_physical_device_groups_len(&self) -> usize {
+        let mut group_count = mem::uninitialized();
+        self.fp_v1_1().enumerate_physical_device_groups(
+            self.handle(),
+            &mut group_count,
+            ptr::null_mut(),
+        );
+        group_count as usize
+    }
+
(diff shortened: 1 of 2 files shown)


ash/src/instance.rs
```diff
@@ -87,25 +87,30 @@ impl<V: FunctionPointers> Instance<V> {
 pub trait InstanceV1_1: InstanceV1_0 {
     fn fp_v1_1(&self) -> &vk::InstanceFnV1_1;
 
-    fn enumerate_physical_device_groups(&self) -> VkResult<Vec<vk::PhysicalDeviceGroupProperties>> {
```
MaikKlein: Can you remind me why this function needs to be changed? What is wrong with returning `Vec<vk::PhysicalDeviceGroupProperties>`? I think the user can just check the length to see if it would be `INCOMPLETE`.

PR #137, opened 2018-11-03
Implement panicking function pointers

MaikKlein:
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/137)

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -775,7 +775,7 @@ fn generate_function_pointers(ident: Ident, commands: &[&vkxml::Command]) -> quo
                         #names_ref: unsafe {
 
                             extern "system" fn #names_ref1 (#expanded_params_unused) -> #return_types_ref {
-                                panic!("Unable to load {}", stringify!(#names_ref2))
+                                panic!(concat!("Unable to load ", stringify!(#names_ref2)))
                             }
                             let raw_name = stringify!(#raw_names_ref);
                             let cname = ::std::ffi::CString::new(raw_name).unwrap();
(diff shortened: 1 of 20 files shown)


generator/src/lib.rs
```diff
@@ -753,31 +767,28 @@ fn generate_function_pointers(ident: Ident, commands: &[&vkxml::Command]) -> quo
...
                 where F: FnMut(&::std::ffi::CStr) -> *const c_void
             {
-                let mut _err_str = Vec::new();
-                let s = #ident {
+                #ident {
                     #(
                         #names_ref: unsafe {
+
+                            extern "system" fn #names_ref1 (#expanded_params_unused) -> #return_types_ref {
+                                panic!("Unable to load {}", stringify!(#names_ref2))
```
Ralith: This could be `panic!(concat!("Unable to load ", stringify!(#names_ref2)))`

Ralith:
Should the `DeviceError` enum be removed, since load errors are no longer a thing? Also, with these changes, is there any value left to having the elaborate version-genericity support, since it seems like loading will succeed unconditionally regardless?

MaikKlein:
>Also, with these changes, is there any value left to having the elaborate version-genericity support, since it seems like loading will succeed unconditionally regardless?

I think we can remove the versioning, but I think I want to retain the version specific traits for now. Like `DeviceV1_0`.

Ralith:
Then what versioning are you removing, exactly?

MaikKlein:
There would only be one Device / Instance struct and all the versioned traits would be implemented for them. That is at the moment the idea that I am toying with.

MatusT:
I am all for simplification. IMO all those traits bring no improvement and only clutter the imports and function arguments.

Ralith:
> There would only be one Device / Instance struct and all the versioned traits would be implemented for them

What use is a trait that has only one implementation?

MaikKlein:
Right now there is only one implementation, those could be implemented for gfx-hal too. Also you can still write generic functions where you specify those traits. 

Also it requires almost no change at all right now, they can always be removed later without much breakage if they are not needed anymore.

PR #139, opened 2018-11-11
Simplify versioning

MaikKlein:
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/139)


MaikKlein:
The reasoning:

There is very little benefit for using versioning today. This PR simplifies that by loading everything and unsuccessful function pointers are initialized with a panicking function.

It is still possible to create a custom loader.

PR #143, opened 2018-11-12
Builder pattern: Bool32 -> bool + edge cases.

MatusT:
This is from discussions in original PR. I'll address the p_next in next PR.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/143)

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1283,34 +1283,43 @@ pub fn derive_debug(_struct: &vkxml::Struct, union_types: &HashSet<&str>) -> Opt
 
 pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
     let name = name_to_tokens(&_struct.name);
-    let name_builder = name_to_tokens(&(_struct.name.to_string() + "Builder"));
+    let name_builder = name_to_tokens(&(_struct.name.clone() + "Builder"));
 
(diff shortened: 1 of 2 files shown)


generator/src/lib.rs
```diff
@@ -1283,34 +1283,43 @@ pub fn derive_debug(_struct: &vkxml::Struct, union_types: &HashSet<&str>) -> Opt
...
 
     let members = _struct.elements.iter().filter_map(|elem| match *elem {
         vkxml::StructElement::Member(ref field) => Some(field),
         _ => None,
     });
 
+    let nofilter_count_members = [
+        "VkPipelineViewportStateCreateInfo.pViewports",
+        "VkPipelineViewportStateCreateInfo.pScissors",
+        "VkDescriptorSetLayoutBinding.pImmutableSamplers",
```
Ralith: Under what circumstances is it correct to have a nonzero immutable sampler count with a null `pImmutableSamplers`?

MatusT: No. You can set descriptorCount while having pImmutableSamplers null.

MatusT: pImmutableSamplers is tied to descriptorCount. There is not immutable sampler count. From https://github.com/KhronosGroup/Vulkan-Docs/blob/1.0/src/spec/vk.xml#L863

PR #146, opened 2018-11-16
Builder next

MatusT:
This PR corresponds to https://github.com/MaikKlein/ash/issues/141.

Current implementation:
- Each builder creates a trait ex.: ```ApplicationInfoBuilder``` -> ```pub trait ApplicationInfoBuilderNext```
- Structs implementing this trait are then taken in the builder's next function
- Each struct then makes a bunch of impls for the traits based on the ```extends``` attribute provided in vk.xml

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/146)

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1353,6 +1353,10 @@ pub fn derive_debug(_struct: &vkxml::Struct, union_types: &HashSet<&str>) -> Opt
 }
 
 pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
+    if &_struct.name == "VkBaseInStructure" || &_struct.name == "VkBaseOutStructure" {
+        return None;
+    }
+
     let name = name_to_tokens(&_struct.name);
     let name_builder = name_to_tokens(&(_struct.name.clone() + "Builder"));
 
@@ -1360,6 +1364,18 @@ pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
         vkxml::StructElement::Member(ref field) => Some(field),
         _ => None,
     });
+
+    let (has_next, is_next_const) = match members
+        .clone()
+        .find(|field| field.param_ident().to_string() == "p_next")
+    {
+        Some(p_next) => if p_next.type_tokens().to_string().starts_with("*const") {
+            (true, true)
+        } else {
+            (true, false)
+        },
+        None => (false, false),
+    };
 
     let nofilter_count_members = [
         "VkPipelineViewportStateCreateInfo.pViewports",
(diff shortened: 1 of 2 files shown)


generator/src/lib.rs
```diff
@@ -1523,21 +1538,62 @@ pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
...
+            nexts.push(quote! {
+                impl #target_ident for #name_builder_next {}
+            });
+        }
+    }
+
+    let next_function = if has_next {
+        if is_next_const {
+            quote!{
+                pub fn next<T>(mut self, next: &T) -> #name_builder<'a> where T: #name_builder_next {
```
Ralith:
```suggestion
                pub fn next<T>(mut self, next: &'a T) -> #name_builder<'a> where T: #name_builder_next {
```

generator/src/lib.rs
```diff
@@ -1523,21 +1538,62 @@ pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
...
+                    self.inner.p_next = next as *const T as *const c_void;
+                    self
+                }
+            }
+        } else {
+            quote!{
+                pub fn next<T>(mut self, next: &mut T) -> #name_builder<'a> where T: #name_builder_next {
```
Ralith:
```suggestion
                pub fn next<T>(mut self, next: &'a mut T) -> #name_builder<'a> where T: #name_builder_next {
```

generator/src/lib.rs
```diff
@@ -1523,21 +1538,62 @@ pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
...
+                    self
+                }
+            }
+        }
+    } else {
+        quote!{}
+    };
+
+    let next_trait = if has_next {
+        quote!{pub trait #name_builder_next {}}
```
Ralith:
```suggestion
        quote!{pub unsafe trait #name_builder_next {}}
```

MatusT: Up to this day (and I wrote about 20 000 lines of code in Rust for my bachelor thesis :D) I didn't know you could also mark a trait as unsafe. I leave this decision to @MaikKlein while I google away to learn about unsafe trait.

MaikKlein: I don't even think that a user would want to implement those traits manually, but if they do they better make sure it is valid, so yes it has to be unsafe.

Should this trait even be public?

MatusT: I agree with the indication of the fact that it is unsafe to implement it.

In theory it shouldn't be public but I can see that someone may be stuck for some reason on older Ash version in future but may want to add an extension from newer Vulkan.

generator/src/lib.rs
```diff
@@ -1391,7 +1406,7 @@ pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
         let param_ty_string = param_ty_tokens.to_string();
 
         let param_ident_string = param_ident.to_string();
-        if param_ident_string == "s_type" || param_ident_string == "p_next" {
+        if param_ident_string == "s_type" || param_ident_string == "p_next"  {
```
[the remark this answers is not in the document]
MatusT: Resolved with rustfmt.

generator/src/lib.rs
```diff
@@ -1557,7 +1615,6 @@ pub fn derive_setters(_struct: &vkxml::Struct) -> Option<Tokens> {
 
     Some(q)
 }
-
```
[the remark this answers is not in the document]
MatusT: Resolved with rustfmt.

generator/src/lib.rs
Ralith: There's at least theoretically a population users of unpublished extensions, too.

MaikKlein:
This looks fine to me, is there anything that you want to add? I am just asking because there is still WIP in the title.

MatusT:
@MaikKlein Nope, just forgot to remove WIP :).

MatusT:
Looks like one of nv extensions broke the next : /

MatusT:
```git submodule update``` wasn't enough because before rebasing it was on branch, I hate git. Anyway, all is solved and squashed.

PR #147, opened 2018-11-16
Compile fixes for macOS

gwihlidal:
The changes for Ash 0.25 resulted in the macOS demo to fail compilation. There is a new runtime loader issue that still remains, which I will open a new issue for.

---
This change is [Reviewable](https://reviewable.io/reviews/maikklein/ash/147)


MaikKlein:
Thanks, it seems like travis doesn't build it on mac. I should probably fix that asap.

Issue #142, opened 2018-11-12
Improve the examples


MaikKlein:
Since `0.25.0` we now have `Default` and builders. We probably should showcase those features in the examples.

PR #169, opened 2018-12-06
Issue 142

coolmedium:
Hey here is an attempt to address #142 .  

I have used builders in cases where a builder method accepts a structure that sets more than one struct field (e.g. fields like `x_count` and `p_to_some_x`).  Otherwise I have usually used `.. Default::default()` when doing so would save typing. 

If there is a preferred way to approach this I'm happy to make any changes.

Also, for creating shader modules I have added `examples::bytes_to_u32_vec` to convert the current shader code into a format suitable for passing to the `vk::ShaderModuleCreateInfo::code`.  It  uses `std::mem::transmute` and is based on the code for `u32::from_ne_bytes` in nightly. I have only tested this on Linux.


Ralith:
IMO pull in byteorder or do the byte-shuffling by hand without unsafe code, or convert the slice without copying. We shouldn't be encouraging gratuitous unsafe code in examples.

coolmedium:
Yeah, I almost want to wait till `u32::from_ne_bytes` hits stable.  This PR might just sit until then?

Ralith:
Why not just use a safe approach until then?

MatusT:
You use .build() almost everywhere while that shouldn't be the case. By using it, you effectively invalidate the few safety guarantees it gives and thus the example crashes on release. You should defer the .build() as late as possible and if possibly you shouldn't even use it. The create_info structs never need .build() info as they automatically defer at call site of create_*() function.

However, when you remove those .build() functions, you will receive quite a lot of errors (related to the crash as well). I only briefly went through the code and fixed the renderpass creation. You should always create arrays before the struct and then refer to them using slice, not create temporary slice inside builder functions such as .color_attachments().

Example of how I changed renderpass creation: 
```rust
        let renderpass = { 
            let renderpass_attachments = [
                vk::AttachmentDescription {
                    format: base.surface_format.format,
                    samples: vk::SampleCountFlags::TYPE_1,
                    load_op: vk::AttachmentLoadOp::CLEAR,
                    store_op: vk::AttachmentStoreOp::STORE,
                    final_layout: vk::ImageLayout::PRESENT_SRC_KHR,
                    ..Default::default()
                },
                vk::AttachmentDescription {
                    format: vk::Format::D16_UNORM,
                    samples: vk::SampleCountFlags::TYPE_1,
                    load_op: vk::AttachmentLoadOp::CLEAR,
                    final_layout: vk::ImageLayout::DEPTH_STENCIL_ATTACHMENT_OPTIMAL,
                    ..Default::default()
                },
            ];

            let color_attachment_references = [
                vk::AttachmentReference {
                    attachment: 0,
                    layout: vk::ImageLayout::COLOR_ATTACHMENT_OPTIMAL,
                }
            ];

            let depth_attachment_reference = vk::AttachmentReference {
                attachment: 1,
                layout: vk::ImageLayout::DEPTH_STENCIL_ATTACHMENT_OPTIMAL,
            };

            let dependencies = [vk::SubpassDependency {
                src_subpass: vk::SUBPASS_EXTERNAL,
                src_stage_mask: vk::PipelineStageFlags::COLOR_ATTACHMENT_OUTPUT,
                dst_access_mask: vk::AccessFlags::COLOR_ATTACHMENT_READ
                    | vk::AccessFlags::COLOR_ATTACHMENT_WRITE,
                dst_stage_mask: vk::PipelineStageFlags::COLOR_ATTACHMENT_OUTPUT,
                ..Default::default()
            }];

            let subpasses = [
                vk::SubpassDescription::builder()
                    .color_attachments(&color_attachment_references)
                    .depth_stencil_attachment(&depth_attachment_reference)
                    .pipeline_bind_point(vk::PipelineBindPoint::GRAPHICS)
                    .build()
            ];

            let create_info = vk::RenderPassCreateInfo::builder()
                .attachments(&renderpass_attachments)
                .subpasses(&subpasses)
                .dependencies(&dependencies);
            
            base
                .device
                .create_render_pass(&create_info, None)
                .unwrap()
        };
```

coolmedium:
@MatusT @Ralith 

Ah thanks for the review. I am unaccustomed to working with `unsafe` so I really appreciate having the relevant concerns pointed out. 

I hadn't even tried to run the examples with `--release` so I did not see the crashes. Thanks.  I might think about it and just trash this PR, opting to try for another that makes use of this input. 

Thanks again.

Ralith:
Note that you *can* use temporary in-line arrays so long as you're passing the built struct to the function that reads it in the same statement.

MaikKlein:
>Thanks. I might think about it and just trash this PR, opting to try for another that makes use of this input.

I don't think you need to trash the PR. 

> Btw: I don't like that we need .build() on SubpassDescription(). @MaikKlein maybe the builder could take slice of structs that can defer into the raw struct? Is that possible using Rust?

Should be possible but I don't think it will be worth the complexity that it adds.

> IMO pull in byteorder or do the byte-shuffling by hand without unsafe code, or convert the slice without copying. We shouldn't be encouraging gratuitous unsafe code in examples.

Can someone tell me what is wrong with casting the ptr? That should be fine if we validate the length before and check for the magic version. Maybe ash should provide such a function? I am not sure if we should do the endian thing, although it should be easy to detect with the magic version.

Ralith:
> Can someone tell me what is wrong with casting the ptr?

That's what I meant by "convert the slice without copying." It's fine if the alignment is right (we should really bug rust upstream for a way to get aligned `include_bytes!`) or you're on an architecture where alignment doesn't really matter (recent x86?). It may or may not produce nasal demons elsewhere (e.g. ARM).

coolmedium:
Here is a version that uses `byteorder`

```rust
pub fn bytes_to_u32_vec(bytes: &[u8]) -> Vec<u32> {
    let mut output = vec![];
    let mut buffer: [u8; 4] = [0, 0, 0, 0];
    for (i, b) in bytes.iter().enumerate() {
        let idx = i % 4;
        buffer[idx] = *b;
        if idx == 3 {
            output.push(Cursor::new(buffer).read_u32::<LittleEndian>().unwrap());
            buffer = [0, 0, 0, 0];
        }
    }
    if bytes.len() % 4 != 0 {
        output.push(Cursor::new(buffer).read_u32::<LittleEndian>().unwrap());
    }
    output
}
``` 

Though I'm not sure how portable it is. I chose little endianness b/c thats what got the examples to work on my 64 bit linux version. I'm all ears about making something like this more robust.

coolmedium:
Something like this?

```rust
const MAGIC_NUMBER: u32 = 0x07230203;

fn spirv_is_little_endian(bytes: &[u8] ) -> bool {
    let buffer: [u8;4] = [bytes[0], bytes[1],bytes[2],bytes[3]];
    let number : u32 = Cursor::new(buffer).read_u32::<LittleEndian>().unwrap();
    number == MAGIC_NUMBER
}

pub fn bytes_to_u32_vec(bytes: &[u8]) -> Vec<u32> {
    let mut output = vec![];
    let mut buffer: [u8; 4] = [0, 0, 0, 0];
    let is_little_endian = spirv_is_little_endian(bytes);
    for (i, b) in bytes.iter().enumerate() {
        let idx = i % 4;
        buffer[idx] = *b;
        if idx == 3 {
            let new_word =  if is_little_endian {
                Cursor::new(buffer).read_u32::<LittleEndian>().unwrap()
            } else {
                Cursor::new(buffer).read_u32::<BigEndian>().unwrap()
            };
            output.push(new_word);
            buffer = [0, 0, 0, 0];
        }
    }
    if bytes.len() % 4 != 0 {
        output.push(Cursor::new(buffer).read_u32::<LittleEndian>().unwrap());
    }
    output
}
```

MaikKlein:
@coolmedium There is already a PR that implements it https://github.com/MaikKlein/ash/pull/170

examples/src/bin/texture.rs
```diff
@@ -828,45 +675,28 @@ fn main() {
...
+            .input_assembly_state(&vertex_input_assembly_state_info)
+            .viewport_state(&viewport_state_info)
+            .rasterization_state(&rasterization_info)
+            .multisample_state(&multisample_state_info)
+            .depth_stencil_state(&depth_state_info)
+            .color_blend_state(&color_blend_state)
+            .dynamic_state(&dynamic_state_info)
+            .layout(pipeline_layout)
+            .render_pass(renderpass)
+            .build();
```
Ralith: Premature build

examples/src/bin/triangle.rs
```diff
@@ -413,45 +331,28 @@ fn main() {
...
+            .input_assembly_state(&vertex_input_assembly_state_info)
+            .viewport_state(&viewport_state_info)
+            .rasterization_state(&rasterization_info)
+            .multisample_state(&multisample_state_info)
+            .depth_stencil_state(&depth_state_info)
+            .color_blend_state(&color_blend_state)
+            .dynamic_state(&dynamic_state_info)
+            .layout(pipeline_layout)
+            .render_pass(renderpass)
+            .build();
```
Ralith: Premature build call

examples/src/lib.rs
```diff
@@ -69,38 +70,32 @@ pub fn record_submit_commandbuffer<D: DeviceV1_0, F: FnOnce(&D, vk::CommandBuffe
...
         device
             .end_command_buffer(command_buffer)
             .expect("End commandbuffer");
-        let fence_create_info = vk::FenceCreateInfo {
-            s_type: vk::StructureType::FENCE_CREATE_INFO,
-            p_next: ptr::null(),
-            flags: vk::FenceCreateFlags::empty(),
-        };
+
+        let fence_create_info = vk::FenceCreateInfo::default();
```
Ralith: Maybe inline this variable?

examples/src/lib.rs
```diff
@@ -69,38 +70,32 @@ pub fn record_submit_commandbuffer<D: DeviceV1_0, F: FnOnce(&D, vk::CommandBuffe
...
-            p_signal_semaphores: signal_semaphores.as_ptr(),
-        };
+
+        let command_buffers = vec![command_buffer];
+        let submit_info = vk::SubmitInfo::builder()
+            .wait_semaphores(wait_semaphores)
+            .wait_dst_stage_mask(wait_mask)
+            .command_buffers(&command_buffers)
+            .signal_semaphores(signal_semaphores)
+            .build();
```
Ralith: Premature build

examples/src/lib.rs
```diff
@@ -345,37 +336,33 @@ impl ExampleBase {
...
-                enabled_extension_count: extension_names_raw.len() as u32,
-            };
+
+            let appinfo = vk::ApplicationInfo::builder()
+                .application_name(&app_name)
+                .application_version(0)
+                .engine_name(&app_name)
+                .engine_version(0)
+                .api_version(vk_make_version!(1, 0, 36))
+                .build();
```
Ralith: Premature build

examples/src/lib.rs
```diff
@@ -345,37 +336,33 @@ impl ExampleBase {
...
+
+            let create_info = vk::InstanceCreateInfo::builder()
+                .application_info(&appinfo)
+                .enabled_layer_names(&layers_names_raw)
+                .enabled_extension_names(&extension_names_raw)
+                .build();
```
Ralith: Premature build

examples/src/lib.rs
```diff
@@ -591,45 +552,45 @@ impl ExampleBase {
...
-            let depth_image_allocate_info = vk::MemoryAllocateInfo {
-                s_type: vk::StructureType::MEMORY_ALLOCATE_INFO,
-                p_next: ptr::null(),
-                allocation_size: depth_image_memory_req.size,
-                memory_type_index: depth_image_memory_index,
-            };
+            let depth_image_allocate_info = vk::MemoryAllocateInfo::builder()
+                .allocation_size(depth_image_memory_req.size)
+                .memory_type_index(depth_image_memory_index)
+                .build();
```
Ralith: Premature build

examples/src/lib.rs
```diff
@@ -591,45 +552,45 @@ impl ExampleBase {
...
+                        .new_layout(vk::ImageLayout::DEPTH_STENCIL_ATTACHMENT_OPTIMAL)
+                        .old_layout(vk::ImageLayout::UNDEFINED)
+                        .subresource_range(
+                            vk::ImageSubresourceRange::builder()
+                                .aspect_mask(vk::ImageAspectFlags::DEPTH)
+                                .layer_count(1)
+                                .level_count(1)
+                                .build(),
+                        )
+                        .build();
```
Ralith: Premature build

examples/src/lib.rs
```diff
@@ -591,45 +552,45 @@ impl ExampleBase {
...
                 .bind_image_memory(depth_image, depth_image_memory, 0)
                 .expect("Unable to bind depth image memory");
+
             record_submit_commandbuffer(
                 &device,
                 setup_command_buffer,
                 present_queue,
-                &[vk::PipelineStageFlags::BOTTOM_OF_PIPE],
+                &[],
+                //                &[vk::PipelineStageFlags::BOTTOM_OF_PIPE],
```
Ralith: Any reason we're keeping this around?

examples/src/lib.rs
```diff
@@ -641,35 +602,26 @@ impl ExampleBase {
...
+                    vk::ImageSubresourceRange::builder()
+                        .aspect_mask(vk::ImageAspectFlags::DEPTH)
+                        .level_count(1)
+                        .layer_count(1)
+                        .build(),
+                )
+                .image(depth_image)
+                .format(depth_image_create_info.format)
+                .view_type(vk::ImageViewType::TYPE_2D)
+                .build();
```
Ralith: Premature build

coolmedium:
@Ralith Ah indeed I did miss some. 

For the sake of consistency I have tried to stick to the following rules:

- whenever a `&SomeInfo` is called for I pass a `&SomeInfoBuilder` and let the `Deref` implementation sort it out
- when a `&[SomeInfo]` is called for, I do `let some_infos = [SomeInfo.builder().build()];` and pass a reference to that.

I *think* that should do it?

coolmedium:
That `use std::ptr;` isn't necessary on my system but the CI seems to need it?

MaikKlein:
@coolmedium windows and mac seem to use it inside the `create_surface` functions.

Ralith:
> when a &[SomeInfo] is called for, and the SomeInfo at hand is suitable for using a builder, I define let some_infos = [SomeInfo.builder().build()]; and pass the slice &some_infos.

If the array will be passed directly to the Vulkan call that reads it (for example, when creating pipelines) you can do `let some_info = SomeInfo::builder(); device.foo(&[some_info.build()]);` for maximum safety.

coolmedium:
@Ralith Very cool, I inlined slices in arguments to `Device` methods and called `.build()` on those elements of the inlined slices. 

Thanks again.

coolmedium:
woo hoo!

PR #170, opened 2018-12-07
SPIR-V reading helper

Ralith:
SPIR-V is defined in terms of 4-byte words and does not have a defined endianness, and compiler and target endianness may differ. Furthermore, Vulkan seems to require SPIR-V supplied to the implementation to be 4-byte-aligned. As a result, reading SPIR-V in a safe and portable fashion is error-prone.

So far we have largely shied away from including nontrivial helper code in ash, but arguably most portable code using ash will want this function. Do we want to break precedent? Or should we introduce a separate, optional, utility crate for code that doesn't directly wrap some part of Vulkan?


MaikKlein:
I think that the driver does the endian conversion, so that you only need to check for the magic version in little/big endian to know if it is *sort of* valid.

I am currently looking through the spec to see if I can find more information. My reasoning is that cpu and gpu might have different endianess.

MaikKlein:
Wait the shader is compiled on the host architecture right? So I guess maybe we really should convert.

coolmedium:
If this is merged I'll use this helper in the examples.  

It does certainly *seem* like many applications would make use of such a helper.  If the concern about including it is about feature creep, or about breaking the "purity" of the crate, then I'd also be for some kind of utility crate, at the very least just to discourage users from writing ad-hoc versions of this function themselves.

MaikKlein:
I think this is common enough that it should be included, but we should probably move it into the util module.

Ralith:
> I think that the driver does the endian conversion, so that you only need to check for the magic version in little/big endian to know if it is sort of valid.

Appendix A states that:
> A SPIR-V module passed into vkCreateShaderModule is interpreted as a series of 32-bit words in
host endianness

so the application indeed must convert to host endianness.

> Wait the shader is compiled on the host architecture right?

SPIR-V can come from wherever the application likes, even dynamically supplied by a user of fetched over the network. I think it's most commonly built on the build host. While in theory shader compilers could be configured for the target, in practice this does not appear to be supported by shaderc, and in any case more advanced use cases may need to handle SPIR-V of statically unknown endianness.

Ralith:
This should be good to go; CI failures are spurious.

MaikKlein:
bors retry

MaikKlein:
bors retry

PR #172, opened 2018-12-08
Builder: fix &'a [c_void] to &'a [8u]

MatusT:
Fix https://github.com/MaikKlein/ash/issues/155. There is a lot of noise because I already used 1.0.0 stable version of rustfmt. 

Probably merge the https://github.com/MaikKlein/ash/pull/171 first and I'll rebase then.


MaikKlein:
bors try

PR #186, opened 2019-02-10
Ray tracing support

gwihlidal:
This PR adds support for VK_NV_ray_tracing! (and initial support for VK_EXT_descriptor_indexing, but more support coming for that).

One piece I'm unsure if you have a more suggested approach is what I added to ext/mod.rs for names (both of those extensions are very commonly used, and also required for ray tracing). Maybe there should be a names.rs or something which can implement a static string function for all names that don't warrant their own extension struct?

I have a local example I've been working on that has been testing against this code, as well as support for descriptor indexing. Originally, I was adding an nv_ray_tracing.rs example alongside triangle and texture, but ray tracing requires some specific extensions and also it doesn't need frame bindings, render states, etc. Additionally, this example only works on NV and with a Turing-class GPU, so I'm unsure if it should co-exist with the basic examples? Because of this, I'm going to first start with a fresh example and just get working to show everything needed - we can chat about merging it in after if desired.

These changes + my upcoming ones should take care of both VK_NV_ray_tracing and VK_EXT_descriptor_indexing from https://github.com/MaikKlein/ash/issues/178


MaikKlein:
> One piece I'm unsure if you have a more suggested approach is what I added to ext/mod.rs for names (both of those extensions are very commonly used, and also required for ray tracing). Maybe there should be a names.rs or something which can implement a static string function for all names that don't warrant their own extension struct?

I wanted to generate all the extension names from the vk.xml file anyway, I just haven't had the time to implement it.`CStr` can't be any associated constant just yet, so maybe for now we just do something like `vk::names::maintenance3()` as you did in this PR. I'll try to implement it this week.

> I have a local example I've been working on that has been testing against this code, as well as support for descriptor indexing. Originally, I was adding an nv_ray_tracing.rs example alongside triangle and texture, but ray tracing requires some specific extensions and also it doesn't need frame bindings, render states, etc. Additionally, this example only works on NV and with a Turing-class GPU, so I'm unsure if it should co-exist with the basic examples? Because of this, I'm going to first start with a fresh example and just get working to show everything needed - we can chat about merging it in after if desired.

Sounds good to me. Ideally we would have one big place with all the examples, they don't necessarily need to share code. My only concern is that I can't really maintain this example because I don't have a turing card.

I'll try to give you a proper review later today or tomorrow. 

Thanks!

ash/src/extensions/nv/ray_tracing.rs
```diff
@@ -0,0 +1,263 @@
...
+use RawPtr;
+
+#[derive(Clone)]
+pub struct RayTracing {
+    handle: vk::Device,
+    ray_tracing_fn: vk::NvRayTracingFn,
+}
+
+impl RayTracing {
+    pub fn new<I: InstanceV1_1, D: DeviceV1_1>(instance: &I, device: &D) -> RayTracing {
```
MaikKlein: RTX is available for Vulkan 1.0, so this should probably be `InstanceV1_0` and `DeviceV1_0`

ash/src/extensions/nv/ray_tracing.rs
```diff
@@ -0,0 +1,263 @@
...
+        let ray_tracing_fn = vk::NvRayTracingFn::load(|name| unsafe {
+            mem::transmute(instance.get_device_proc_addr(device.handle(), name.as_ptr()))
+        });
+        RayTracing {
+            handle: device.handle(),
+            ray_tracing_fn,
+        }
+    }
+
+    pub unsafe fn get_properties<I: InstanceV1_1>(
```
MaikKlein: Having convenience functions is okay I think. Users without `1.1` need to do this manually with the correct extensions.

ash/src/extensions/khr/mod.rs
```diff
@@ -15,3 +15,25 @@ mod wayland_surface;
 mod win32_surface;
 mod xcb_surface;
 mod xlib_surface;
+
+pub fn memory_requirements2_name() -> &'static std::ffi::CStr {
```
MaikKlein: All those extension names are not necessary any more when https://github.com/MaikKlein/ash/pull/188 is merged.

ash/src/extensions/ext/descriptor_indexing.rs
```diff
@@ -0,0 +1,10 @@
+use std::ffi::CStr;
+
+#[derive(Clone)]
```
MaikKlein: I don't think we have an extension in ash yet that doesn't contain any function pointers. I guess it is more consistent. @Ralith Any opinions? I am generally okay with "empty" extensions.

Ralith: Seems fine to me; as you say, it's nice for consistency. There are many empty `*Fn` already, anyway.

ash/src/extensions/khr/mod.rs
MaikKlein: #188 has been merged. You can now do `vk::KhrGetMemoryRequirements2Fn::name()`.

ash/src/extensions/ext/descriptor_indexing.rs
gwihlidal: Sounds great

gwihlidal:
Will update my stuff now that #188 is in! Cheers

gwihlidal:
@MaikKlein would it be possible to do a (hopefully quick) update to Vulkan 1.1.97.0? There was a lot of Turing churn and and I'm debugging a weird crash that could be from a broken FFI in the past. (1.1.97.0 is really stable now)

MaikKlein:
@gwihlidal https://github.com/MaikKlein/ash/pull/189 I'll review it tomorrow.

MaikKlein:
@gwihlidal #189 is now merged.

ash/src/extensions/nv/ray_tracing.rs
gwihlidal: Done!

ash/src/extensions/khr/mod.rs
gwihlidal: Done!

gwihlidal:
Works! The remaining issues were in my app, not this PR, so it's ready for merging.

image

This PR now gives full access to NV RT from rust :)

PR #189, opened 2019-02-13
Update to 1.1.97

MaikKlein:


MaikKlein:
Surprisingly quite a lot of structs have changed. [For example](https://github.com/MaikKlein/ash/pull/189/files#diff-06980beb7c96424568ed38b471ee57e9R32045) All of this seems correct though.

I need to properly go through those changes to see if we break anything.

PR #191, opened 2019-03-08
Pass constant size arrays by reference

kvark:
Note: this is a breaking change!
Some related reading: https://lkml.org/lkml/2015/9/3/428
This has been plaguing WebRender on gfx-rs port running Windows/Vulkan.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -1132,7 +1132,7 @@ pub trait DeviceV1_0 {
     unsafe fn cmd_set_blend_constants(
         &self,
         command_buffer: vk::CommandBuffer,
-        blend_constants: [f32; 4],
+        blend_constants: &[f32; 4],
     ) {
         self.fp_v1_0()
             .cmd_set_blend_constants(command_buffer, blend_constants);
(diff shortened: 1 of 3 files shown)


MaikKlein:
Interesting, thanks for the catch!

I also leave here the current implementation for it in radv. Indeed it is just used as a pointer.

```
void radv_CmdSetBlendConstants(
	VkCommandBuffer                             commandBuffer,
	const float                                 blendConstants[4])
{
	RADV_FROM_HANDLE(radv_cmd_buffer, cmd_buffer, commandBuffer);
	struct radv_cmd_state *state = &cmd_buffer->state;

	if (!memcmp(state->dynamic.blend_constants, blendConstants, sizeof(float) * 4))
		return;

	memcpy(state->dynamic.blend_constants, blendConstants, sizeof(float) * 4);

	state->dirty |= RADV_CMD_DIRTY_DYNAMIC_BLEND_CONSTANTS;
}
```

kvark:
Thank you! Any plans to release a version with this fix?

MaikKlein:
There are currently a few breaking changes in place, I'll push a new version after https://github.com/MaikKlein/ash/pull/183 is merged, which should be today. (hopefully)

kvark:
Great, thanks!

PR #194, opened 2019-03-15
Add Khronos doc links and other ergonomic cleanup

aloucks:
* Khronos doc links are now automatically generated in vk.rs
* Added doc links to all other non-generated functions
* Implemented std::error::Error for LoadingError
* Made EntryCustom public in addition to the Entry typedef so that
  rustdoc no longer ignores it
* Moved Entry::new into EntryCustom (non-breaking)
* Added EntryCustom::try_enumerate_instance_version. The EntryV1_1 trait
  is not implemented for any type. Even if it were, the function would
  panic for Vulkan 1.0 implementations
* Added entry and instance creation example to lib docs
* Updated the Display impl for vk::Result so that it matches
  other bitmask and enum conventions
* Removed lazy_static dependency because it was no longer being
  referenced


generator/src/lib.rs
```diff
@@ -85,13 +85,30 @@ named!(cfloat<&str, f32>,
...
+
+pub fn define_khronos_link_macro() -> Tokens {
+    quote! {
+        macro_rules! khronos_link (
+            ($name:expr) => {
+                Literal::string(&format!("<https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/{name}.html>", name=$name))
+            }
+        );
+    }
+}
```
Ralith: If this is going to be removed in the next commit, it might as well not be introduced.

generator/src/lib.rs
```diff
@@ -85,13 +85,30 @@ named!(cfloat<&str, f32>,
     terminated!(nom::float_s, char!('f'))
 );
 
+macro_rules! khronos_link (
+    ($name:expr) => {
+        Literal::string(&format!("<https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/{name}.html>", name=$name))
+    }
+);
```
Ralith: Why is this a macro?

ash/src/entry.rs
```diff
@@ -46,6 +44,22 @@ pub enum LoadingError {
...
+impl fmt::Display for LoadingError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "LoadingError::{:?}", self)
+    }
+}
+
+impl Error for LoadingError {
+    fn description(&self) -> &str {
+        "LoadingError"
+    }
```
Ralith: This method is deprecated; we shouldn't implement it explicitly.

ash/src/entry.rs
```diff
@@ -46,6 +44,22 @@ pub enum LoadingError {
     LibraryLoadError(String),
 }
 
+impl fmt::Display for LoadingError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "LoadingError::{:?}", self)
```
Ralith: The `Display` impl should not be equal to the `Debug` impl; people can just use `Debug` for that. Instead, it should provide a written description of the error. In this case, probably just print the wrapped string?

ash/src/entry.rs
```diff
@@ -46,6 +44,22 @@ pub enum LoadingError {
...
+}
+
+impl Error for LoadingError {
+    fn description(&self) -> &str {
+        "LoadingError"
+    }
+
+    fn cause(&self) -> Option<&Error> {
+        None
+    }
```
Ralith: This method is also deprecated, and this definition is redundant to the default definition anyway.

generator/src/lib.rs
```diff
@@ -1237,7 +1254,7 @@ pub fn generate_result(ident: Ident, _enum: &vkxml::Enumeration) -> Tokens {
         impl fmt::Display for #ident {
             fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
                 let name = match *self {
-                    #(#notation2),*,
+                    #(#display_notation),*,
```
Ralith: `Error::description` is deprecated in favor of `to_string`, so this represents a step backwards in API friendliness to idiomatic code. Users who want the constant's name instead of it's description can already use the `Debug` impl.

aloucks: The debug impl does not represent the constant name, which is why I made the change. The debug impl looks like `Result(1000000)`

generator/src/lib.rs
aloucks: It started out as something more complicated. I also thought it might need to be embedded into vk.rs like some of the others, but that didn't turn out to be needed. Reducing to a function should be doable.

ash/src/entry.rs
aloucks: I was following the precedent set with the other error types.

ash/src/entry.rs
aloucks: Again, only following precedent with how the other were implemented. I (possibly wrongly?) assumed that smaller changes would go over better than sweeping refactoring.

generator/src/lib.rs
aloucks: It was introduced after appveyor failed. The commit message serves as a reminder as to why it was disabled and potentially enabling it after sorting out how to get vulkan appveyor. I can squash and force push if it's a problem.

generator/src/lib.rs
aloucks: Many of the constants are not covered by the Debug impls. I'd like to tackle that too, but it's a bigger change and I wanted to get these rolling out.

ash/src/entry.rs
aloucks: As stated below, the debug impl does not render the constant name for some things (specifically `Result`). The name of the constant is more valuable IMO, as you can plug that into google. I do like having the comments from the xml added to the rustdoc though.

I would prefer to show the constant name for display and the type and value for debug.

e.g.

Display: `VK_SUCCESS`
Debug: `Result(0)`

generator/src/lib.rs
Ralith: I'm not sure what you're referring to. I'm talking about the function definition that this comment is on.

ash/src/entry.rs
Ralith: You can update those to current conventions if you like, but either way we shouldn't be introducing new code using obsolete patterns.

ash/src/entry.rs
Ralith: This comment refers to `LoadingError`, not `vk::Result`. There are no constants here.

ash/src/entry.rs
Ralith: You're not obliged to do sweeping refactoring, but we shouldn't introduce new code using obsolete patterns.

generator/src/lib.rs
Ralith: Whoops, looks like I was looking at an old private branch of mine by accident. Regardless, `description` should not be used. For consistency with Rust error enum conventions, the `Debug` impl should display the symbolic name (or a numeric value in newtype struct form if no name is found), and the `Display` impl should provide a written description as it currently already does.

ash/src/entry.rs
aloucks: I followed the convention that was established by the crate's author:

```rust
impl fmt::Display for InstanceError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        write!(f, "InstanceError::{:?}", self)
    }
}
```

generator/src/lib.rs
aloucks: It's been removed

generator/src/lib.rs
aloucks: Now a function

ash/src/entry.rs
aloucks: Also removed

generator/src/lib.rs
aloucks: Error::description impls are removed

ash/src/entry.rs
Ralith: In this case, I believe that's a poor example to follow. `Display` impls for errors in rust should always produce a written description of the error; this convention is well established. In this case, unlike `InstanceError`, it should be completely trivial to do so since there's just the one wrapped string anyway.

generator/src/lib.rs
```diff
@@ -1185,38 +1213,27 @@ pub fn generate_enum<'a>(
...
+            vkxml::EnumerationElement::Enum(ref constant) => constant.name.as_str(),
             _ => {
                 return None;
             }
         };
 
         let variant_ident = variant_ident(&_enum.name, variant_name);
         Some(quote! {
-            #ident::#variant_ident => Some(#notation)
+            #ident::#variant_ident => Some(stringify!(#variant_ident))
```
Ralith: I stand by my judgement that this is a step backwards. Especially for error enums, `Debug` impls should display symbolic names (where available), and `Display` impls should display written descriptions for human-friendly error output. The descriptions here are taken directly from the Vulkan spec and should be plenty googlable, and will be more convenient in the common case of debugging.

Consistency is nice, but it's not sufficient reason to introduce a regression. If anything, we should move other parts of the system towards the more idiomatic pattern, though the idiom is most well-established for error enums and `Result` is the only one of those.

aloucks: The doc comments from the spec are in the doc comments of the generated output. I disagree that this is in any way a "regression". 

The error component of the result is an enum. When I print the enum I want to know the name of the enum, not a sentence describing it. The debug impls _do not print the name of the enum_, they print the struct type and the containing integer. This is helpful in some cases is, but there absolutely needs to be a way to see the **name** of the enum.

ash/src/entry.rs
aloucks: It's not one wrapped string. The other error type contains a `Vec`.

```rust
pub enum InstanceError {
    LoadError(Vec<&'static str>),
    VkError(vk::Result),
}
```  

I've changed them both to print the inner value though, based on this feedback.

aloucks: Personally, I think the author's way is better overall. I'm happy to rip off the last commit to revert this:

https://github.com/MaikKlein/ash/pull/194/commits/8cb149e94b19aff9e5f156461b60412740586d12

generator/src/lib.rs
Ralith: > When I print the enum I want to know the name of the enum, not a sentence describing it

In practically all Rust code, this is what `Debug` impls on enums do; in particular, this is the behavior of a derived `Debug` impl on an enum.

> The debug impls do not print the name of the enum, they print the struct type and the containing integer

This is easily corrected; users who actually need the integer can still access it with `as_raw`.

ash/src/entry.rs
Ralith: > It's not one wrapped string

This thread is discussing your change to `LoadingError`, which is always exactly one wrapped string.

ash/src/entry.rs
```diff
@@ -54,33 +62,28 @@ pub enum InstanceError {
...
-impl Error for InstanceError {
-    fn description(&self) -> &str {
-        "InstanceError"
-    }
-
-    fn cause(&self) -> Option<&Error> {
-        if let &InstanceError::VkError(ref err) = self {
-            return err.source();
+        match self {
+            InstanceError::LoadError(e) => write!(f, "{}", e.join("\"n")),
```
Ralith: This is a strange delimiter. Perhaps you meant `"\n"`? I'd prefer `": "`, personally; sticking newlines in a `Display` impl tends to break peoples' formatting.

ash/src/entry.rs
```diff
@@ -54,33 +62,28 @@ pub enum InstanceError {
...
+            InstanceError::VkError(e) => write!(f, "{}", e),
```
Ralith:
```suggestion
            InstanceError::VkError(e) => e.fmt(f),
```

ash/src/entry.rs
aloucks: Good eye! This is fixed

generator/src/lib.rs
aloucks: These are not *rust* enums, they're unit structs. The formatting behavior is different.

ash/src/entry.rs
aloucks: I prefer the consistency with the other impls.

generator/src/lib.rs
Ralith: They're newtype structs whose function is to emulate enums across an ABI that doesn't guarantee a closed set of possible values. In other words, they're enums except that they may have some statically unknown cases.

ash/src/entry.rs
aloucks: They should be consistent.

ash/src/entry.rs
```diff
@@ -54,33 +62,28 @@ pub enum InstanceError {
...
 }
 
+impl Error for InstanceError {}
+
 #[allow(non_camel_case_types)]
 pub trait EntryV1_0 {
     type Instance;
     fn fp_v1_0(&self) -> &vk::EntryFnV1_0;
     fn static_fn(&self) -> &vk::StaticFn;
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkCreateInstance.html>"]
```
MaikKlein: I wish there would be a link to the newest docs like `vulkan/specs/newest`, maybe we should open an issue for this.

ash/src/entry.rs
```diff
@@ -194,21 +237,45 @@ impl<L> EntryCustom<L> {
...
+    ///         let major = vk_version_major!(version);
+    ///         let minor = vk_version_minor!(version);
+    ///         let patch = vk_version_patch!(version);
+    ///     },
+    ///     // Vulkan 1.0
+    ///     None => {},
+    /// }
+    /// # Ok(()) }
+    /// ```
+    pub fn try_enumerate_instance_version(&self) -> VkResult<Option<u32>> {
```
MaikKlein: I think we should just add this to the `Entry1_1` trait, as a convenience function. The function pointer is already in there.

The type signature would also be swapped I think `Option<VkResult<u32>`, makes more sense to me.

generator/src/lib.rs
```diff
@@ -85,13 +87,21 @@ named!(cfloat<&str, f32>,
...
+        name = name
+    ))
+}
+
 pub fn define_handle_macro() -> Tokens {
     quote! {
         #[macro_export]
         macro_rules! define_handle{
-            ($name: ident, $ty: ident) => {
+            ($name: ident, $ty: ident, $doc_link: meta) => {
```
MaikKlein: I think doc links should be optional. Those macros are sometimes used outside of ash and I don't think the user wants to always specify the links manually.

generator/src/lib.rs
MaikKlein: Can you give me a diff of how the new output vs the old output would look like?

ash/src/entry.rs
aloucks: I thought about removing the `Entry1_1` trait but thought it might be too dramatic :) It's not actually implemented *anywhere*. Even if it were implemented, it's not safe to call because it will panic for Vulkan 1.0 drivers.

I specifically used  `VkResult<Option<u32>>` so that `?` syntax would work. This is also aligned with other `try_` style functions in the standard library (they always return `Result`).

generator/src/lib.rs
aloucks: https://github.com/MaikKlein/ash/pull/194/files#diff-06980beb7c96424568ed38b471ee57e9L40632

generator/src/lib.rs
aloucks: Ah, good point. I've updated this to make the doc link an optional parameter.

ash/src/entry.rs
MaikKlein: `?` also works with option. We should still implement `Entry1_1`, I probably just forgot to add an `impl`. It doesn't really matter if it safe to call. No >1.0 functions are. Although this is a bit of a special case and I think a convenience function that checks if it is available seems reasonable.

I think `try_xxx` is still okay for options, `Try` is implemented for both Option and Result, unless the style guide says something more specific about this.

The question is what is more convenient for the user?

Ralith: `?` works with `Option`in functions returning `Option`. Functions returning an error type constructible from `vk::Result` will not be able to make much use out of `?` with a function returning `Option`, so I think the current strategy makes sense.

This is a bit academic, because the function is documented as having no error cases.

MaikKlein: Yeah I guess it doesn't really make much of a difference. 

```
ctx.try_enumerate_instance_version().map_err(...)?.unwrap_or(1.0);
```

Ralith: Actually, why not make a `EntryV1_0` function that just does that internally? If the function's missing it's by definition 1.0, so this simplifies the interface without losing any expressiveness.

aloucks: >I think try_xxx is still okay for options, Try is implemented for both Option and Result, unless the style guide says something more specific about this.

[NoneError](https://doc.rust-lang.org/std/option/struct.NoneError.html) isn't stabilized. As @Ralith stated, you can't user `?` on `Option` if the function returns a `Result`.

MaikKlein: @Ralith it probably also includes the patch version. Not sure if that is ever useful but its trivial to implement for the user anyway. I just used `1.0` above but it is actually an `u32`.

As I said either approach is fine by me, but 1.1 functions should stay in `1_1`, this is most likely where the user expects them to be. You also can only default implement them in `1_1`.

I would also be okay for `try_enumerate_instance_version` to just be a method as it is now, just reuse the function pointer that was already loaded and implement the `Entry1_1` trait.

aloucks: >just reuse the function pointer that was already loaded and implement the Entry1_1 trait.

It needs to check with `get_instance_proc_addr` because the `EntryFnV1_0` implementation stores a dummy function that panics (i.e. it will never be null).

MaikKlein: Sorry I forgot about the panic functions. I guess this is fine then?

ash/src/entry.rs
aloucks: @MaikKlein Do you have a preference here? Does this need further changes?

MaikKlein: I think we should just print the string. I'll add my comments on Debug/Display in the other discussion.

generator/src/lib.rs
MaikKlein: I think it is already implemented correctly. Display is for userfacing output, I have to admit that I did not follow this rule this when I started ash. Still I believe we could improve the Debug output though. `Result(0)` is not super useful in practice. I think we should add something like `Result::Success(0)`, if possible. @Ralith Thoughts?

Ralith: I think the ideal approach is to match on the code and display the name if known, and the numeric value otherwise, similar to what's done in the `Display` impls of other Vulkan enums. I think I made a mistake in making those be `Display` impls in the first place, considering rust conventions.

aloucks: Do you want me to just pull out the changes with `Display` and `Result` so that it can be handled later?

MaikKlein: Yeah I think that would be best, we address this later.

ash/src/entry.rs
aloucks: > I've changed them both to print the inner value though, based on this feedback.

```rust
impl fmt::Display for LoadingError {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match self {
            LoadingError::LibraryLoadError(e) => write!(f, "{}", e),
        }
    }
}```

generator/src/lib.rs
aloucks: The `Display` changes were reverted in https://github.com/MaikKlein/ash/pull/194/commits/d52f2c9b3e59ddde9ee44f385210327d99392689

Is there anything else that needs to be changed?

PR #196, opened 2019-03-20
Rewrite builder in the readme

MaikKlein:


MaikKlein:
fixes https://github.com/MaikKlein/ash/issues/192

MaikKlein:
bors try

MaikKlein:
Yeah I probably should add that.

README.md
```diff
@@ -72,15 +72,46 @@ let desc_alloc_info = vk::DescriptorSetAllocateInfo {
...
+        mut self,
+        queue_create_infos: &'a [DeviceQueueCreateInfo],
+    ) -> DeviceCreateInfoBuilder<'a> {...}
+    //...
+```
+
+Every reference has to live as long as the builder itself. Builders implement `Deref` targeting their corresponding Vulkan struct, so references to builders can be passed directly
+to Vulkan functions.
+
+Calling `.build()` will **discard** that lifetime because Vulkan structs use raw pointers internally. This should be avoided as much as possible because this can easily lead to dangling pointers. If `.build()` has to be called, it should be called as late as possible. The reason for this is that the lifetime of [temporaries](https://doc.rust-lang.org/reference/expressions.html#temporary-lifetimes) will be extended to the enclosing block.
```
Ralith:
```suggestion
Calling `.build()` will **discard** that lifetime because Vulkan structs use raw pointers internally. This should be avoided as much as possible because this can easily lead to dangling pointers. If `.build()` has to be called, it should be called as late as possible. Lifetimes of [temporaries](https://doc.rust-lang.org/reference/expressions.html#temporary-lifetimes) are extended to the enclosing statement, ensuring they are valid for the duration of a Vulkan call occurring in the same statement.
```

MaikKlein: I just sort of quoted it from the reference
> the temporary is created with the lifetime of the enclosing block instead

Enclosing block here would be the function call itself.

Ralith: That quote is taken severely out of context. The part you cut off is "When a temporary value expression is being created that is assigned into a let declaration, however, ..."; it is describing an exception to the usual enclosing-statement rule.

MaikKlein: I mean you could write it as  `let _ = foo(&temp())`. I guess yours is more clear though, I just merge it in.

Ralith: In that example, you are assigning the return value of `foo(...)`, not the temporary. The lifetime of the temporary is only that statement.

PR #207, opened 2019-05-09
Fix typo

felixrabe:
Not sure whether this should actually read *appended* or *prepended*. Or maybe it is to make sure the chain ends in a pep? 🤷‍♂️

PR #208, opened 2019-05-09
Typo

felixrabe:
Like #207, but the opposite.


Ralith:
Thanks! For future reference, when amending a PR you can just force-push to the pre-existing branch to save having to make a new one.

PR #209, opened 2019-05-11
Vulkano is now vulkano-rs/vulkano

felixrabe:


felixrabe:
`77fff85` is my new favorite Git commit prefix :)

Issue #205, opened 2019-05-07
VulkanSDK 1.1.92.1 breaks examples


NLincoln:
# The issue

Running the examples with a vulkansdk >= 1.1.92.1 on macOS renders a black screen. Nothing on stdout / stderr. I have tested with both debug mode and release mode configuration.

For completeness, I have also checked every release after 1.1.92.1. They render a black screen.

# How to reproduce:

- `git clone` this repo
- download vulkansdk 1.1.92.1 from [this link](https://vulkan.lunarg.com/sdk/home#mac)
- set up the env vars using the README instructions
- run the examples, black screen

You can verify that everything should be working by downloading version 1.1.85.0, resetting the env vars to use that, and the examples should work.

# My environment

- latest master branch of this repo
- macOS 10.12
- rustc 1.34 stable

NLincoln:
Note that it's unlikely this is a duplicate of #149, since the latest vulkansdk at that point was 1.1.85.0, which I've confirmed to work.

NLincoln:
I should also note that the sdk-provided example (vkcube) _does work_ in the latest release of the vulkansdk for me, so it appears to be a problem related to ash.

MaikKlein:
Thanks for the report. Sadly I don't have immediate access to a mac atm. I'll try to reproduce it as soon as I can.

unknownue:
This problem reproduce on my macbook too. 
It does take me some time to experiment what is the problem.
And I found that after changing all the requested memory flags from `vk::MemoryPropertyFlags::HOST_VISIBLE` to `vk::MemoryPropertyFlags::HOST_VISIBLE | vk::MemoryPropertyFlags::HOST_COHERENT`, the triangle finally appeared on screen.

Following is what the `vulkaninfo` command prints on my macbook with SDK version 1.1.101.

	VkPhysicalDeviceMemoryProperties:
	=================================
	memoryHeapCount       = 1
	memoryHeaps[0] :
		size          = 4294967296 (0x100000000) (4.00 GiB)
		flags:
			VK_MEMORY_HEAP_DEVICE_LOCAL_BIT
	memoryTypeCount       = 3
	memoryTypes[0] :
		heapIndex     = 0
		propertyFlags = 0x1:
			VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
	memoryTypes[1] :
		heapIndex     = 0
		propertyFlags = 0xb:
[...]

PR #212, opened 2019-05-20
Fix black screen for examples on macOS platform

unknownue:
The memory types may vary between different versions of MoltenVK.
This solution just changes all the memory request flags to both `HOST_VISIBLE` and `HOST_COHERENT` to fix the delay of memory data transfer.


unknownue:
#205

unknownue:
I have tested this commit with different SDKs from v1.1.82 to v1.1.106 on macOS 10.14.4. The triangle appears on screen successfully.

NLincoln:
I've also tested this, and it fixes the issue 🙌. Thank you!

PR #215, opened 2019-05-25
Replace mem::uninitialized with mem::zeroed

### LINKED ISSUE #214: Replace mem::uninitialized
See https://gankro.github.io/blah/initialize-me-maybe/, an easy global grep replace to `mem::zeroed()` or `Default` would probably suffice.

aloucks:
Fixes #214

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -43,7 +43,7 @@ pub trait DeviceV1_1: DeviceV1_0 {
         local_device_index: u32,
         remote_device_index: u32,
     ) -> vk::PeerMemoryFeatureFlags {
-        let mut peer_memory_features = mem::uninitialized();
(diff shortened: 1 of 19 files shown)


ash/src/device.rs
```diff
@@ -1510,7 +1510,7 @@ pub trait DeviceV1_0 {
         size: vk::DeviceSize,
         flags: vk::MemoryMapFlags,
     ) -> VkResult<*mut c_void> {
-        let mut data: *mut c_void = mem::uninitialized();
+        let mut data: *mut c_void = mem::zeroed();
```
MaikKlein: We should probably use `std::ptr::null_mut` here.

aloucks: Done!

aloucks:
@MaikKlein Are other changes needed?

PR #217, opened 2019-05-25
Make handle and bitflags constructor functions const

### LINKED ISSUE #187: const fn for BitFlags (where it's currently possible)
In my vulkan application, I use a constant struct to define my engine base configuration. The thing is, it is not current possible to combine the ash bitflags in constants or static expressions. For example, right now this code do not work:

```rust 
const REQUIRED_QUEUE_FLAGS = vk::QueueFlags::COMPUTE | vk::QueueFlags::GRAPHICS;
```

That's because (as far as I know) `BitOr` cannot use `const fn`. I guess that's a Rust limitation. So I tried doing this instead:

```rust
const REQUIRED_QUEUE_FLAGS =  vk::QueueFlags::from_raw(0b1 | 0b10),  // GRAPHICS / COMPUTE 
```

Which didn't work either because the `vk_bitflags_wrapped` do not declare `from_raw` as a const fn. So I cloned the repo and declared `empty`, `all`, `from_raw` and `as_raw` as `const fn` to see if anything would explode. So far everything works fine.

Would it be fine to merge my changes in the main branch?

aloucks:
Fixes #187

### DIFF
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -85,19 +85,19 @@ macro_rules! vk_bitflags_wrapped {
         }
         impl $name {
             #[inline]
-            pub fn empty() -> $name {
+            pub const fn empty() -> $name {
                 $name(0)
             }
             #[inline]
-            pub fn all() -> $name {
+            pub const fn all() -> $name {
                 $name($all)
             }
             #[inline]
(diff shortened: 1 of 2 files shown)


MaikKlein:
Also thanks for all the PRs. Very much appreciated!

bors r+

PR #222, opened 2019-07-09
Update VulkanHeaders to 1.1.117

MaikKlein:
The generator needed a few changes. *Note* the generator really needs a cleanup.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -731,9 +731,9 @@ fn generate_function_pointers<'a>(
             let ident = cmd.name.as_str();
             if !fn_cache.contains(ident) {
                 fn_cache.insert(ident);
-                return true;
+                true
             } else {
-                return false;
+                false
             }
         })
         .collect();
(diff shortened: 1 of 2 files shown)


msiglreith:
imageless framebuffers \o/

generator/src/lib.rs
```diff
@@ -378,6 +379,11 @@ pub fn platform_specific_types() -> Tokens {
         // Opage types
         pub type ANativeWindow = c_void;
         pub type AHardwareBuffer = c_void;
+        // I guessed the type definition here because they are apperantly under NDA, inside the
+        // Stadia SDK.
+        pub type GgpStreamDescriptor = u32;
+        pub type GgpFrameToken = u32;
```
Ralith: To avoid unnecessary breaking changes, perhaps we should not expose these yet? NDA parties can always define the struct themselves in the mean time.

MaikKlein: I don't know, I guess we could filter out NDA functions/types but that feels like needless work. I just don't understand why they would include NDA stuff in a public api. And then some people would want to use those extensions but can't and I don't want to support multiple branches for NDAs. I don't think that people would be allowed to fix an incorrect definition because of the NDA.

I'd really just want an `#[unstable]` tag for things like that in Rust. 

vk.xml is really a bit annoying outside of c/cpp.

Because this is the only NDA case, I think I'll just add an doc comment that those definitions won't adhere to semver, and I'll see if I can talk to someone at google.

PR #233, opened 2019-08-25
Address all the clippy lints

MaikKlein:


generator/src/lib.rs
```diff
@@ -859,6 +859,7 @@ fn generate_function_pointers<'a>(
             }
         }
         impl #ident {
+            #[allow(clippy::cognitive_complexity)]
```
Ralith: Could we just apply this to the whole module instead? It's not like there should be any handwritten code of meaningful size here at all.

generator/src/lib.rs
```diff
@@ -441,6 +441,22 @@ impl quote::ToTokens for ConstVal {
...
+        .enumerate()
+        .fold(String::new(), |mut acc, (idx, next)| {
+            if idx != 0 && idx % count == 0 {
+                acc.push(symbol);
+            }
+            acc.push(next);
+            acc
+        });
+    number.chars().rev().collect()
+}
```
Ralith: Would it make sense to explicitly disable this lint for vk.rs instead, since none of vk.rs is meant for direct human consumption anyway?

aloucks: I actually really like formatting the number to make it easier to read. It makes it easier to grok when debugging.

generator/src/lib.rs
MaikKlein: Yeah true, I moved it to the top.

Issue #226, opened 2019-08-04
Raw pointers not exposed in extension loading structs


Didgy74:
The raw function pointers are not exposed for extension-loader structs, among them are `ash::extensions::khr::Surface` and `ash::extensions::khr::Swapchain`.

I see that these structs contain underlying structs for the raw function pointers (i.e `vk::KhrSurfaceFn`), making the structs public and providing getters to non-mut references for them can work as a quick fix.

Ralith:
I believe this was fixed by #234.

PR #234, opened 2019-08-26
Expose raw function pointers in extensions

Didgy74:
Disclaimer: This pull request is not ready to be merged into master. This is a proposal for an interface addition

So in order to remedy issue #226, I've made this pull request. I've implemented an interface proposal for `ash::extensions::khr::Surface` and `ash::extensions::khr::Swapchain` as example code that needs to be reviewed.

The idea is to add a `raw_fp()` function to every extension struct (i.e Surface) that returns a non-mutable reference to the inner structure that holds the function pointer(s). In the case of Surface, this is the `ash::vk::KhrSurfaceFn` struct.

Additonally, we could also add `instance_handle()` and `device_handle()` functions where applicable, since the structures that hold these handles, have grabbed the function pointers specifically for those handles.

Example usage:
```rust
let mut swapchain_image_count: u32 = 0;

unsafe {
    let result = swapchain_loader.raw_fp().get_swapchain_images_khr(
        swapchain_loader.device_handle(), 
        swapchain, 
        &mut swapchain_image_count, 
        std::ptr::null_mut::<vk::Image>());
    assert_eq!(result, vk::Result::SUCCESS);
}

let mut images = // Allocate custom buffer based on swapchain_image_count

unsafe {
    let result = swapchain_loader.raw_fp().get_swapchain_images_khr(
        swapchain_loader.device_handle(), 
        swapchain, &mut swapchain_image_count, 
        images.as_mut_ptr());
    assert_eq!(result, vk::Result::SUCCESS);
}
```

If the interface seems fine, I'll go ahead and implement it for all the extension structs except `experimental`

### DIFF
--- a/ash/src/extensions/ext/debug_marker.rs
+++ b/ash/src/extensions/ext/debug_marker.rs
@@ -64,4 +64,8 @@ impl DebugMarker {
         self.debug_marker_fn
             .cmd_debug_marker_insert_ext(command_buffer, marker_info);
     }
+
+    pub fn fp(&self) -> &vk::ExtDebugMarkerFn {
+        &self.debug_marker_fn
+    }
 }
--- a/ash/src/extensions/ext/debug_report.rs
+++ b/ash/src/extensions/ext/debug_report.rs
@@ -58,4 +58,12 @@ impl DebugReport {
             _ => Err(err_code),
         }
     }
+
+    pub fn fp(&self) -> &vk::ExtDebugReportFn {
+        &self.debug_report_fn
+    }
+
+    pub fn instance(&self) -> vk::Instance {
+        self.handle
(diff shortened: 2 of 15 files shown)


Didgy74:
Personally I think your second naming suggestion looks better because it follow the naming scheme of `ash::Instance::fp_v1_0()` better in my opinion.

So, should I go ahead and implement the interface: `.fp()`, `.instance()` and `.device()` ?

Didgy74:
Update: I've implemented the interface for all extensions except the ones in `experimental`. The code is ready to be pulled. The pull request does not contain an interface breaking change.

MaikKlein:
Looks good to me, thanks for the PR!

PR #235, opened 2019-09-02
Replaces Vec in return types with generic FromIterator

nicbn:
**This is a breaking change.**

This replaces every occurence of `Vec<T>` in a return type with a generic `FromIterator<T> + DerefMut<Target = [T]>`.
That way, the library is way more flexible, while still not needing mutable slice parameters. If one knows the size of the collection at compile time, for example, `ArrayVec` may be used. If one haves a guess, `SmallVec` may be used, and so on.
The behaviour should be unchanged if one chooses to use `Vec`.

With this change, the only structs that implicitly allocate are `Entry` and `InstanceError`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,12 +36,15 @@ let instance = entry.create_instance(&create_info, None)
 ### `Vec<T>` instead of mutable slices
 
 ```Rust
-pub fn get_swapchain_images(&self,
-                            swapchain: vk::SwapchainKHR)
-                            -> VkResult<Vec<vk::Image>>;
-let present_images = swapchain_loader.get_swapchain_images_khr(swapchain).unwrap();
+pub fn get_swapchain_images<
+    T: FromIterator<vk::Image> + DerefMut<Target = [vk::Image]>,
+>(
+    &self,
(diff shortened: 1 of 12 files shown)


MaikKlein:
Hey thanks for the PR. I'll have a closer look at those changes in a few days.

nicbn:
Ok on a second thought there is a big problem with that approach. Nothing guarantees that `FromIterator` will create a slice with the same length as the iterator.
One option would be to use `assert`, but that would add overhead and the users should be able to hand this kind of situation themselves before calling the function.

I think the best solution is to just pass mutable slice as parameter. It's not the most idiomatic thing, and will require more work on the user end, but is the solution which matches the underlying API, causes no overhead and gives the user the most freedom and control.

MaikKlein:
I sort of designed ash with allocators in mind. The problem is that I didn't expect them to take that long to be stabilized.  I think that would be the cleanest way for ash.

Vulkan doesn't allocate in the hot path so it should not be a problem in practice and you can always fall back to the low level fps if you want to.

PR #245, opened 2019-10-26
Helper macro for including precompiled SPIR-V

Ralith:
Should help suppress the temptation to court UB via naive include_bytes. Implementation placed in the root because macros are always imported via the root and it'd be weird for the returned type
to not be beside it.


ash/src/lib.rs
```diff
@@ -52,6 +52,32 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
 }
 
+/// Include correctly aligned and typed precompiled SPIR-V
+///
+/// Does not account for endianness mismatches between the SPIR-V file and the target. See
+/// `util::read_spv` for a more general solution.
+#[macro_export]
+macro_rules! include_spv {
+    ($path:expr) => {
+        &$crate::util::Align4(*include_bytes!($path)) as &$crate::Spirv
```
MaikKlein: It is a bit unclear to me why this is useful. I don't think this is what the user would expect. 

It is aligned correctly if the len of the byte array is `[u8; LEN], LEN % 4 == 0` right? And this invariant can't break because it would not be legal spir-v otherwise. Loading anything that violates this is not correct and we should probably panic instead.

Ralith: > It is a bit unclear to me why this is useful. I don't think this is what the user would expect.

What specifically are you referring to?

> It is aligned correctly if the len of the byte array is [u8; LEN], LEN % 4 == 0 right?

No, alignment is a property of where it's stored, not its length.

> we should probably panic

That would preclude use in a const context, which is a goal.

MaikKlein: I am confused. Why do you think it has to be aligned? The vulkan spec only states that the length has to be a multiple of 4 (because 1 spirv op is u32).

Ralith: To begin with, `vk::ShaderModuleCreateInfo` takes a `*const u32`, and the builder takes a `&[u32]`, so at a baseline this is more ergonomic. Moreover, [the spec](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkShaderModuleCreateInfo.html) says:

>  pCode must be a valid pointer to an array of codeSize / 4 uint32_t values

An unaligned `&[u8]` violates that. Under [the C11 memory model](http://port70.net/~nsz/c/c11/n1570.html#6.3.2.3p7), even *constructing* a `*const u32` from unaligned memory is UB.

We've discussed this previously in https://github.com/MaikKlein/ash/pull/117 and https://github.com/MaikKlein/ash/pull/170.

MaikKlein: You are totally right.

MaikKlein: I'll look at it more closely tomorrow. But it seems to me that this copies the spv to the stack right?

Ralith: In principle it could; you can put it in a `const` or even a `static` if you're worried that the blindingly obvious optimization might not happen and your shaders are extremely large.

Ralith: Actually, given that it has `'static` lifetime, that might not be a risk at all.

parasyte: I believe you're right, the spv will only make it to the stack if it is copied.

Ralith: It looks like dereffing `include_bytes` does *technically* count as a copy when not in const context, so probably best to encourage people to stick it in a `const`, though I imagine it'll optimize down the same either way.

MaikKlein:
I don't fully understand the purpose of this. We can already use `read_spirv` with include_bytes to load shaders. This implementation copies to the stack which is not great.

Ralith:
This can be used in a `const` context, which should not copy at all. In the cases where it does, LLVM will likely optimize it away. By contrast `read_spirv` always copies to the heap. The obscure nature of the `Cursor` API will further tend to encourage people without a firm grasp on UB to just transmute the slice when using `include_bytes`, so this provides a more convenient alternative.

MaikKlein:
> This can be used in a const context, which should not copy at all.

Then it will do the stack allocation in a const context, which is probably also not what you want.

> The obscure nature of the Cursor API will further tend to encourage people without a firm grasp on UB to just transmute the slice when using include_bytes, so this provides a more convenient alternative.

I think we should rather improve the API of `read_spv` then. I don't think many people will include their shaders in the binary, and if they do 1 allocation doesn't matter. And at one point `read_spv` might also become a const fn, although that probably takes a while.

rust also really should have a `include_u32` or `include_bytes_aligned`, one can also implement this in a proc macro.

Ralith:
> Then it will do the stack allocation in a const context, which is probably also not what you want.

If you don't believe that will optimize it, you can use it to initialize a static and get a hard guarantee.

> I think we should rather improve the API of read_spv then.

The API of that function is correct for what it's meant for, which is not fixing up `include_bytes`.

> I don't think many people will include their shaders in the binary, and if they do 1 allocation doesn't matter.

People keep doing this and getting it wrong, and if the allocation doesn't matter then I'm not sure what the problem with the macro is. Some SPIR-V won't blow your stack, in the unlikely event that it ends up there to begin with.

> rust also really should have a include_u32 or include_bytes_aligned, one can also implement this in a proc macro.

We've herein implemented this in a declarative macro, which has the substantial advantage of not requiring you to compile syn.

aloucks:
I think this variant of the macro will work in a const context and should also guarantees against a copy on the stack:

```rust
#[macro_export]
macro_rules! include_spv {
    ($path:expr) => {{
        static ALIGNED: &'static $crate::util::Align4<[u8]> = &$crate::util::Align4(*include_bytes!($path));
        ALIGNED
    }};
}
```

And the example would be:

```rust
let vertex_code: &Spirv = include_spv!("../../shader/triangle/vert.spv");
let vertex_shader_info = vk::ShaderModuleCreateInfo::builder().code(vertex_code);
```

Ralith:
> I think this variant of the macro will work in a const context

It doesn't, which is exactly why you needed to replace `const` with `let` in the example. I'm not categorically opposed to this approach, but it seems like a reduction in generality for dubious benefit.

PR #248, opened 2019-11-11
Make more structs derive PartialEq/Eq/Hash

swooster:
Adds equality-related traits to VkClearRect, VkOffset2D, VkOffset3D,
VkRect2D and VkSurfaceFormatKHR. Fixes a typo preventing said traits
from being applied to VkExtent2D.

### DIFF
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -7926,7 +7926,7 @@ impl ::std::default::Default for BaseInStructure {
     }
 }
 #[repr(C)]
-#[derive(Copy, Clone, Default, Debug)]
+#[derive(Copy, Clone, Default, Debug, PartialEq, Eq, Hash)]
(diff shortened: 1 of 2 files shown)


swooster:
NOTE: I haven't regenerated `vk.rs` because when I try to do that it comes out without newlines. (Maybe because I'm on Windows?) If somebody else is able to get that working, I'd be grateful.

Ralith:
> I haven't regenerated vk.rs because when I try to do that it comes out without newlines.

This is the expected behavior; the workflow is to run `cargo fmt` after regenerating.

aloucks:
I use `core.autocrlf=input` even on Windows and I think rustfmt will fallback to use native line endings if it can't detect any in the file (and I think the generated vk.rs is all on one line). 

Not sure if this will help your specific case, but I leave an uncommitted `rustfmt.toml` in my ash directory:

```toml
newline_style = "Unix"
```

swooster:
Ok, thanks for the explanation. I've pushed a version of the commit with changes to `vk.rs` included.

PR #253, opened 2019-12-01
Allow builders to be used inside slices

MaikKlein:
fixes https://github.com/MaikKlein/ash/issues/223

Quick and dirty implementation for builders inside slices

We replaces all slices with `&'a [Foo]` => `&'a [impl Cast<Foo>]`, and we implement `unsafe impl crate::Cast<Foo> for FooBuilder<'_> {}`.

This allows us to have a slice for both `Foo` and `FooBuilder`.

This is only a preliminary implementation, so that we can talk about it.

Alternatively we might just only use builders for builders without any generics.

tl;dr

You don't have to call build anymore 
```Rust
// new
device.create_graphics_pipelines(.., &[graphic_pipeline_infos], None)
// old
device.create_graphics_pipelines(.., &[graphic_pipeline_infos.build()], None)
```

If you call build as late as possible, you everything is still lifetime checked. But if you don't you can run into some lifetime issues as build just throws away all the lifetime information. 

With this PR you can use builders everywhere and you don't need to bother anymore with calling build manually.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -1,7 +1,7 @@
 #![allow(clippy::trivially_copy_pass_by_ref)]
 use crate::prelude::*;
 use crate::vk;
-use crate::RawPtr;
+use crate::{Cast, RawPtr};
 use std::mem;
 use std::os::raw::c_void;
 use std::ptr;
@@ -1373,7 +1373,7 @@ pub trait DeviceV1_0 {
     unsafe fn create_graphics_pipelines(
         &self,
         pipeline_cache: vk::PipelineCache,
-        create_infos: &[vk::GraphicsPipelineCreateInfo],
+        create_infos: &[impl Cast<vk::GraphicsPipelineCreateInfo>],
         allocation_callbacks: Option<&vk::AllocationCallbacks>,
     ) -> Result<Vec<vk::Pipeline>, (Vec<vk::Pipeline>, vk::Result)> {
         let mut pipelines = Vec::with_capacity(create_infos.len());
@@ -1381,7 +1381,7 @@ pub trait DeviceV1_0 {
             self.handle(),
             pipeline_cache,
             create_infos.len() as u32,
-            create_infos.as_ptr(),
+            create_infos.as_ptr() as *const _,
             allocation_callbacks.as_raw_ptr(),
             pipelines.as_mut_ptr(),
         );
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -722,7 +722,7 @@ pub trait DeviceV1_0 {
         src_buffer: vk::Buffer,
         dst_image: vk::Image,
         dst_image_layout: vk::ImageLayout,
-        regions: &[vk::BufferImageCopy],
+        regions: &[impl Cast<vk::BufferImageCopy>],
     ) {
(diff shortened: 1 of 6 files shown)


MaikKlein:
Interestingly this does not only apply to slices

```
impl<'a> ImageSubresourceLayersBuilder<'a> {
    pub fn aspect_mask(
        mut self,
        aspect_mask: ImageAspectFlags,
    ) -> ImageSubresourceLayersBuilder<'a> {
        self.inner.aspect_mask = aspect_mask;
        self
    }
```

So we might also want to address that.

Ralith:
> This does not only apply to slices

What doesn't?

MaikKlein:
```
pub fn image_subresource(
    self,
    image_subresource: ImageSubresourceLayers
) -> BufferImageCopyBuilder<'a>
``` 
https://docs.rs/ash/0.29.0/ash/vk/struct.BufferImageCopyBuilder.html#method.image_subresource

Takes it by value, so the user has to call build on `ImageSubresourceLayers::builder().build()`, in this case it is fine, as it doesn't contain any pointers.

So we could also do something like

```
pub fn image_subresource(
    self,
    image_subresource: impl Into<ImageSubresourceLayers>
) -> BufferImageCopyBuilder<'a>
```

Ralith:
Given the very marginal benefit (you lose almost nothing by calling `build` explicitly, not even safety if you're doing it in the same statement that passes the aggregate value to the implementation) I'd like to know how much this impacts the already somewhat dire build time cost ash imposes on downstream crates.

MaikKlein:
> Given the very marginal benefit (you lose almost nothing by calling build explicitly, not even safety if you're doing it in the same statement that passes the aggregate value to the implementation) I'd like to know how much this impacts the already somewhat dire build time cost ash imposes on downstream crates.

I know at least of a couple of people that run into some issues like that. And it can be easily avoided by just using builders.

I highly doubt that if will affect build times in any meaningful way, they are essentially setter functions. And we can also monomorphize inside ash, so no added build times for downstream crates.

The only con that I can see is the added complexity, and I am not sure if there is a better way to do it.

Ralith:
> they are essentially setter functions

It's not the complexity of the functions that I'm concerned about, but rather the additional monomorphization load placed on downstream compiles.

> And we can also monomorphize inside ash

I don't think Rust has this feature.

> I am not sure if there is a better way to do it.

I believe this and the status quo are the only options.

MaikKlein:
> I don't think Rust has this feature.

I have been doing it manually for quite a while. with `Into` etc.

```
pub fn do_generic(i: impl Into<Foo>) {
    fn inner(i: Foo) {
    }
    inner(i.into());
}
```

Where usually `inner` would do the heavy lifting, which is compiled in the lib. But for ash it is literally just a few assignments, so it doesn't matter much.

I often do this for serde stuff ^.

MaikKlein:
I have also showed this PR to a bunch of people, and no one was really convinced that it was necessary. So we might just close it.

Ralith:
> Where usually inner would do the heavy lifting, which is compiled in the lib. But for ash it is literally just a few assignments, so it doesn't matter much.

Ah, I know that as "manual outlining"; the code is still polymorphic, but the burden is indeed reduced in the case of a complex definition.

PR #262, opened 2020-01-14
Implement VK_KHR_draw_indirect_count

gabdube:
I also wanted to implement VK_AMD_draw_indirect_count at the same time considering it's the same API, but there is no "amd" under `extension`. If everything is fine with this PR, just tell me and I'll do the copy pasting. ;)

### DIFF
--- a/ash/src/extensions/khr/draw_indirect_count.rs
+++ b/ash/src/extensions/khr/draw_indirect_count.rs
@@ -1,0 +1,74 @@
+#![allow(dead_code)]
+use crate::version::{DeviceV1_0, InstanceV1_0};
+use crate::vk;
+use std::ffi::CStr;
+use std::mem;
+
+#[derive(Clone)]
+pub struct DrawIndirectCount {
+    handle: vk::Device,
+    draw_indirect_count_fn: vk::KhrDrawIndirectCountFn,
+}
+
+impl DrawIndirectCount {
+    pub fn new<I: InstanceV1_0, D: DeviceV1_0>(instance: &I, device: &D) -> DrawIndirectCount {
+        let draw_indirect_count_fn = vk::KhrDrawIndirectCountFn::load(|name| unsafe {
+            mem::transmute(instance.get_device_proc_addr(device.handle(), name.as_ptr()))
+        });
+
+        DrawIndirectCount {
+            handle: device.handle(),
+            draw_indirect_count_fn,
+        }
+    }
+
+    pub fn name() -> &'static CStr {
+        vk::KhrDrawIndirectCountFn::name()
+    }
+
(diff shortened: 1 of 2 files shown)


gabdube:
Hmm I'm using the same code style as the other extensions, but the imports fails. I wonder why....

gabdube:
Ok thatwas because I was using a very old version of ash locally. With everything up to date things work fine. Rustfmt and Clippy both fails, but it's not coming from my code.

ash/src/extensions/khr/draw_indirect_count.rs
```diff
@@ -0,0 +1,73 @@
...
+            handle: device.handle(),
+            draw_indirect_count_fn,
+        }
+    }
+
+    pub fn name() -> &'static CStr {
+        vk::KhrDrawIndirectCountFn::name()
+    }
+
+    pub unsafe fn cmd_draw_indexed_indirect_count(
```
aloucks: Can you links to the _individual_ man pages? Here is an example:

https://github.com/MaikKlein/ash/blob/31d14486e5c03c64d6bed9aa59d2c473abcd4339/ash/src/extensions/khr/swapchain.rs#L69-L70

aloucks:
It's interesting that the `#[doc ="<url>"]` attributes satisfy clippys check for unsafe `# Safety` sections. I guess this will enforce adding doc links to all new wrapper functions :)

MaikKlein:
Thanks for the PR, can you rebase from the newest master? The rustfmt/clippy errors should go away.

gabdube:
oh no. looks like https://github.com/MaikKlein/ash/commit/00f52cc5ad119b1f12710e4029de53794505d71d removed the functions members in KhrDrawIndirectCountFn

Friz64:
`VK_KHR_draw_indirect_count` is now included in Vulkan 1.2. It looks like that in vk.xml (the file vk.rs is generated from) the functions are now aliased to the included functions. So, i don't know if i'm wrong with this, but shouldn't you now be able to just request the extension normally in versions below Vulkan 1.2 and use the included functions (or the corresponding wrapper functions that now also exist)?

MaikKlein:
`KhrDrawIndirectCountFn` should not be empty, I assume we handle it incorrectly.

MaikKlein:
Yeah so ash handles command aliases incorrectly atm. Mostly because ash still uses the outdated vkxml format. I tried to get a quick fix in but it is too hacky, I'll just fix it properly. Not sure how long that will take.

gabdube:
No pressure. My own hacky workaround works fine in my current project.

MaikKlein:
@gabdube Sorry for taking so long, I just merged in a workaround to generate function aliases correctly. Can you rebase this PR on master?

gabdube:
nice

PR #264, opened 2020-01-16
Vulkan 1.2 Support

### LINKED ISSUE #263: Add support for 1.2
Version 1.2 got released today https://www.khronos.org/news/press/khronos-group-releases-vulkan-1.2

We should try to support it soonish. Might need to update the vk.xml parser again.

- [x] Update the parser / generator to output 1.2 ffi bindings
- [x] Expose a 1.2 high level wrapper
- [ ] Add a simple 1.2 example to test some of the new features.

AlexanderEkdahl:
Remaining work:
- [x] Include `VkPipelineExecutableInternalRepresentationKHR`(currently excluded as it generates invalid Rust code)
- [x] Passes `cargo check`
    ```
    VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_FEATURES = 49,
    VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_PROPERTIES = 50,
    VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_2_FEATURES = 51,
    VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_2_PROPERTIES = 52,
    ```

    The above associated types are not generated.
- [x] ~~Decide what to do about breaking API changes making pre-1.2 types unreachable.~~ *type aliases generated**
- [ ] Update non-generated code
- [x] Test new features against a Vulkan 1.2 Driver
- [ ] Test dependent crates(`gfx-hal`, `vk-mem`) [full list](https://crates.io/crates/ash/reverse_dependencies)

Fixes #263

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -946,7 +946,16 @@ pub fn generate_extension_constants<'a>(
                     let value = if *positive { value } else { -value };
                     Some((Constant::Number(value as i32), Some(extends.clone())))
                 }
-                _ => None,
+                EnumSpec::Value { value, extends } => {
+                    if let (Some(extends), Ok(value)) = (extends, value.parse::<i32>()) {
+                        Some((Constant::Number(value), Some(extends.clone())))
+                    } else {
+                        None
+                    }
+                }
+                _ => {
+                    None
+                },
             }?;
             let extends = extends?;
             let ext_constant = ExtensionConstant {
@@ -2176,6 +2185,7 @@ pub fn write_source_code(path: &Path) {
         })
         .flat_map(|definitions| definitions.elements.iter())
         .collect();
+    // println!("{:#?}", definitions);
 
     let enums: Vec<&vkxml::Enumeration> = spec
         .elements
@@ -2201,6 +2211,7 @@ pub fn write_source_code(path: &Path) {
         })
         .flat_map(|constants| constants.elements.iter())
         .collect();
+    // println!("{:#?}", constants);
 
     let mut fn_cache = HashSet::new();
     let mut bitflags_cache = HashSet::new();
@@ -2218,6 +2229,7 @@ pub fn write_source_code(path: &Path) {
             };
             acc
         });
+    // println!("{:#?}", const_values);
 
     let constants_code: Vec<_> = constants
         .iter()
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -89,7 +89,7 @@ named!(cfloat<&str, f32>,
 
 fn khronos_link<S: Display>(name: &S) -> Literal {
     Literal::string(&format!(
-        "<https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/{name}.html>",
+        "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/{name}.html>",
         name = name
     ))
 }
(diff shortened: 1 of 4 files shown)


AlexanderEkdahl:
@MaikKlein The generator currently fails to extend the `VkStructureType` enum with the following new hardcoded values. The XML is:

```
<feature api="vulkan" name="VK_VERSION_1_2" number="1.2" comment="Vulkan 1.2 core API interface definitions.">
    <require>
        <type name="VK_API_VERSION_1_2"/>
    </require>
    <require>
        <enum extends="VkStructureType" value="49" name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_FEATURES"/>
        <enum extends="VkStructureType" value="50" name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_PROPERTIES"/>
        <enum extends="VkStructureType" value="51" name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_2_FEATURES"/>
        <enum extends="VkStructureType" value="52" name="VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_2_PROPERTIES"/>
```

These `enum` fields are normally in the form of `extnumber/offset`.

Instrumenting `vk-parse` I can see `enum` being read but I don't understand if/how it is propagated to the parser in this library.

If I manually add these associated constants to `StructureType` everything compiles!

AlexanderEkdahl:
This PR is now able to generate all of `vk.rs` from the latest version of `Vulkan-Headers` :fire: 

The generator only needed minor changes.

The remaining code has to be written by hand or am I missing something?

Ralith:
> This PR is now able to generate all of vk.rs from the latest version of Vulkan-Headers fire

Awesome work, thank you!

> The remaining code has to be written by hand or am I missing something?

Code outside of vk.rs is handwritten, yeah. I don't think we *need* complete coverage before releasing, but it'd certainly be nice!

Friz64:
Shouldn't there also be new traits in the `version` module? Also, all of the spec links are still on `1.1`, right?

MaikKlein:
> The remaining code has to be written by hand or am I missing something?

You don't need to add the manual part in this PR, generating the underlying ffi is good enough. I really was hoping to fully generate ash before 1.2 but I haven't found the time yet.

Had a quick peek at it, looks good so far! But I'll have a closer look later today.

AlexanderEkdahl:
> > The remaining code has to be written by hand or am I missing something?
> 
> You don't need to add the manual part in this PR, generating the underlying ffi is good enough. I really was hoping to fully generate ash before 1.2 but I haven't found the time yet.
> 
> Had a quick peek at it, looks good so far! But I'll have a closer look later today.

Good to hear! I would like to this against an implementation of Vulkan 1.2 before this PR is merged. That would also require me to add the minimal code for a new `Device`/`Instance`. NVIDIA has released a beta version of their drivers supporting Vulkan 1.2.

https://developer.nvidia.com/vulkan-driver

I will update and test this weekend.

Friz64:
I have installed `mesa-git` from the AUR, which means i got a working Vulkan 1.2 driver for my RX 580. If you have something specific for me to test, let me know.

generator/src/lib.rs
```diff
@@ -2273,7 +2281,7 @@ pub fn write_source_code(path: &Path) {
     let version_macros = vk_version_macros();
     let platform_specific_types = platform_specific_types();
     let source_code = quote! {
-        #![allow(clippy::too_many_arguments, clippy::cognitive_complexity)]
+        #![allow(clippy::too_many_arguments, clippy::cognitive_complexity, clippy::wrong_self_convention)]
```
AlexanderEkdahl: `clippy::wrong_self_convention` is required due to a new `is_text` attribute.

AlexanderEkdahl:
@MaikKlein Cool! I removed the mention of Vulkan 1.2 support from the readme and I merged latest master. 

I also tested this with a new driver supporting Vulkan 1.2 and everything appears to be working but I haven't had a chance to test any of the new features yet.

Friz64:
Why remove the mention of Vulkan 1.2 support? You can use the Vulkan 1.2 API with it already, so it has support. There's just no fancy wrappers for it yet.

AlexanderEkdahl:
@Friz64 I thought it was prudent as it does not fully support Vulkan 1.2 at the same level as previous versions. Thank you for offering your help in testing.

PR #265, opened 2020-01-20
Add high level wrappers for Vulkan 1.2

Friz64:
This PR adds high level Vulkan 1.2 wrappers for `Entry`, `Instance` and `Device`. I also updated the doc links, added the missing `InstanceV1_1::get_physical_device_features2` function (fixes https://github.com/MaikKlein/ash/issues/258) and implemented `EntryV1_1` with accompanying note (fixes https://github.com/MaikKlein/ash/issues/254).

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -166,6 +166,10 @@ impl<L> EntryV1_0 for EntryCustom<L> {
 pub trait EntryV1_1: EntryV1_0 {
     fn fp_v1_1(&self) -> &vk::EntryFnV1_1;
 
+    /// Please note that it is recommended to use [`EntryCustom::try_enumerate_instance_version`]
+    /// instead because it also works with Vulkan 1.0.
+    ///
+    /// [`EntryCustom::try_enumerate_instance_version`]: ../struct.EntryCustom.html#method.try_enumerate_instance_version
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumerateInstanceVersion.html>"]
     fn enumerate_instance_version(&self) -> VkResult<u32> {
         unsafe {
@@ -176,6 +180,12 @@ pub trait EntryV1_1: EntryV1_0 {
                 _ => Err(err_code),
             }
         }
+    }
+}
+
+impl<L> EntryV1_1 for EntryCustom<L> {
+    fn fp_v1_1(&self) -> &vk::EntryFnV1_1 {
+        &self.entry_fn_1_1
     }
 }
 
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -167,10 +167,7 @@ impl<L> EntryV1_0 for EntryCustom<L> {
 pub trait EntryV1_1: EntryV1_0 {
     fn fp_v1_1(&self) -> &vk::EntryFnV1_1;
 
-    /// Please note that it is recommended to use [`EntryCustom::try_enumerate_instance_version`]
-    /// instead because it also works with Vulkan 1.0.
-    ///
-    /// [`EntryCustom::try_enumerate_instance_version`]: ../struct.EntryCustom.html#method.try_enumerate_instance_version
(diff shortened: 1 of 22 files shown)


ash/src/entry.rs
```diff
@@ -166,7 +167,11 @@ impl<L> EntryV1_0 for EntryCustom<L> {
 pub trait EntryV1_1: EntryV1_0 {
     fn fp_v1_1(&self) -> &vk::EntryFnV1_1;
 
-    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/vkEnumerateInstanceVersion.html>"]
+    /// Please note that it is recommended to use [`EntryCustom::try_enumerate_instance_version`]
+    /// instead because it also works with Vulkan 1.0.
+    ///
+    /// [`EntryCustom::try_enumerate_instance_version`]: ../struct.EntryCustom.html#method.try_enumerate_instance_version
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumerateInstanceVersion.html>"]
```
aloucks: I think this would segfault on `1.0` instances because the function pointer will always be be null. This is why the `try_enumerate_instance_version` was created. It was created on the `EntryV1_0` trait because _the `EntryV1_1` trait is not implemented at all_. You can remove the trait and it's `pub use` from `version.rs` and ash compiles just fine. 

I think a better solution is to refactor the Entry API but that probably should be tackled separately.

Friz64: I thought the trait not being implemented was a mistake because the code for it just sat there not being used. Also, i'm pretty sure it won't segfault, it will just panic (looking at the loader code).

```rust
        EntryFnV1_1 {
            enumerate_instance_version: unsafe {
                extern "system" fn enumerate_instance_version(_p_api_version: *mut u32) -> Result {
                    panic!(concat!(
                        "Unable to load ",
                        stringify!(enumerate_instance_version)
                    ))
                }
                let raw_name = stringify!(vkEnumerateInstanceVersion);
                let cname = ::std::ffi::CString::new(raw_name).unwrap();
                let val = _f(&cname);
                if val.is_null() {
                    enumerate_instance_version
                } else {
                    ::std::mem::transmute(val)
                }
            },
        }
```

I think having it implemented is the better decision for now because of consistency. The doc comment makes the situation pretty clear. If you disagree please let me know, i am free to discuss.

aloucks: Right,. I forgot about the pre-loaded dummy functions, but I still don't think this is a good idea. Now we're adding a second way to do this but that second way should **never** be used. You need to check the API version before calling 1.1+ functions but this is the function to get the API version.

I agree that it's a bit awkward right now because these traits seem pointless. I just don't think adding these improves the situation.

Friz64: I get what you are saying. Maybe we should deprecate the function in favor of `EntryCustom::try_enumerate_instance_version` with the depreciation note explaining that it will panic prior to Vulkan 1.1 so that the compiler will warn about it's usage.

AlexanderEkdahl:
I tested this and it works :champagne:  Maybe add a note in the README mentioning Vulkan 1.2 support? 

Let me know if there is anything specific you want me to test.

Ralith:
From the description I expected this to be complete, but it seems to be missing e.g. imageless framebuffers.

Friz64:
Imageless framebuffers are already automatically generated (`vk::FramebufferCreateFlags`), i just added wrapper functions for `vk::DeviceFnV1_2`.

Ralith:
Oh, right, there's no functions in that extension. :+1:

ash/src/entry.rs
Friz64: This was also suggested by @MaikKlein (https://github.com/MaikKlein/ash/issues/254#issuecomment-576385256), let me know what you think of this so i can implement it.

Friz64:
Because of no further comments on `enumerate_instance_version`, i decided to go the route of marking it deprecated as @MaikKlein also suggested. I think this PR is ready to be merged now.

Friz64:
No problem! Maybe we should release a new version on crates.io now?

PR #272, opened 2020-02-29
Add a Safety section to the documentation of create_instance and create_device

### LINKED ISSUE #271: Encoding Entry Lifetime Requirements in Types
The documentation currently states that the `Entry` which is created in order to load an `Instance`, which can be used to create a `Device`, must outlive said `Instance`, and that the `Instance` must outlive any `Device`s that are created using it, however this documentation currently only exists in the README, and was something that I missed when I initially started using this library and only noticed when carefully going through the examples and I wondered why a copy of the `Entry` was kept around after the `Instance` was created.

In my opinion (and I'd love to hear reasoning otherwise and discussion on this) this lifetime requirement should either be encoded directly in the types as lifetimes, or if that isn't feasible, then adding documentation to `Entry::create_instance` with a Safety section which talks about that lifetime requirement would be a good substitute.

I'm currently working through the codebase a little to see if I can create a pull request which makes no significant change to the API besides requiring a lifetime on the `Instance` and `Device` structs to ensure that they are outlived by the thing which created them, however since I am still relatively new to Rust, it's slow going.

IGJoshua:
This PR is the result of the fact that as a new user of this library, it caught me off guard that I wasn't able to drop `Entry` after successfully creating an `Instance`.
It adds documentation to specify these invariants which the caller must uphold when creating `Device`s and `Instance`s.

Closes #271

One caveat to this, I am currently under the impression that the entry, instance, and device traits are maintained by hand, and are not auto-generated. If that is incorrect, then likely this PR only includes edits to auto-generated code, which likely isn't desirable, and I'll happily close the PR without it merging, since this interaction *is* documented in the README.

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -77,7 +77,12 @@ pub trait EntryV1_0 {
     type Instance;
     fn fp_v1_0(&self) -> &vk::EntryFnV1_0;
     fn static_fn(&self) -> &vk::StaticFn;
-    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCreateInstance.html>"]
+    /// <https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCreateInstance.html>
+    ///
+    /// # Safety
+    /// In order for the created `Instance` to be valid for the duration of its
+    /// usage, the `Entry` this was called on must be dropped later than the
+    /// resulting `Instance`.
     unsafe fn create_instance(
         &self,
         create_info: &vk::InstanceCreateInfo,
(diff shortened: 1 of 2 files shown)


ash/src/entry.rs
```diff
@@ -77,7 +77,12 @@ pub trait EntryV1_0 {
     type Instance;
     fn fp_v1_0(&self) -> &vk::EntryFnV1_0;
     fn static_fn(&self) -> &vk::StaticFn;
-    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCreateInstance.html>"]
```
aloucks: Can you change the doc links back? These were originally generated and if we need to re-generate them it will be easier if they all have the same pattern (i.e. all doc attributes).

IGJoshua: Got those changed back!

MaikKlein:
Looks good, thanks for adding the comments.

Ralith:
This has tripped up a number of people in the past; thanks for working to improve it!

ash/src/entry.rs
aloucks: It currently renders as:

image

I think putting it above would be nicer (**note the empty line**):

```
#[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCreateInstance.html>"]
///
/// # Safety
/// In order for the created `Instance` to be valid for the duration of its
/// usage, the `Entry` this was called on must be dropped later than the
/// resulting `Instance`.
```

Will render as:

image

PR #273, opened 2020-02-29
Adjust doc link rendering with safety sections

aloucks:
Minor adjustment to how the doc links are rendered with the new safety sections.

See here for example:

https://github.com/MaikKlein/ash/pull/272#discussion_r386053611


aloucks:
@MaikKlein Can this be merged?

PR #275, opened 2020-03-14
Switch to `libloading`

### LINKED ISSUE #251: Running with wine on linux fails with "Unable to load get_instance_proc_addr"
To reproduce, build the triangle example for windows (msvc/gnu), and run the binary with wine. I set up an appveyor job to build the triangle example: https://ci.appveyor.com/project/stevebob/ash/build/job/tac8goyl6t8aghm9/artifacts

Downloading, extracting, and running that with wine gives the output:
```
$ wine triangle.exe
002c:fixme:msg:IsGUIThread 1: stub
002c:fixme:rawinput:RegisterRawInputDevices Unhandled flags 0x2100 for device 0.
002c:fixme:rawinput:RegisterRawInputDevices Unhandled flags 0x2100 for device 1.
thread 'main' panicked at 'Unable to load get_instance_proc_addr', ash\src\vk.rs:282:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

My version of wine is 4.20.
I've tested that I can run other vulkan programs with wine, namely https://github.com/skeeto/vulkan-test

Friz64:
This switches the symbol loading from `shared_library` to `libloading` because `shared_library` is not *really* maintained (see https://github.com/tomaka/shared_library/issues/6#issuecomment-264773459 for example).

One noticeable benefit this brings is that it provides a fix for #251. Note that the examples currently require the `VK_EXT_debug_report` extension and therefore won't work on wine without changes.
So: fixes #251

### DIFF
--- a/ash/Cargo.toml
+++ b/ash/Cargo.toml
@@ -11,7 +11,7 @@ documentation = "https://docs.rs/ash"
 edition = "2018"
 
 [dependencies]
-shared_library = "0.1.9"
+libloading = "0.5.2"
 
 [features]
 default = []
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -2,13 +2,13 @@ use crate::instance::Instance;
 use crate::prelude::*;
 use crate::vk;
 use crate::RawPtr;
-use shared_library::dynamic_library::DynamicLibrary;
+use libloading::Library;
 use std::error::Error;
 use std::fmt;
+use std::io;
 use std::mem;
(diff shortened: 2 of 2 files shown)


MaikKlein:
Thanks I wanted to switch to a different libloading lib a long time ago. I think I only looked at this lib briefly and was put off by the lifetimes, but it seems you can just deref them. 

Looks good thanks. I'll merge it later today.

Seems like I have to clean up some new clippy lints

Issue #178, opened 2019-01-19
List of unimplemented extensions


MaikKlein:
The following extensions are currently not implemented:

* VK_NV_ray_tracing
* VK_KHR_push_descriptor
* VK_EXT_descriptor_indexing

If an extensions that you need is not exposed in `ash` leave a comment, and I will add it to the list. 

If you would like to implement some extensions but you don't how, have a look at the currently exposed [extensions](https://github.com/MaikKlein/ash/tree/master/ash/src/extensions). Otherwise don't hesitate to ask, help is always appreciated.

msiglreith:
Would it make sense to also add some 'hidden' extensions (ie `VK_AMD_gpa_interface` or `VK_AMD_wave_limits`), which are not really specified at this point?

MaikKlein:
@msiglreith I am fine with experimental extensions. They would have to live inside the extension itself because vk.rs is generated.

I think we should new module `extensions::experimental::amd::Foo`. Semver is then ignored for those extensions.

gwihlidal:
I've been working on the VK_NV_ray_tracing support now, and likely also VK_EXT_descriptor_indexing

farnoy:
@MaikKlein There's a useful new extension I'd love to see: [VK_KHR_timeline_semaphore][1]

[1]: https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VK_KHR_timeline_semaphore

hrydgard:
(I know you know about this already, just posting for the record).

For MoltenVK use on Mac, the following extension is quite important:

https://github.com/KhronosGroup/MoltenVK/blob/4e0abab7dbf16b2d64a7e54a43e07327bd40e1d5/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h

It's unofficial but rather essential since it allows querying for some quirks that are not exposed through the Vulkan API features, such as the inability to combine layered rendering with MSAA.

pudnax:
Khronos released [VK_KHR_synchronization2](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VK_KHR_synchronization2) extension recently to simplify the interface and improve usability of these API.

K0bin:
`VK_KHR_dynamic_rendering` please. :)

MarijnS95:
[...]

PR #276, opened 2020-03-16
Add `VK_KHR_timeline_semaphore` extension support

zedrian:
Introduced `ash::extensions::khr::TimelineSemaphore` structure written by analogy with `khr::Swapchain` (except the fact that it does not hold any special functions besides those ones that are needed for extension loading).

Enabling timeline semaphore support in the application now:
```rust
let device_extension_names = [
    ash::extensions::khr::TimelineSemaphore::name().as_ptr(),
    // other device extensions
];

let mut timeline_semaphore_features = vk::PhysicalDeviceTimelineSemaphoreFeatures::builder()
    .timeline_semaphore(true);

let device_create_info = vk::DeviceCreateInfo::builder()
    .enabled_extension_names(&device_extension_names)
    .push_next(&mut timeline_semaphore_features)
    // other settings
    .build();
let device = unsafe { instance.create_device(physical_device, &device_create_info, None)? };
```


zedrian:
@farnoy I guess this PR may be interesting for you because you already requested this extension in #178  ☺

zedrian:
@farnoy I tested on my side, functions related to timeline semaphores defined in `DeviceFnV1_2` become available just using `ash::device::Device`, e.g.
```rust
unsafe { device.wait_semaphores(device.handle(), &semaphore_wait_info, std::u64::MAX).unwrap(); }
```

so you don't have to grab function pointers by yourself anymore.

farnoy:
Have you tested on a 1.0/1.1 application with this? I don't think that would work since the functions have a `KHR` suffix there. So using a `DeviceFnV1_2.get_semaphore_counter_value` would not work because it would try to load a `vkGetSemaphoreCounterValue`, whereas this extension should load `vkGetSemaphoreCounterValueKHR`.

zedrian:
@farnoy I tried this on API 1.1, and it doesn't work. As I understand, that happens because of `VK_KHR_timeline_semaphore` appeared in Vulkan 1.2. 
Is this a problem?

zedrian:
@farnoy I misread your comment at first, sorry. Yes, it will not work because of the `KHR` suffix in older versions of Vulkan.

Ralith:
The point is, the extension should provide access to the `KHR`-suffixed versions, and therefore be usable on 1.1 impls.

zedrian:
So expected implementation is that Vulkan 1.2 users call functions like
```rust
device.wait_semaphores(device.handle(), &semaphore_wait_info, timeout)?;
```
and Vulkan 1.1 users call functions like
```rust
timeline_semaphore_loader.wait_semaphores(&semaphore_wait_info, timeout)?;
```
right?

farnoy:
I would summarize it as:

1.2 users don't use this and call `device.wait_semaphores(device.handle(), &semaphore_wait_info, timeout)?;`

1.0/1.1 users need to enable the device extension and call `timeline_semaphore_loader.wait_semaphores_khr(device.handle(), &semaphore_wait_info, timeout)?;`

I think the suffix is usually preserved in ash, and the interface should be the same (because it is the same in the spec between the extension and 1.2 core)

zedrian:
BTW, `VK_KHR_timeline_semaphore` extension additionally requires `VK_KHR_get_physical_device_properties2`. But as I can see from my tests, it's enough to just add that extension name to the list of enabled extensions.

zedrian:
I added the `*_khr` functions to the `TimelineSemaphore` as discussed, so the following code is available now for Vulkan 1.0/1.1 users:
```rust
timeline_semaphore_loader.get_semaphore_counter_value_khr(device.handle(), semaphore)?;

timeline_semaphore_loader.wait_semaphores_khr(device.handle(), &semaphore_wait_info, timeout)?;
timeline_semaphore_loader.signal_semaphore_khr(device.handle(), &semaphore_signal_info, timeout)?;
```

MaikKlein:
Hey thanks for the PR. Sorry for not responding sooner. I made a PR to include function aliases in extensions https://github.com/MaikKlein/ash/pull/279


```Rust
pub struct KhrTimelineSemaphoreFn {
    pub get_semaphore_counter_value:
        extern "system" fn(device: Device, semaphore: Semaphore, p_value: *mut u64) -> Result,
    pub wait_semaphores: extern "system" fn(
        device: Device,
        p_wait_info: *const SemaphoreWaitInfo,
        timeout: u64,
    ) -> Result,
    pub signal_semaphore:
        extern "system" fn(device: Device, p_signal_info: *const SemaphoreSignalInfo) -> Result,
}

```

No need to fetch those function pointers yourself and we should update this PR when this is merged in.

zedrian:
That would be great! Waiting for #279.

PR #291, opened 2020-04-19
Update `libloading` to `0.6.1`

Friz64:
This improves the build time because `libloading` doesn't depend on `cc` anymore. See: https://github.com/nagisa/rust_libloading/issues/61

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -5,7 +5,6 @@ use crate::RawPtr;
 use libloading::Library;
 use std::error::Error;
 use std::fmt;
-use std::io;
 use std::mem;
 use std::os::raw::c_char;
 use std::os::raw::c_void;
@@ -42,7 +41,7 @@ pub struct EntryCustom<L> {
 
 #[derive(Debug)]
 pub enum LoadingError {
-    LibraryLoadError(io::Error),
+    LibraryLoadError(libloading::Error),
 }
 
 impl fmt::Display for LoadingError {
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -40,19 +40,19 @@ pub struct EntryCustom<L> {
 }
 
 #[derive(Debug)]
-pub enum LoadingError {
-    LibraryLoadError(libloading::Error),
-}
+pub struct LoadingError(libloading::Error);
 
 impl fmt::Display for LoadingError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            LoadingError::LibraryLoadError(e) => write!(f, "{}", e),
-        }
+        fmt::Display::fmt(&self.0, f)
     }
 }
 
-impl Error for LoadingError {}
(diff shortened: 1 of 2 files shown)


ash/src/entry.rs
```diff
@@ -42,7 +41,7 @@ pub struct EntryCustom<L> {
 
 #[derive(Debug)]
 pub enum LoadingError {
-    LibraryLoadError(io::Error),
+    LibraryLoadError(libloading::Error),
```
MaikKlein, line 44: I am not so sure about that change. I don't really want to leak types of other crates into the public interface of ash. 
- This is a breaking change
- Couplse ash and libloading. Updating libloading might also be a breaking change for ash in the future.

I am not too sure what the best practice here would be.

Ralith: I think the idiomatic solution is to wrap `libloading::Error` in an ash-defined newtype that implements `Display` and `Error`. Or, since this is an enum with a single case, just replace the enum with such a newtype directly.

Friz64: I think we can tackle this problem elsewhere. Currently `new_custom` looks this this
```rust
pub fn new_custom<Open, Load>(open: Open, mut load: Load) -> Result<Self, LoadingError>
    where
        Open: FnOnce() -> Result<L, LoadingError>,
        Load: FnMut(&mut L, &::std::ffi::CStr) -> *const c_void { // ... }
```
which is kind of nonsensical because we explicitly move the error handling *into* the function. I think it should look like this:
```rust
pub fn new_custom<Load>(lib: L, mut load: Load) -> Self
    where
        Load: FnMut(&mut L, &::std::ffi::CStr) -> *const c_void { // ... }
```
This way the user can handle the error of creating the lib in his own code. Then we could change the signature of the `libloading` based `Entry` new function to this
```rust
pub fn new() -> std::result::Result<CoreLoader<libloading::Library>, libloading::Error> { // ... }
```
Now we only expose `libloading` types in `#[cfg(feature = "loading")]` code

Ralith: That's a separate issue. We still need to newtype the error to avoid gratuitous breakage.

Friz64: Yes, my proposal would still be breaking, but wouldn't it be better if we fixed the root of the issue instead of working around it?

Ralith: The root of the issue is that a foreign crate's type, namely `libloading::Error`, is being added to ash's public interface. Wrapping a newtype around it is the necessary fix.

Friz64: @Ralith I implemented your suggested change in 36a3942f0426f0aee21d0d9a74470bdcca06a157

Ralith:
Looks like github's a bit confused about the merge situation; there's duplicate and unrelated commits in this PR now. Could you rebase the intended changes onto master?

Friz64:
PR https://github.com/MaikKlein/ash/pull/292 was merged, which was based on this PR and therefore also includes these changes.

PR #294, opened 2020-04-21
Initialize MemoryRequirements2 with `Default`

### LINKED ISSUE #239: SubmitInfoBuilder: have two slices, but only one count.
wait_semaphore_count is shared by p_wait_semaphores and p_wait_dst_stage_mask, but there are two functions used to set these and they can be passed different counts.  This silently leads to crashes that validation layers don't catch.

### LINKED ISSUE #293: get_acceleration_structure_memory_requirements
```rust
pub unsafe fn get_acceleration_structure_memory_requirements(
        &self,
        info: &vk::AccelerationStructureMemoryRequirementsInfoKHR,
    ) -> vk::MemoryRequirements2KHR {
        let mut requirements = mem::zeroed();
        self.ray_tracing_fn
            .get_acceleration_structure_memory_requirements_khr(
                self.handle,
                info,
                &mut requirements,
            );
        requirements
    }
```

You shouldn't create requirements  with `mem::zeroed()`, because s_type must be VK_STRUCTURE_TYPE_MEMORY_REQUIREMENTS_2, I think it should be `Default::default()`

This is the error from validation layer:

`Validation Error: [ VUID-VkMemoryRequirements2-sType-sType ] Object 0: handle = 0x16c51b01b08, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xa2404bc8 | vkGetAccelerationStructureMemoryRequirementsKHR: parameter pMemoryRequirements->sType must be VK_STRUCTURE_TYPE_MEMORY_REQUIREMENTS_2. The Vulkan spec states: sType must be VK_STRUCTURE_TYPE_MEMORY_REQUIREMENTS_2`

Friz64:
[`MemoryRequirements2`](https://docs.rs/ash/0.30.0/ash/vk/struct.MemoryRequirements2.html) has a `sType` field, which should not be zero-initialized because it results in a invalid instance. I looked through some more usages of `mem::zeroed` and this seems like the only wrong one.

Fixes #293 (I don't have hardware that supports this extension but this should fix it)

### DIFF
--- a/ash/src/extensions/khr/ray_tracing.rs
+++ b/ash/src/extensions/khr/ray_tracing.rs
@@ -72,7 +72,7 @@ impl RayTracing {
         &self,
         info: &vk::AccelerationStructureMemoryRequirementsInfoKHR,
     ) -> vk::MemoryRequirements2KHR {
-        let mut requirements = mem::zeroed();
+        let mut requirements = Default::default();
         self.ray_tracing_fn
             .get_acceleration_structure_memory_requirements_khr(
                 self.handle,


Ralith:
Nice catch!

PR #295, opened 2020-04-29
Use DebugUtils instead of DebugReport in example.

bzm3r:
This patch changes the example to use the DebugUtils extension instead of DebugReport. Beyond the basic message, additional info provided by DebugUtils is not printed, in order to keep things short and sweet. Let me know if additional info should be printed too. 

[x] ran `cargo fmt`
[x] ran `triangle` and `texture` binaries

### DIFF
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -23,7 +23,7 @@ use std::mem;
 #[cfg(all(unix, not(target_os = "android"), not(target_os = "macos")))]
 use ash::extensions::khr::XlibSurface;
 use ash::extensions::{
-    ext::DebugReport,
+    ext::DebugUtils,
     khr::{Surface, Swapchain},
 };
 
@@ -33,11 +33,12 @@ use ash::extensions::khr::Win32Surface;
 use ash::extensions::mvk::MacOSSurface;
 pub use ash::version::{DeviceV1_0, EntryV1_0, InstanceV1_0};
 use ash::{vk, Device, Entry, Instance};
+use std::borrow::Cow;
 use std::cell::RefCell;
 use std::default::Default;
 use std::ffi::{CStr, CString};
 use std::ops::Drop;
-use std::os::raw::{c_char, c_void};
+use std::os::raw::c_void;
 
 // Simple offset_of macro akin to C++ offsetof
 #[macro_export]
@@ -181,7 +182,7 @@ fn extension_names() -> Vec<*const i8> {
     vec![
         Surface::name().as_ptr(),
         XlibSurface::name().as_ptr(),
-        DebugReport::name().as_ptr(),
+        DebugUtils::name().as_ptr(),
     ]
 }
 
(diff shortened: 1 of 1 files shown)


examples/src/lib.rs
```diff
@@ -199,21 +200,52 @@ fn extension_names() -> Vec<*const i8> {
...
+) -> vk::Bool32 {
+    let callback_data = *p_callback_data;
+
+    let message_severity = match message_severity {
+        vk::DebugUtilsMessageSeverityFlagsEXT::ERROR => String::from("ERROR"),
+        vk::DebugUtilsMessageSeverityFlagsEXT::WARNING => String::from("WARNING"),
+        vk::DebugUtilsMessageSeverityFlagsEXT::INFO => String::from("INFO"),
+        vk::DebugUtilsMessageSeverityFlagsEXT::VERBOSE => String::from("VERBOSE"),
+        _ => String::from("WARNING"),
+    };
```
Ralith, line 221: There's no need to allocate a `String` for these. Moreover, this roughly reimplements the existing `Debug` impl; better for an example to demonstrate use of that for improved concision, handling of unknown values, and free updates in the future.

examples/src/lib.rs
```diff
@@ -199,21 +200,52 @@ fn extension_names() -> Vec<*const i8> {
...
+    let message_type = &format!("{:?}", message_type);
```
Ralith, line 222: This could be inlined into the `println!` below rather than allocating an additional intermediate `String`.

examples/src/lib.rs
```diff
@@ -199,21 +200,52 @@ fn extension_names() -> Vec<*const i8> {
...
+    println!(
+        "{}: {}\n",
+        message_severity,
+        format!(
+            "\n{} [{} ({})] : {}",
+            message_type,
+            message_id_name,
+            &message_id_number.to_string(),
+            message
+        )
```
Ralith, line 246: This should be inlined into the enclosing `println`.

examples/src/lib.rs
bzm3r: Done!

examples/src/lib.rs
bzm3r: Done!

examples/src/lib.rs
bzm3r: Done!

examples/src/lib.rs
```diff
@@ -33,11 +33,12 @@ use ash::extensions::khr::Win32Surface;
 use ash::extensions::mvk::MacOSSurface;
 pub use ash::version::{DeviceV1_0, EntryV1_0, InstanceV1_0};
 use ash::{vk, Device, Entry, Instance};
+use std::borrow::Cow;
 use std::cell::RefCell;
 use std::default::Default;
 use std::ffi::{CStr, CString};
 use std::ops::Drop;
-use std::os::raw::{c_char, c_void};
+use std::os::raw::c_void;
```
Ralith, line 41: Dead `use`

PR #308, opened 2020-06-22
Add interoperability functions for raw-window-handle

msiglreith:
Adds functions from [ash-window](https://github.com/norse-rs/ash-window)

https://github.com/norse-rs/ash-window/blob/master/examples/winit.rs shows the usage of the API together with `winit`

A few points up for discussion:
- Metal uses the portability extension `VK_EXT_metal_surface`, which requires at least an recent enough MoltenVK version, gfx-portability exposes it as well
- Metal uses the `raw-window-metal` (which I maintain so far) for acquiring the `CAMetalLayer`. In case it needs to be allocated we currently leak it unfortunately

Closes https://github.com/MaikKlein/ash/issues/232


MaikKlein:
Sorry for the late reply.

I have been a bit wary about util functionality in `ash`, but `ash-window` seems like something that should be in core. A few concerns are the external dependencies that are not `1.0`, and I know that is a bit hypocritical because ash is still below `1.0` but I have. Both of those dependencies should be easy to stabilize.

The dependencies also seem to use `libc` only for `c_void` and `c_ulong`. Any reason to not use the rust std here?

So I think moving this into `ash` would be something that I am open to. But I am also thinking if we should just move `ash-window` into the ash workspace. At least I can make sure that it stays up to date with ash releases. Is that something that you be okay with?

msiglreith:
> I have been a bit wary about util functionality in ash, but ash-window seems like something that should be in core. A few concerns are the external dependencies that are not 1.0, and I know that is a bit hypocritical because ash is still below 1.0 but I have. Both of those dependencies should be easy to stabilize.

I can't speak for `raw-window-handle`, but for `raw-window-metal` it's mostly down to having some more real application building on top of it. Thanks to contributors the general library was tested in a few basic examples but even though there might be issues. Code is basically extracted from gfx-rs.

> The dependencies also seem to use libc only for c_void and c_ulong. Any reason to not use the rust std here?

Not sure, for `raw-window-handle` see https://github.com/rust-windowing/raw-window-handle/issues/3
The OSX/IOS libraries are handled by the servo group.

> So I think moving this into ash would be something that I am open to. But I am also thinking if we should just move ash-window into the ash workspace. At least I can make sure that it stays up to date with ash releases. Is that something that you be okay with?

Sure, fine for me as well (:

msiglreith:
Changed it towards a workspace member.

MaikKlein:
Looks good, but it seem there is a build error on the CI for the sdl2 example. I think you can just enabled the `static_link` feature here for fermium.

msiglreith:
Hmm, doesn't seem to be work without installing SDL2 as the `build.rs` seems to expect it in all cases for non-win32 env. I will just drop the example.

PR #313, opened 2020-07-10
 Added `VK_KHR_pipeline_executable_properties` extension support. 

zedrian:
BTW, what do you folks think about adding small documentation to extension types (like this `PipelineExecutableProperties`) as they are implemented by people, not autogenerated? Link to the corresponding page in the _Khronos_ registry at least, but as for me small code examples would also be nice. Or this crate was designed to be super-dry?

### DIFF
--- a/ash/src/extensions/khr/pipeline_executable_properties.rs
+++ b/ash/src/extensions/khr/pipeline_executable_properties.rs
@@ -1,0 +1,78 @@
+#![allow(dead_code)]
+use crate::prelude::*;
+use crate::version::{EntryV1_0, InstanceV1_0};
+use crate::vk;
+use std::ffi::CStr;
+use std::mem;
+use std::ptr;
+use crate::vk::PipelineExecutableStatisticFormatKHR;
+
+#[derive(Clone)]
+pub struct PipelineExecutableProperties {
+    handle: vk::Instance,
+    pipeline_executable_properties_fn: vk::KhrPipelineExecutablePropertiesFn,
+}
+
+impl PipelineExecutableProperties {
+    pub fn new<E: EntryV1_0, I: InstanceV1_0>(entry: &E, instance: &I) -> PipelineExecutableProperties {
+        let pipeline_executable_properties_fn = vk::KhrPipelineExecutablePropertiesFn::load(|name| unsafe {
+            mem::transmute(entry.get_instance_proc_addr(instance.handle(), name.as_ptr()))
+        });
+
+        PipelineExecutableProperties {
+            handle: instance.handle(),
+            pipeline_executable_properties_fn,
+        }
+    }
+
+    pub fn name() -> &'static CStr {
+        vk::KhrPipelineExecutablePropertiesFn::name()
+    }
+
+    #[doc = "https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPipelineExecutableInternalRepresentationsKHR.html"]
+    pub unsafe fn get_pipeline_executable_internal_representations(&self, device: vk::Device, executable_info: &vk::PipelineExecutableInfoKHR) -> VkResult<Vec<vk::PipelineExecutableInternalRepresentationKHR>> {
+        let mut count = 0;
+        self.pipeline_executable_properties_fn.get_pipeline_executable_internal_representations_khr(device, executable_info, &mut count, ptr::null_mut());
+        let mut v = Vec::with_capacity(count as usize);
+        let err_code = self.pipeline_executable_properties_fn.get_pipeline_executable_internal_representations_khr(device, executable_info, &mut count, v.as_mut_ptr());
+        v.set_len(count as usize);
+        match err_code {
+            vk::Result::SUCCESS => Ok(v),
+            _ => Err(err_code),
+        }
+    }
+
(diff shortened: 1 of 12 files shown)


MaikKlein:
> BTW, what do you folks think about adding small documentation to extension types (like this PipelineExecutableProperties) as they are implemented by people, not autogenerated?

My overall plan is to auto generate all the extensions in the future which would include the docs, but I haven't had much time to work on it.

ash/src/extensions/khr/pipeline_executable_properties.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) -> VkResult<Vec<vk::PipelineExecutablePropertiesKHR>> {
+        let mut count = 0;
+        self.pipeline_executable_properties_fn
+            .get_pipeline_executable_properties_khr(
+                device,
+                pipeline_info,
+                &mut count,
+                ptr::null_mut(),
+            );
+        let mut v: Vec<_> = (0..count).map(|_| Default::default()).collect();
```
MaikKlein, line 79: You can just use `vec![Default::default(); count]`

ash/src/extensions/khr/pipeline_executable_properties.rs
```diff
@@ -0,0 +1,132 @@
...
+        let err_code = self
+            .pipeline_executable_properties_fn
+            .get_pipeline_executable_properties_khr(
+                device,
+                pipeline_info,
+                &mut count,
+                v.as_mut_ptr(),
+            );
+        v.set_len(count as usize);
```
MaikKlein, line 88: No need to set the len if you already initialized the vec.

ash/src/extensions/khr/pipeline_executable_properties.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) -> VkResult<Vec<vk::PipelineExecutableInternalRepresentationKHR>> {
+        let mut count = 0;
+        self.pipeline_executable_properties_fn
+            .get_pipeline_executable_internal_representations_khr(
+                device,
+                executable_info,
+                &mut count,
+                ptr::null_mut(),
+            );
+        let mut v: Vec<_> = (0..count).map(|_| Default::default()).collect();
```
MaikKlein, line 49: Same as above

ash/src/extensions/khr/pipeline_executable_properties.rs
```diff
@@ -0,0 +1,132 @@
...
+        let err_code = self
+            .pipeline_executable_properties_fn
+            .get_pipeline_executable_internal_representations_khr(
+                device,
+                executable_info,
+                &mut count,
+                v.as_mut_ptr(),
+            );
+        v.set_len(count as usize);
```
MaikKlein, line 58: Same as above

ash/src/extensions/khr/pipeline_executable_properties.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) -> VkResult<Vec<vk::PipelineExecutableStatisticKHR>> {
+        let mut count = 0;
+        self.pipeline_executable_properties_fn
+            .get_pipeline_executable_statistics_khr(
+                device,
+                executable_info,
+                &mut count,
+                ptr::null_mut(),
+            );
+        let mut v: Vec<_> = (0..count).map(|_| Default::default()).collect();
```
MaikKlein, line 109: Same as above

ash/src/extensions/khr/pipeline_executable_properties.rs
```diff
@@ -0,0 +1,132 @@
...
+        let err_code = self
+            .pipeline_executable_properties_fn
+            .get_pipeline_executable_statistics_khr(
+                device,
+                executable_info,
+                &mut count,
+                v.as_mut_ptr(),
+            );
+        v.set_len(count as usize);
```
MaikKlein, line 118: Same as above

ash/src/extensions/khr/pipeline_executable_properties.rs
Ralith: In general you *do* actually need to re-check the length after the second call, since there's usually no guarantee that it won't change. Perhaps that's unlikely in these cases, though.

MaikKlein: It doesn't write more than count, and if the count would be less it would return `Incomplete`. I don't think we handle that case anywhere in ash, maybe we should?

Ralith: `Incomplete` is returned if it would have to return *more*. If less is returned it just does so. Per the spec:

> `VK_INCOMPLETE` - A return array was too small for the result

Ralith: For maximum robustness, you actually need a loop that checks the return code and length after each call, and only exits when both are sensible. We should handle this in ash, especially for stuff like physical device enumeration where races are actually likely.

MaikKlein: There is also:

> If pPhysicalDeviceCount is smaller than the number of physical devices available, VK_INCOMPLETE will be returned instead of VK_SUCCESS, to indicate that not all the available physical devices were returned.

From https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumeratePhysicalDevices.html

MaikKlein: Oh I see now what you mean.

PR #320, opened 2020-08-11
Correction of the language name in README.md

RustyNixieTube:
crates.io does the syntax coloration only when the word `rust` is written in lowercase.
For example [https://crates.io/crates/debug_print](https://crates.io/crates/debug_print): 
in the [README.md](https://github.com/RustyNixieTube/debug_print/blob/master/README.md) rust is written in lowercase and then in [crates.io](https://crates.io/crates/debug_print) the syntax coloration appears.
but in your [README.md](https://github.com/MaikKlein/ash/blob/master/README.md) `Rust` is written an with uppercase R and syntax coloration doesn't appear in your [crates.io](https://crates.io/crates/ash).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ A very lightweight wrapper around Vulkan
 
 ## Features
 ### Explicit returns with `Result`
-```Rust
+```rust
 // function signature
 pub fn create_instance(&self,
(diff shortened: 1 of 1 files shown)


Ralith:
Nice catch, thanks!

PR #322, opened 2020-09-01
Expose header version and extension spec version constants

MarijnS95:
We need to retrieve the header version and extension spec versions `ash` was compiled against. This includes some ugly magic to rework `vk.xml` C-macros (:disappointed:).

Please see inline `TODO`s for areas open to improvement before merging.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1124,14 +1124,28 @@ pub fn generate_extension<'a>(
     Some(q)
 }
 pub fn generate_define(define: &vkxml::Define) -> Tokens {
-    if let Some(value) = &define.value {
-        let name = constant_name(&define.name);
-        if name == "NULL_HANDLE" {
-            return quote!();
-        }
-        let ident = Ident::from(name.as_str());
+    let name = constant_name(&define.name);
+    let ident = Ident::from(name.as_str());
+    let deprecated = define
+        .comment
+        .as_ref()
+        .map_or(false, |c| c.contains("DEPRECATED"));
 
+    if name == "NULL_HANDLE" || deprecated {
+        quote!()
+    } else if let Some(value) = &define.value {
         str::parse::<u32>(value).map_or(quote!(), |v| quote!(pub const #ident: u32 = #v;))
+    } else if let Some(c_expr) = &define.c_expression {
+        if define.defref.contains(&"VK_MAKE_VERSION".to_string()) {
+            let c_expr = c_expr.replace("VK_", "");
+            let c_expr = c_expr.replace("MAKE_VERSION", "crate::vk::make_version");
+            use proc_macro2::TokenStream;
+            use std::str::FromStr;
+            let c_expr = TokenStream::from_str(&c_expr).unwrap();
+            quote!(pub const #ident: u32 = #c_expr;)
+        } else {
+            quote!()
+        }
     } else {
         quote!()
     }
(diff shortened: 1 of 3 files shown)


generator/src/lib.rs
```diff
@@ -1123,6 +1146,40 @@ pub fn generate_extension<'a>(
...
+        quote!()
+    } else if let Some(value) = &define.value {
+        if let Ok(v) = str::parse::<u32>(value) {
+            quote!(pub const #ident: u32 = #v;)
+        } else {
+            quote!()
+        }
+    } else if let Some(c_expr) = &define.c_expression {
+        if define.defref.contains(&"VK_MAKE_VERSION".to_string())
+            && !define
```
MaikKlein, line 1164: Can you move this into a variable and then do `&& !deprecated`?

generator/src/lib.rs
```diff
@@ -1123,6 +1146,40 @@ pub fn generate_extension<'a>(
     };
     Some(q)
 }
+pub fn generate_define(define: &vkxml::Define) -> Tokens {
+    // TODO: Would rather leverage generate_constant
```
MaikKlein, line 1150: you can also remove the todos

MaikKlein (changes requested):
Looks good, sorry for responding so late. I added a few nitpicks

generator/src/lib.rs
```diff
@@ -1123,6 +1146,40 @@ pub fn generate_extension<'a>(
...
+    let name = constant_name(&define.name);
+    let ident = Ident::from(name.as_str());
+
+    if name == "NULL_HANDLE" {
+        quote!()
+    } else if let Some(value) = &define.value {
+        if let Ok(v) = str::parse::<u32>(value) {
```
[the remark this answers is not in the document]
MarijnS95: How about

```rust
        str::parse::<u32>(value).map_or_else(|_| quote!(), |v| quote!(pub const #ident: u32 = #v;))
```

Or

```rust
        str::parse::<u32>(value).map_or(quote!(), |v| quote!(pub const #ident: u32 = #v;))
```

?

generator/src/lib.rs
MarijnS95: Yeah, I don't think we'll refactor the code surrounding this. Tried working on a tokenstream too, while that might be advantageous if operating on more complex expressions in the future, it's too much overhead right now.

generator/src/lib.rs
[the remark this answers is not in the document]
MaikKlein: Either works :)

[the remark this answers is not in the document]
MarijnS95: Would have loved a `map_or_default` in addition to `unwrap_or_default` :grin:

generator/src/lib.rs
MarijnS95: Moved it to the top-level:

```rust
    if name == "NULL_HANDLE" || deprecated {
        quote!()
```

Guess we don't want to generate any of these if they are deprecated ;)

generator/src/lib.rs
```diff
@@ -1075,13 +1075,36 @@ pub fn generate_extension_commands<'a>(
...
+            _ => None,
+        })
+        .and_then(|e| {
+            if let vk_parse::EnumSpec::Value { value, .. } = &e.spec {
+                let v: u32 = str::parse(value).unwrap();
+                Some(quote!(pub const SPEC_VERSION: u32 = #v;))
+            } else {
+                None
+            }
+        });
```
MarijnS95, line 1098: How about collapsing the `and_then` case into `find_map`?

MaikKlein: I think this is fine

MarijnS95: Saves on nesting, that's worth something tool and might make the filter reusable in the future.

PR #324, opened 2020-09-03
Update example to use ash-window

adrien-ben:
Hi ! Don't know if it makes sense but here is a PR updating examples to use ash-window.


Ralith (approved):
Thanks, IMO this was badly needed!

PR #327, opened 2020-10-02
Optimized example fences. Waited immediately. Now waits before reuse (next frame)

sebbbi:
The examples (triangle and texture) wait for a fence immediately after submitting a command buffer. This is not a good practice for performance. I moved the wait until the command buffer is reused (one frame later).

Measured average frame times on Intel i7-1065g7 (4K display, 200% scaling).

**Triangle sample:**
Old = 3.5ms
New = 2.4ms

**Texture sample:**
Old = 4.2ms
New = 3.0ms

### DIFF
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -29,6 +29,7 @@ macro_rules! offset_of {
 pub fn record_submit_commandbuffer<D: DeviceV1_0, F: FnOnce(&D, vk::CommandBuffer)>(
     device: &D,
     command_buffer: vk::CommandBuffer,
+    command_buffer_reuse_fence: vk::Fence,
     submit_queue: vk::Queue,
     wait_mask: &[vk::PipelineStageFlags],
     wait_semaphores: &[vk::Semaphore],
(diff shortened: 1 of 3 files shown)


MaikKlein:
Thanks for the PR :), I'll have a look at it later today.

examples/src/lib.rs
```diff
@@ -26,16 +26,26 @@ macro_rules! offset_of {
     }};
 }
```
MaikKlein, line 28:
```suggestion
/// Helper function for submitting command buffers. Immediately waits for the fence before the command buffer
/// is executed. That way we can delay the waiting for the fences by 1 frame which is good for performance.
/// Make sure to create the fence in a signaled state on the first use.
```

MaikKlein (commented):
Thanks, looks good :). I inserted a small comment as the waiting might throw of beginners.

examples/src/lib.rs
```diff
@@ -459,9 +468,20 @@ impl ExampleBase {
...
+                .create_fence(&fence_create_info, None)
+                .expect("Create fence failed.");
+            let setup_commands_reuse_fence = device
+                .create_fence(&fence_create_info, None)
+                .expect("Create fence failed.");
+
             record_submit_commandbuffer(
                 &device,
                 setup_command_buffer,
+                setup_commands_reuse_fence,
```
MaikKlein, line 484: Wait is that actually correct now? Because `record_submit_commandbuffer` only waits on the 2nd call now, and we do the setup only once.

sebbbi: There's no need to wait for the setup command buffer on CPU side if nobody will reuse that command buffer.

sebbbi: You don't even need a separate setup command buffer. That could be removed.

MaikKlein (approved):

PR #328, opened 2020-10-14
Simplify example surface format selection

swooster:
According to specs for vkGetPhysicalDeviceSurfaceFormatsKHR(), at least
one surface format must be returned, and returned formats must not be
VK_FORMAT_UNDEFINED.


Ralith (approved):
Thanks!

PR #329, opened 2020-10-16
Simplify example memory selection

swooster:
Hopefully this will make it easier to understand `find_memorytype_index()`:

* I find it harder to reason about loops that write to variables they read from, so this avoids repeatedly updating `memory_type_bits` in favor of making the relationship between `index` and `memory_req.memory_type_bits` more explicit.
* This has merged `find_memorytype_index_f()` into `find_memorytype_index()` to reduce the amount of jumping around needed to understand what's going on.

~Unfortunately, this _does_ come with a behavior change: If an exactly matching memory type cannot be found, this now returns the last acceptable memory type, whereas it previously returned the first.~

### DIFF
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -122,34 +122,15 @@ pub fn find_memorytype_index(
     memory_prop: &vk::PhysicalDeviceMemoryProperties,
     flags: vk::MemoryPropertyFlags,
 ) -> Option<u32> {
-    // Try to find an exactly matching memory flag
-    let best_suitable_index =
-        find_memorytype_index_f(memory_req, memory_prop, flags, |property_flags, flags| {
-            property_flags == flags
-        });
-    if best_suitable_index.is_some() {
-        return best_suitable_index;
-    }
-    // Otherwise find a memory flag that works
-    find_memorytype_index_f(memory_req, memory_prop, flags, |property_flags, flags| {
-        property_flags & flags == flags
-    })
-}
-
-pub fn find_memorytype_index_f<F: Fn(vk::MemoryPropertyFlags, vk::MemoryPropertyFlags) -> bool>(
-    memory_req: &vk::MemoryRequirements,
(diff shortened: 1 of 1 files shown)


swooster:
Note: Clippy is complaining about files this hasn't touched.

Ralith:
> Unfortunately, this does come with a behavior change: If an exactly matching memory type cannot be found, this now returns the last acceptable memory type, whereas it previously returned the first.

This is actually undesired behavior. The ordering of memory types in Vulkan is such that, of the types which satisfy a mask, the fastest ones will be first. See discussion and example code at https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#memory-device.

swooster:
Ok, I've fixed that. I wasn't sure if it's readable or confusing to use [`Option`'s `get_or_insert()`](https://doc.rust-lang.org/beta/std/option/enum.Option.html#method.get_or_insert) without reading its return value...

MaikKlein:
>.. and the fallback mechanism is unnecessary
Can you explain why the fallback is unnecessary?

swooster:
I've added a couple additional tweaks:
* I expanded the commit message because I realized I need to justify the removal of the two-pass search.
* I decided I preferred `(1 << index) & memory_req.memory_type_bits != 0` over `(memory_req.memory_type_bits >> index) & 1 == 1`

swooster:
@MaikKlein: Sorry, I didn't see your comment until just now. The link that Ralith posted says "For each pair of elements X and Y returned in memoryTypes, X must be placed at a lower index position than Y if [...] the set of bit flags returned in the propertyFlags member of X is a strict subset of the set of bit flags returned in the propertyFlags member of Y [...]". So any memory type that exactly matches `flags` will come before a memory type that contains `flags`.

MaikKlein:
Ah interesting they changed the example code :). That seems reasonable to me.

MaikKlein (approved):

PR #334, opened 2020-10-29
Only scan memory_type_count memory types in examples

swooster:
The example code originally checked all 32 memory types even if the instance returned fewer memory types.


Ralith:
Good catch!

PR #335, opened 2020-11-11
remove erroneous assert in get_query_pool_results

filnet:
the stride can be bigger than a u64
for example when querying more than one pipeline statistic

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -1452,10 +1452,6 @@ pub trait DeviceV1_0 {
     ) -> VkResult<()> {
         let data_length = query_count as usize;
         assert!(
-            mem::size_of::<T>() <= mem::size_of::<u64>(),
-            "T can not be bigger than an u64"
-        );
-        assert!(
             data_length <= data.len(),
             "query_count was higher than the length of the slice"
         );


filnet:
Any chance to get this reviewed and merged ?

MarijnS95:
Perhaps it is better to turn this assert into a check for multiples of `4`, or `8` if `VK_QUERY_RESULT_64_BIT` is set? (after all this is what the spec mandates, nothing about stride being at most `8` bytes).

Also any idea why `query_count: u32,` is in the function signature, and not using the length of `data: &mut [T]`? That parameter is (at most) a multiple of `T` elements anyway, and slice ranges are easily taken.

MaikKlein:
Related: https://github.com/MaikKlein/ash/issues/100

Sorry I'll get to it later today

filnet:
Some of those asserts duplicate what the Vulkan Validation Layer (VVL) does.
So I am not sure we want to replicate them.

MaikKlein:
I think we can merge as is. Still we'd probably need to expose stride as well + either `data: &mut [T]` or `data: &mut [u8]`.

PR #339, opened 2020-11-20
Preliminary cleanup pass

MarijnS95:
Hi Maik,

Apologies for this massive PR. It's some upfront cleanup work that came up while working on `ash`, in particular around `Result` types and error handling.

I can only recommend to review the changes per commit to get around the bulk of it easily. While this aims to make the code ever so slightly better it results in lots of noise, let me know if any commit (hash) is not worth taking in and I'll gladly remove it again.

To highlight the changes:
- Make sure all results are checked with `#[must_use]` on `vk::Result`;
- Conversion helpers from `vk::Result` to `Result<T, vk::Result>`;
- Clippy fixes;
- Simplifications in the generator around array code handling;
- Parsing some C expressions into `TokenStream` and replacing identifiers with their Rust counterpart. This is going to be more relevant in the future;
- A breaking API change: 19552b241d34a483c9fdc03e04fef20788bf67b1. I don't think this one is worth taking in now that I look at it again.

Thanks!

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -619,26 +619,15 @@ pub trait ToTokens {
 }
 impl ToTokens for vkxml::ReferenceType {
     fn to_tokens(&self, is_const: bool) -> Tokens {
-        let ptr_name = match self {
-            vkxml::ReferenceType::Pointer => {
-                if is_const {
-                    "*const"
-                } else {
-                    "*mut"
-                }
-            }
-            vkxml::ReferenceType::PointerToPointer => "*mut *mut",
-            vkxml::ReferenceType::PointerToConstPointer => {
-                if is_const {
-                    "*const *const"
-                } else {
-                    "*mut *const"
-                }
-            }
+        let r = if is_const {
+            quote!(*const)
+        } else {
+            quote!(*mut)
         };
-        let ident = Term::intern(ptr_name);
-        quote! {
-            #ident
+        match self {
+            vkxml::ReferenceType::Pointer => quote!(#r),
+            vkxml::ReferenceType::PointerToPointer => quote!(#r *mut),
+            vkxml::ReferenceType::PointerToConstPointer => quote!(#r *const),
         }
     }
 }
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -2278,8 +2278,8 @@ pub fn generate_const_debugs(const_values: &BTreeMap<Ident, Vec<ConstantMatchInf
         #(#impls)*
     }
 }
-pub fn generate_aliases_of_types<'a>(
-    types: &'a vk_parse::Types,
+pub fn generate_aliases_of_types(
+    types: &vk_parse::Types,
     ty_cache: &mut HashSet<Ident, impl BuildHasher>,
 ) -> Tokens {
     let aliases = types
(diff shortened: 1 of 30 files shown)


generator/src/lib.rs
```diff
@@ -1695,81 +1758,78 @@ pub fn derive_setters(
...
+                let slice_param_ty_tokens = field.safe_type_tokens(quote!('a));
                 return Some(quote!{
-                        pub fn #param_ident_short(mut self, #param_ident_short: #slice_param_ty_tokens) -> #name_builder<'a> {
-                            self.inner.#param_ident = #param_ident_short;
-                            self
-                        }
+                    pub fn #param_ident_short(mut self, #param_ident_short: #slice_param_ty_tokens) -> #name_builder<'a> {
+                        self.inner.#param_ident = #param_ident_short;
+                        self
+                    }
```
MarijnS95, line 1827: While refactoring I came across this case and wondered why it is not dealing with the mutable pointer case. This PR introduces `to_safe_tokens` that already "substitutes" `*` with `&'a` and the appropriate mutability keyword. It could be used but breaks some `*mut c_void` setters. Addressing that and it turns up there are quite a few `&'a c_void` in the codebase as well. What do we want to do here? See the change this results in here: https://github.com/MaikKlein/ash/commit/ffa0a76ef28c553946136dcadc60a5dbf4689028

MaikKlein:
Sorry for being a bit unresponsive, I am a bit swamped right now. The result changes look great :100:, I'll have a look at the rest in a few days.

MarijnS95:
@MaikKlein No worries, there's no hurry in merging this as the official RT spec is still in the process of being dropped. It's actually convenient so that I could sneak in some extra fixes, but do let me know if this gets too large, which it already is to some extent!

generator/src/lib.rs
MaikKlein: Hm good question, I am not sure.. I think we want to use pointers for any opaque types. IIRC `ANativeWindow` is one as well.

MarijnS95: I've since added `ANativeWindow` and `CAMetalLayer`, see the new commit here: 4d9647f0f8617393f2e26dbba0d94d91e309b50e

Now the only relevant mutable reference change is:
```diff
@@ -37491,7 +37491,7 @@ impl<'a> ::std::ops::DerefMut for PipelineCreationFeedbackCreateInfoEXTBuilder<'
 impl<'a> PipelineCreationFeedbackCreateInfoEXTBuilder<'a> {
     pub fn pipeline_creation_feedback(
         mut self,
-        pipeline_creation_feedback: *mut PipelineCreationFeedbackEXT,
+        pipeline_creation_feedback: &'a mut PipelineCreationFeedbackEXT,
     ) -> PipelineCreationFeedbackCreateInfoEXTBuilder<'a> {
         self.inner.p_pipeline_creation_feedback = pipeline_creation_feedback;
         self
```

And converting a bunch of `&'a <opaque/void type>` to `*const`.

MaikKlein (approved):
Okay I think everything looks good :). Thanks for the cleanup.

MarijnS95:
Thanks @MaikKlein for reviewing this rather large chunk of changes, glad you like it :)

PR #341, opened 2020-11-23
Update Vulkan-Headers to 1.2.162 with stable ray-tracing spec

### LINKED ISSUE #277: Add support for the new VK_KHR_ray_tracing extension
See https://www.khronos.org/blog/ray-tracing-in-vulkan

* [ ] Generate vk.rs with the new xml
* [ ] Expose a "high level" extension for `VK_KHR_ray_tracing`
  * I expect this to be a lot of work, a basic version would be good enough for now.

MarijnS95:
Depends on #339
Fixes #277

Update to the first stable ray-tracing spec released this morning in https://github.com/KhronosGroup/Vulkan-Headers/commit/d2308183f2846a1453ce7ae8c77ea7b5a7424ad2. `VK_KHR_ray_tracing` is now separated in `VK_KHR_ray_tracing_pipeline`, `VK_KHR_ray_query` and `VK_KHR_acceleration_structure`. Safe extensions wrappers have been updated to account for this.

The only interesting change in the generator is dealing with constant-sized arrays, used in `VkAccelerationStructureVersionInfoKHR`:
```xml
<member len="2*ename:VK_UUID_SIZE" altlen="2*VK_UUID_SIZE">const <type>uint8_t</type>* <name>pVersionData</name></member>
```


generator/src/lib.rs
```diff
@@ -1705,66 +1755,70 @@ pub fn derive_setters(
...
+                                quote!(&'a #mutable_ref [#slice_type; #c_size]),
+                                quote!(),
+                            )
+                        } else {
+                            let array_size_ident = Ident::from(array_size.to_snake_case().as_str());
+                            (
+                                quote!(&'a #mutable_ref [#slice_type]),
+                                quote!(self.inner.#array_size_ident = #param_ident_short.len() as _;),
+                            )
+                        };
```
MarijnS95, line 1801: I don't really like hacking in this static size in `vkxml::ArrayType::Dynamic`, but this field is weird: it's definitely a pointer according to the `p` prefix and the header:
https://github.com/KhronosGroup/Vulkan-Headers/commit/d2308183f2846a1453ce7ae8c77ea7b5a7424ad2#diff-e222ae95c2b0d5082b94d6086fb1c24da18ee31384c1a39840df3b9152023ee6R11550, but the size of the data it points to is statically known. Perhaps we need a third state for this in `vkxml::Field`?

Otherwise I'd have implemented it something like this:
https://github.com/MaikKlein/ash/commit/a9b79a38abdde297d9e442e998da03ba441d87e0

MarijnS95: Spent some more time on this over the weekend (have an itch for perfecting "irrelevant" bits of code), and with there being no proper way to represent this `const uint8_t (* versionData)[2 * VK_UUID_SIZE]` in `vk.xml` it seems easiest to override the array type to static straight after parsing the file, or adding the "static `len` expression" heuristic to vk-parse instead.

That'd look about like this: https://github.com/Traverse-Research/ash/compare/cleanup..simplify
- Note that it includes changes to how non-const pointers in `setters` are now treated as safe references as well, and instead providing an override for "really void" things;
  - (This makes sense for things like `PipelineCreationFeedbackEXT` in the builder that are now passed as `*mut` :thinking:);
  - `ANativeWindow` should likely be void as well;
- `vkxml` really needs to add `Eq`, `Clone` (and `Copy`) to their types, but the repo hasn't been touched since 2017...
- The heuristic is wrong, and should be moved to `vk-parse` as this nested `for` is beyond unacceptable;
- Finally, `data_void` is now a proper `*const [u8; 2 * UUID_SIZE]` instead of `*const u8` in the struct :grimacing:

ash/src/extensions/khr/acceleration_structure.rs
```diff
@@ -0,0 +1,321 @@
...
+    ) {
+        assert_eq!(infos.len(), indirect_device_addresses.len());
+        assert_eq!(infos.len(), indirect_strides.len());
+        assert_eq!(infos.len(), max_primitive_counts.len());
+
+        let max_primitive_counts = max_primitive_counts
+            .iter()
+            .map(|cnt| *cnt as *const _)
+            .collect::<Vec<_>>();
+
```
MaikKlein, line 113: This pattern is slightly unfortunate. I think this could actually be safe to transmute, but safe code for now is totally fine.

MaikKlein (approved):
Looks good. I think we need a raytracing example in the ash repro now that it is official. Not needed for this PR, I'll open an issue for it.

ash/src/extensions/khr/acceleration_structure.rs
MarijnS95: Yeah that'd be better, one now has to be aware that `cnt` is `&&u32` to make the pointer cast work.

PR #347, opened 2020-12-19
Add missing doc link in AccelerationStructure

tangmi:
I was working with `ash::extensions::khr::AccelerationStructure` and noticed one method was missing the docs link!

Cheers!


MarijnS95 (approved):
Whoops, shouldn't have double-checked but triple-checked, thanks!

PR #352, opened 2020-12-29
Pass raw pointers in FFI, generalize generator type handling

MarijnS95:
Similar to #351 this change was already well in the works when a shift to a second generator was announced. Submitting a PR here just to not loose the concepts, even if it might end up not getting reviewed/merged.

The main goal of this PR is to unify handling of types between unsafe (FFI) types and their "safe" counterpart (builder functions), and remove some case-specific broken code introduced in previous PRs again. Vkxml arrays versus references are pretty troublesome to deal with, here's hoping that can be dealt with in the pure vk_parse implementation of the new generator :)

Finally, FFI signatures are adjusted to use `*const` instead of `&` which represents a "fat" pointer in certain scenarios (16-bytes).

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -2472,7 +2472,30 @@ pub fn write_source_code<P: AsRef<Path>>(vk_xml: &Path, src_dir: P) {
         })
         .collect();
 
-    let spec = vk_parse::parse_file_as_vkxml(vk_xml).expect("Invalid xml file.");
+    let mut spec = vk_parse::parse_file_as_vkxml(vk_xml).expect("Invalid xml file.");
+
+    for elem in &mut spec.elements {
+        if let vkxml::RegistryElement::Definitions(definitions) = elem {
+            for el in &mut definitions.elements {
+                if let vkxml::DefinitionsElement::Struct(_struct) = el {
+                    for mem in &mut _struct.elements {
+                        if let vkxml::StructElement::Member(field) = mem {
+                            if matches!(field.array, Some(vkxml::ArrayType::Dynamic))
+                                && field
+                                    .size
+                                    .as_ref()
+                                    // TODO: Should rather check if no identifier is a member variable
+                                    .map_or(false, |s| s.contains('*'))
+                            {
+                                field.array = Some(vkxml::ArrayType::Static);
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
     let cmd_aliases: HashMap<String, String> = spec2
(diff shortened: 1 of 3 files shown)


generator/src/lib.rs
```diff
@@ -2395,7 +2440,39 @@ pub fn write_source_code<P: AsRef<Path>>(vk_xml: &Path, src_dir: P) {
...
+                                }
+                            }
+                        }
+                        _ => {}
+                    }
+                }
+            }
+            _ => {}
+        }
+    }
```
MarijnS95, line 2474: This entire bit is here because vk.xml does not contain something like `type_t (*myvariable)[sz]`, where `[` is the only way to get it detected as a static-length array. Instead, pointers to static sized arrays are represented as a _single_ pointer with static length (`len=` and `altlen=` fields in `vk.xml`).

It should be moved to vkxml/vk_parse.

MaikKlein:
>Finally, FFI signatures are adjusted to use *const instead of & which represents a "fat" pointer in certain scenarios (16-bytes).

Uhh that is really bad. Can't really release a new version with that bug. I'll try to hammer out the generator in the next couple of days, otherwise we might have to get in the fix for the "old" generator.

MarijnS95:
@MaikKlein It shouldn't be a problem here because fat pointers are only used for DSTs: slice and trait types. `&[FragmentShadingRateCombinerOpKHR; 2]` (and `blend_constants: &[f32; 4]`) are neither of those:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=88acb9736455f7262ecf4af2077f3a59

Blend constants are in the current release already.

We can possibly go through this PR and merge it after cleaning up the override (eg. move it to an ash fork of vkxml/vk_parse?) or remove it altogether for the time being. Anyway, this issue is trivial to fix with just a single line, I'll PR it.

MaikKlein:
Ah puh hehe. I just skimmed the source code and thought that would be slices instead of arrays. Okay then we are good for now.

MarijnS95:
@MaikKlein Pushed a PR to clean it up (trivial fix) and to prevent such confusion in the future. My bad for writing it like that without clarifying what "certain scenarios" means, on master as well as in 358f4806d0679393b3f3d7e11d6416caedd2716c the situation is clearly around static-sized arrays anyway.

MarijnS95:
Most of these patches have made it in one way or another, only type unification which is just unnecessary noise in the generator that's set for a rewrite only. Closing.

PR #355, opened 2021-01-11
Update Vulkan-Headers to 1.2.166

MarijnS95:
Hi @MaikKlein 

I still had these sitting locally and recall reading somewhere that it'd be nice to keep Ash somewhat updated. Seems to be vendor changes only so I doubt how immediately useful it is though :)

### DIFF
--- a/ash/src/vk/enums.rs
+++ b/ash/src/vk/enums.rs
@@ -1054,55 +1054,30 @@ impl ObjectType {
 }
 impl ObjectType {
     pub const UNKNOWN: Self = Self(0);
-    #[doc = "VkInstance"]
     pub const INSTANCE: Self = Self(1);
-    #[doc = "VkPhysicalDevice"]
     pub const PHYSICAL_DEVICE: Self = Self(2);
-    #[doc = "VkDevice"]
     pub const DEVICE: Self = Self(3);
-    #[doc = "VkQueue"]
     pub const QUEUE: Self = Self(4);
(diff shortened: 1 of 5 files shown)


ash/src/vk/enums.rs
```diff
@@ -1054,55 +1054,30 @@ impl ObjectType {
 }
 impl ObjectType {
     pub const UNKNOWN: Self = Self(0);
-    #[doc = "VkInstance"]
```
MarijnS95, line 1057: All these `comment=` went away, we have to do a reverse lookup on `objtypeenum` for every type now if we want this comment.

 https://github.com/KhronosGroup/Vulkan-Headers/compare/d2308183f2846a1453ce7ae8c77ea7b5a7424ad2...9efc4a631161eeea11082bbd16d605cca0b7a01e#diff-a852a94832433f254f7e5ff6cf889035be373606623d793dfe4a75e535600cbeR5681

MarijnS95:
Maybe interesting, `rustdoc` recently got [taught about aliases](https://doc.rust-lang.org/rustdoc/advanced-features.html#add-aliases-for-an-item-in-documentation-search): we could generate a `#[doc(alias = "vkFunctionOrStructName"]` in addition to the `#[doc = "<URL>"]`. Currently none of our symbols are indexed by their C/API name, only `PFN_vk...` shows up.

MaikKlein (approved):

PR #356, opened 2021-01-12
Refactor triangle and texture examples to use better practices

### LINKED ISSUE #190: Examples demonstrate bad patterns
The examples have a few issues that should be corrected to help guide new Vulkan users towards good practice:

- [ ] Window resizes should be handled gracefully rather than panicing
- [ ] Validation layer should not be hardcoded on, as it can be easily enabled via environment variables
- [ ] Fences should be reused, rather than reconstructed every frame
- [ ] Presentation should be pipelined with rendering, rather than blocked on the CPU until rendering completes
- [ ] The windowing system event queue should be drained independently of frame rendering, rather than rendering one frame per event, as that leads to massive input lag and low framerates.

hgallagher1993:
Fixes #190

It's a huge commit but almost all of it is just moving around already existing code rather than newly written code. The reason it's so big is because I decided to refactor the ExampleBase so it matches https://vulkan-tutorial.com/ more closely which has the creation of everything broken down into small functions so I decided to do the same with the `ExampleBase:: new(..)` so that's pretty much where all the changed lines have come from.

As well as that I fixed the order of `destroy_buffer`, `free_memory` etc so when you close the window it doesn't error saying the device hasn't been destroyed.

And I created the `render_triangle` and `render_texture` functions just to keep the event_loop a bit smaller and less convoluted

### DIFF
--- a/examples/src/bin/texture.rs
+++ b/examples/src/bin/texture.rs
@@ -7,7 +7,17 @@ use std::os::raw::c_void;
 use ash::util::*;
 use ash::vk;
 
+use winit::{
+    dpi::{LogicalSize, PhysicalSize},
+    event::{Event, KeyboardInput, VirtualKeyCode, WindowEvent},
+    event_loop::{ControlFlow, EventLoop},
+    window::WindowBuilder
+};
+
 use examples::*;
+
+const APP_NAME: &str = "Ash - Example";
+const WINDOW_SIZE: [u32; 2] = [1024, 792];
 
 #[derive(Clone, Debug, Copy)]
 struct Vertex {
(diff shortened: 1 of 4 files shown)


examples/src/bin/texture.rs
```diff
@@ -688,149 +641,97 @@ fn main() {
...
-                    vk::Fence::null(),
-                )
-                .unwrap();
-            let clear_values = [
-                vk::ClearValue {
-                    color: vk::ClearColorValue {
-                        float32: [0.0, 0.0, 0.0, 0.0],
+        let graphics_pipeline = graphics_pipelines[0];
+
+        let mut can_render = true;
```
hgallagher1993, line 657: Not really sure why this variable is needed to be honest but without it the program segfaults when you close the window. The problem is after it goes into CloseRequested the event loop loops back round again and goes into RedrawEventsCleared where it calls the render functions with already destroyed variables hence the segfault. I'm not very experienced with winit at all but I would have assumed after CloseRequested it would just break?

Friz64: I think you need to render on the `RedrawRequested` event instead of the `RedrawEventsCleared`

Ralith: In fact, you should render in `MainEventsCleared`. [See the winit docs.](https://docs.rs/winit/0.24.0/winit/event/enum.Event.html#variant.MainEventsCleared)

Ralith:
> The reason it's so big is because I decided to refactor the ExampleBase so it matches https://vulkan-tutorial.com/

I haven't reviewed any changes yet, but this immediately gives me pause. vulkan-tutorial.com is a bit notorious for teaching poor practice; please don't take its patterns as gospel.

examples/src/bin/texture.rs
Ralith: > I would have assumed after CloseRequested it would just break

You should probably put the cleanup code in [`LoopDestroyed`](https://docs.rs/winit/0.24.0/winit/event/enum.Event.html#variant.LoopDestroyed).

hgallagher1993: Ya this solved it :+1:

hgallagher1993:
@Ralith I've seen that the build failed on Clippy and Rustfmt but I'll wait until a full review is done before submitting fixes for them?

ColdIce1605:
@hgallagher1993 eh push it

PR #361, opened 2021-01-26
Added `VK_EXT_sampler_filter_minmax` extension support

FuriouZz:


MarijnS95:
Hi!

These manually defined extension structs are used to provide safe(r) wrappers around extension functions. `VK_EXT_sampler_filter_minmax` does not seem to contain any, hence all this does is re-export the name which is also available to your application as `ash::vk::ExtSamplerFilterMinmaxFn::name()`. What is your use-case for this struct?

FuriouZz:
Hi 👋🏽 Yeah I figured that this struct was not really useful, but I saw that [ray_query](https://github.com/MaikKlein/ash/blob/master/ash/src/extensions/khr/ray_query.rs) was implemented the same way. I just thought it was a common nomenclature to expose an extension.

MarijnS95:
@FuriouZz Right, and it was added by me without noticing :facepalm: 

I'm removing it before a new crate version is released: https://github.com/MaikKlein/ash/pull/364

FuriouZz:
Haha no worry @MarijnS95. I am going to close my PR 😅. 
At the begining I wanted to work on the implemention for gfx and wgpu and I just discovered behind bunch of codes that the gfx part was already implemented and just the wgpu side is required. I lost some time too.
I was glad to contribute but finally it will be for a next time 😅🙄.

MarijnS95:
This eye-opener was worth it :tada: . There are always open issues if you feel like contributing  :wink:

PR #364, opened 2021-01-26
extensions/khr: Remove empty "safe wrapper" for RayQuery

MarijnS95:
The `VK_KHR_ray_query` extension has no functions and this safe wrapper
(specifically to re-wrap functions) was added by accident (by me :roll_eyes: ).
To prevent confusion, remove it before a new crate version is released.
Applications can easily access the name (which is the only useful thing
reexported here) from `ash::vk::extensions:KhrRayQueryFn::name()`.


MaikKlein:
Oops well I also missed it in the review :)

PR #365, opened 2021-01-27
Fix safe CmdBuildAccelerationStructuresIndirect signature and add extra length validation

MarijnS95:
Came across the original PR introducing headers 1.2.162 with the official KHR RayTracing spec, and found a slightly related comment (https://github.com/MaikKlein/ash/pull/341#discussion_r541967759) about unfortunate casts that could have been safe transmutes because references to PODs have the same size as raw pointers. Turns out these shouldn't be normal pointers but slices (fat pointers) :anguished: 

In addition this PR introduces extra asserts: we're exposing a "safe" (still marked `unsafe` though, just type reinterpretation) layer that should at the very least validate the length of this and other nested arrays.

### DIFF
--- a/ash/src/extensions/khr/acceleration_structure.rs
+++ b/ash/src/extensions/khr/acceleration_structure.rs
@@ -82,7 +82,7 @@ impl AccelerationStructure {
         let build_range_infos = build_range_infos
             .iter()
             .map(|slice| slice.as_ptr())
-            .collect::<Vec<*const _>>();
+            .collect::<Vec<_>>();
 
         self.acceleration_structure_fn
             .cmd_build_acceleration_structures_khr(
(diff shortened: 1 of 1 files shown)


MaikKlein (approved):
Lets get it in. Still a bit sad that this requires an allocation, but I don't see a way around it.

MaikKlein (approved):

MarijnS95:
@MaikKlein That has actually been slightly alleviated not to require a potential secondary allocation with https://github.com/MaikKlein/ash/issues/380#issuecomment-787103648 / https://github.com/Traverse-Research/ash/commit/a94b018b4101b97cd04fd42fa0e5d67f23a72e8d. Perhaps we should purse such solutions, or otherwise add functions that accept raw pointers like `&[*const u32]`?

PR #366, opened 2021-02-04
README update: updated link paths for vulkan sdk on mac

cultofmetatron:
I tried to build and run against the lunar sdk but kept running nto build and run issues when linking to the sdk

I dug into the file paths and found they did not match what was in the readme so I updated the paths to match the paths based on the latest release. Can confirm it builds now.


Ralith:
I believe this was subsumed by #609; feel free to reopen if not.

PR #370, opened 2021-02-11
rename parameter of get_physical_device_surface_support 

### LINKED ISSUE #369: confusing parameter name
The method `get_physical_device_surface_support` expects a `queue_index` as parameter. According to the documentation, it should rather be a `queue_family_index`.

https://docs.rs/ash/0.31.0/ash/extensions/khr/struct.Surface.html#method.get_physical_device_surface_support

Drodofsky:
fixes #369


MaikKlein:
Thanks!

PR #373, opened 2021-02-20
Implement indexing for Align<T>

AidanConnelly:
Small speculative PR.

This is useful for a variety of situations where the alternative, recreating the Align<T> with the relevant start and sizes, would involve the need to lug around information already stored in Align<T> itself.

### DIFF
--- a/ash/src/util.rs
+++ b/ash/src/util.rs
@@ -4,6 +4,7 @@ use std::marker::PhantomData;
 use std::mem::size_of;
 use std::os::raw::c_void;
 use std::{io, slice};
+use std::ops::Range;
 
 /// `Align` handles dynamic alignment. The is useful for dynamic uniform buffers where
 /// the alignment might be different. For example a 4x4 f32 matrix has a size of 64 bytes
--- a/ash/src/util.rs
+++ b/ash/src/util.rs
@@ -59,6 +59,15 @@ impl<T> Align<T> {
         }
     }
 
+    pub unsafe fn index(&self, index: Range<usize>) -> Align<T> {
+        Align {
+            ptr: (self.ptr as *mut u8).offset((index.start as u64 * self.elem_size) as isize) as *mut c_void,
+            elem_size: self.elem_size,
+            size: (index.end - index.start) as u64 * self.elem_size as u64,
+            _m: self._m,
+        }
+    }
+
     pub fn iter_mut(&mut self) -> AlignIter<T> {
         AlignIter {
             current: 0,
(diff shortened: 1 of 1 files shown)


ash/src/util.rs
```diff
@@ -59,6 +60,15 @@ impl<T> Align<T> {
         }
     }
 
+    pub unsafe fn index(&self, index: Range<usize>) -> Align<T> {
+        Align {
+            ptr: (self.ptr as *mut u8).offset((index.start as u64 * self.elem_size) as isize) as *mut c_void,
+            elem_size: self.elem_size,
+            size: (index.end - index.start) as u64 * self.elem_size as u64,
+            _m: self._m,
```
MarijnS95, line 68: You should be able to use struct update syntax to carry `elem_size` and `_m` over:

```suggestion
            ptr: (self.ptr as *mut u8).offset((index.start as u64 * self.elem_size) as isize) as *mut c_void,
            size: (index.end - index.start) as u64 * self.elem_size as u64,
            ..*self
```

ash/src/util.rs
```diff
@@ -59,6 +60,15 @@ impl<T> Align<T> {
         }
     }
 
+    pub unsafe fn index(&self, index: Range<usize>) -> Align<T> {
+        Align {
+            ptr: (self.ptr as *mut u8).offset((index.start as u64 * self.elem_size) as isize) as *mut c_void,
```
MarijnS95, line 65: Afaik `.offset` is implemented for `*mut c_void` and jumps by a byte-size of `1` - the cast to `*mut u8` and back to `*mut c_void` should not be necessary.

ash/src/util.rs
```diff
@@ -59,6 +60,15 @@ impl<T> Align<T> {
...
+            elem_size: self.elem_size,
+            size: (index.end - index.start) as u64 * self.elem_size as u64,
```
MarijnS95, line 67: `elem_size` is a `vk::DeviceSize` which should already be an alias to `u64`?

MarijnS95 (changes requested):
Would it make sense to implement [`std::ops::index`](https://doc.rust-lang.org/std/ops/trait.Index.html) at some point?

Shouldn't this function perform any kind of bounds checking? If not, perhaps that should be mentioned in `# Safety` docs?

MarijnS95:
@AidanConnelly Are you still planning to update this PR?

Ralith:
Closing as this has been abandoned.

PR #374, opened 2021-02-26
ash: Update libloading from 0.6 to 0.7

MarijnS95:
This new version marks `Library::new` as unsafe, because loading a library invokes the libraries' entrypoint which may or may not have certain preconditions and perform unsafe/undefined operations, just like any other native function loaded through `libloading`.

---

Not sure if we should wrap this in `unsafe` or mark the entire `fn with_library` as unsafe, just like all other Vulkan functions?

### DIFF
--- a/ash/Cargo.toml
+++ b/ash/Cargo.toml
@@ -11,7 +11,7 @@ documentation = "https://docs.rs/ash"
 edition = "2018"
 
 [dependencies]
-libloading = { version = "0.6.1", optional = true }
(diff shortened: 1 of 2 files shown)


Ralith:
because you can, in principle, pass anything to `with_library` and have it be dlopened, we should probably mark it as unsafe.

MarijnS95:
@Ralith Indeed, I was already expecting the answer to my question to be yes. Implemented that way now, it's worth the API change IMO.

PR #379, opened 2021-02-27
Remove unnecessary CString allocation when loading functions

MarijnS95:
CString::new needs to perform an allocation to own the `&'static str` and append a NULL byte to it. We can instead perform this "allocation" in the generator and emit a byte-string literal with trailing NULL byte that can immediately be used by the function loader.

---

~~Note this is just a local test I wrote up. Not sure if it should go in but~~ It might be beneficial towards all the embedded questions/issues where such allocations are out of the question.

### DIFF
--- a/ash/src/vk/extensions.rs
+++ b/ash/src/vk/extensions.rs
@@ -102,9 +102,9 @@ impl KhrSurfaceFn {
                 ) {
                     panic!(concat!("Unable to load ", stringify!(destroy_surface_khr)))
                 }
-                let raw_name = stringify!(vkDestroySurfaceKHR);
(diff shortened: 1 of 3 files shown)


Ralith (approved):
I, for one, am a fan of not having completely pointless mallocs even if it's off a hotpath, at least when the complexity impact is insignificant.

MarijnS95:
@Ralith Same thought here. It doesn't bring us much closer to `no_std` yet nor do I depend on using `ash` that way, but the sheer thought of saving a bunch of runtime allocations is satisfying.

MaikKlein (approved):

PR #381, opened 2021-03-02
README: Add lifetime-preserving builder example using slice::from_ref

MarijnS95:
It isn't too uncommon to pass a single builder element into another builder, and `slice::from_ref` fits perfectly while preserving lifetime of the builder without calling `.build()`. We have used this successfully in our codebase to uncover and prevent numerous use-after-free bugs, and thought it's worth to share on the front page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ let queue_info = [vk::DeviceQueueCreateInfo::builder()
     .queue_priorities(&priorities)
     .build()];
 
-// We don't need to call build here because builders implement `Deref`.
+// We don't need to call `.build()` here because builders implement `Deref`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,6 +89,18 @@ let device: Device = instance
     .unwrap();
 ```
 
+To not loose this lifetime single items can be "cast" to a slice of length _one_ with `std::slice::from_ref` while still taking advantage of `Deref`:
```
MaikKlein, line 92: Interesting, I didn't actually knew that you could do that.

Btw it is "lose"

MarijnS95: Ah, great catch, sorry!

MaikKlein (approved):

PR #383, opened 2021-03-05
generator: Only expect vendor name at the end of constants

MarijnS95:
Constants (enumeration bitflags) containing the word `external` were not stripped of their common typename, and quick debugging showed the word `VKTERNAL_...` in `struct_name`. `_EXT` would be the first extension to match in `VK_EXTERNAL_...` messing up `struct_name` with a global `.replace`, in turn not removing it from the variant name.


MaikKlein (approved):
Good catch

PR #384, opened 2021-03-05
generator: Only replace "VK_" at the start of enum variant names

MarijnS95:
Replacing the occurence of "VK" is dangerous as shown by this diff; MVK shorthands for MoltenVK are accidentally mangled. Using `strip_prefix/suffix` is not only more correct and performs more runtime checks, it is also more performant by returning a slice into the string.


generator/src/lib.rs
```diff
@@ -1302,12 +1303,15 @@ pub fn variant_ident(enum_name: &str, variant_name: &str) -> Ident {
...
-        .trim_matches('_')
-        .to_shouty_snake_case()
-        .replace("_BIT", "")
-        .replace(vendor, "");
+    let new_variant_name = variant_name
+        .strip_prefix(struct_name)
+        .unwrap_or_else(|| variant_name.strip_prefix("VK").unwrap());
+    // Both of the above strip_prefix leave a leading `_`:
+    let new_variant_name = new_variant_name.strip_prefix("_").unwrap();
+    let new_variant_name = new_variant_name.trim_end_matches(vendor);
```
MarijnS95, line 1311: Note that this is not completely correct like `.replace` as it'll repeatedly strip the vendor suffix. There are some enumerations like `VK_DEBUG_REPORT_OBJECT_TYPE_VALIDATION_CACHE_EXT_EXT` that should really only have one of two postfixes stripped.

However, `vkxml` doesn't contain the alias from `VK_DEBUG_REPORT_OBJECT_TYPE_VALIDATION_CACHE_EXT` to `VK_DEBUG_REPORT_OBJECT_TYPE_VALIDATION_CACHE_EXT_EXT` so it's actually convenient that it's replaced twice - there are no overlapping names anyway.

MaikKlein: > However, vkxml doesn't contain the alias from VK_DEBUG_REPORT_OBJECT_TYPE_VALIDATION_CACHE_EXT to VK_DEBUG_REPORT_OBJECT_TYPE_VALIDATION_CACHE_EXT_EXT so it's actually convenient that it's replaced twice - there are no overlapping names anyway.

Had a quick peek at the xml, and it seems to be in there? But yeah it only should be replaced once I think.

```xml
<enum               name="VK_DEBUG_REPORT_OBJECT_TYPE_VALIDATION_CACHE_EXT" alias="VK_DEBUG_REPORT_OBJECT_TYPE_VALIDATION_CACHE_EXT_EXT" comment="Backwards-compatible alias containing a typo"/>
```

Seems to be correct on main. Am I missing something?

```rust
    pub const VALIDATION_CACHE_EXT: Self = Self(33);
    /// Backwards-compatible alias containing a typo
    pub const VALIDATION_CACHE: Self = Self::VALIDATION_CACHE_EXT;
```

MaikKlein (approved):

generator/src/lib.rs
MarijnS95: > > However, vkxml doesn't contain the alias [...]
> 
> Had a quick peek at the xml, and it seems to be in there? But yeah it only should be replaced once I think.

Meant to say that `vkxml` doesn't _expose_ (or `vk-parse` doesn't propagate) that XML node. Iirc I was looking at this:

https://github.com/terrybrashaw/vkxml/blob/32c366b1c44536670905b606c7577dc7b61f8236/src/lib.rs#L398-L405

> ```rust
>     pub const VALIDATION_CACHE_EXT: Self = Self(33);
>     /// Backwards-compatible alias containing a typo
>     pub const VALIDATION_CACHE: Self = Self::VALIDATION_CACHE_EXT;
> ```

Where are you seeing this? Grepped my local tree and GitHub but can't find this anywhere.

MaikKlein: Oops never mind, I accidentally grepped it from the new generator branch.

MarijnS95: Glad to know this is addressed in the new generator; you could perhaps detect various `backwards compatible` patterns in the `comment` attribute and mark these as `#[deprecated = "the comment"]`?

For now I've gone ahead and fixed it so that the proper/desired `VALIDATION_CACHE_EXT` shows up. Don't think it's worth providing the broken "alias" version if that's not supposed to be used anyway.

MaikKlein (approved):

PR #386, opened 2021-03-06
Add VK_KHR_ray_query device extension

filnet:


MarijnS95 (changes requested):
This function-less (empty) wrapper was intentionally removed in #364 because it adds nothing that isn't already accessible, like all the other function-less extensions.

filnet:
Not a problem :)

PR #390, opened 2021-03-08
Entry::new returns Err when entry point isn't found

### LINKED ISSUE #387: Return error if library functions aren't found, instead of panicking
`EntryCustom::new_custom` doesn't return any error, it assumes the given library contains the functions it needs. If they aren't present, then the function panics. That's less than ideal if smoother error handling is needed, I would prefer it if the function returned `Result` if the library isn't a valid Vulkan library in some way. This would also propagate to `Entry::new`. Could this be implemented?

Rua:
Fixes #387. The error is returned by `StaticFn::load` and then propagated by `EntryCustom::new_custom` and `Entry::new`.


ash/src/vk/features.rs
```diff
@@ -19,31 +19,22 @@ impl ::std::clone::Clone for StaticFn {
...
-                        "Unable to load ",
-                        stringify!(get_instance_proc_addr)
-                    ))
-                }
                 let cname =
                     ::std::ffi::CStr::from_bytes_with_nul_unchecked(b"vkGetInstanceProcAddr\0");
                 let val = _f(cname);
                 if val.is_null() {
-                    get_instance_proc_addr
+                    return Err(MissingEntryPoint(cname.to_owned()));
```
MarijnS95, line 32: You could possibly store `&'static str` here of the direct string literal `"vkGetInstanceProcAddr"` instead of copying, though for an unlikely error clause that might not matter too much.

ash/src/vk/features.rs
```diff
@@ -19,31 +19,22 @@ impl ::std::clone::Clone for StaticFn {
     }
 }
 impl StaticFn {
-    pub fn load<F>(mut _f: F) -> Self
+    pub fn load<F>(mut _f: F) -> std::result::Result<Self, MissingEntryPoint>
```
MarijnS95, line 22: Careful, this file is autogenerated!

https://github.com/MaikKlein/ash/blob/b3a010a3154d0cd9d4943e794a0e90916da93936/generator/src/lib.rs#L2210-L2217

We could possibly disable generating `fn load` and implement that manually elsewhere.

ash/src/vk/features.rs
```diff
@@ -54,6 +45,14 @@ impl StaticFn {
         (self.get_instance_proc_addr)(instance, p_name)
     }
 }
+#[derive(Clone, Debug)]
+pub struct MissingEntryPoint(pub std::ffi::CString);
+impl std::fmt::Display for MissingEntryPoint {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        write!(f, "missing Vulkan entry point")
```
MarijnS95, line 52: Should perhaps include `self.0` in the format so that the caller knows what failed to load? If reserving `MissingEntryPoint` purely for the `vkGetInstanceProcAddr` "entrypoint" (which should be the only dynamic symbol in a Vulkan library) it is perhaps better to remove that member entirely.

ash/src/entry.rs
```diff
@@ -181,20 +184,20 @@ impl<L> EntryCustom<L> {
...
-        }
+        })
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumerateInstanceVersion.html>"]
     /// ```rust,no_run
     /// # use ash::{Entry, vk};
     /// # fn main() -> Result<(), Box<std::error::Error>> {
-    /// let entry = Entry::new()?;
+    /// let entry = unsafe { Entry::new()? };
```
MarijnS95, line 200: Oops, my bad! Making sure the CI chokes on that (and fixing it in the end) in #392, thanks for catching!

MarijnS95 (changes requested):
Cool, thanks! Not as easy as it seems, one of the changes is in a generated file and will get lost on the next regeneration attempt. I think the CI lost its ability to run the generator and error if that results in a diff, or am I misremembering?

MarijnS95:
CI now spots missed code regeneration in #393.

ash/src/vk/features.rs
Rua: I took some inspiration from Vulkano on this, and figured there may be future cases where we want to include other function names. Then that can be changed without breaking anyone's code.

ash/src/vk/features.rs
Rua: Good point, it's very unlikely that you ever want to dynamically generate the function name, so a `&'static` will do.

ash/src/vk/features.rs
Rua: Ah, I wasn't aware. Maybe a good idea to put a big comment at the top of the file for future editors.

I don't really trust myself to go messing with the generator just yet, but there is perhaps an alternative. Instead of letting `StaticFn` do the check, it could be done by `Entry`, and leave `StaticFn` as it is. What do you think?

ash/src/vk/features.rs
MarijnS95: Sounds good, but don't forget to print it in `write!` in that case :)

Rua:
I reworked it a bit so only `Entry` does the error checking now. Unfortunately, in order to save the library name from within the closure, I had to use an owned `CString` because the argument to the closure is not `&'static`.

ash/src/entry_libloading.rs
```diff
@@ -71,13 +83,25 @@ impl EntryCustom<Arc<Library>> {
...
-        Ok(Self::new_custom(lib, |vk_lib, name| {
+        let mut library_name = None;
+        let result = Self::new_custom(lib, |vk_lib, name| {
             vk_lib
                 .get(name.to_bytes_with_nul())
                 .map(|symbol| *symbol)
-                .unwrap_or(ptr::null_mut())
-        }))
+                .unwrap_or_else(|_| {
+                    library_name = Some(name.to_owned());
```
MarijnS95, line 96: That's the name of the function, not the library?

ash/src/entry_libloading.rs
```diff
@@ -71,13 +83,25 @@ impl EntryCustom<Arc<Library>> {
...
 
-        Ok(Self::new_custom(lib, |vk_lib, name| {
+        let mut library_name = None;
+        let result = Self::new_custom(lib, |vk_lib, name| {
             vk_lib
                 .get(name.to_bytes_with_nul())
                 .map(|symbol| *symbol)
-                .unwrap_or(ptr::null_mut())
-        }))
+                .unwrap_or_else(|_| {
```
MarijnS95, line 95: I don't really like loosing error information here. Since you're extending `LoadingError` with a case specific to symbol loading, perhaps keep `libloading::Error` inside the `MissingEntryPoint` variant in addition to the name.

MarijnS95 (changes requested):
Have you tested this change for your use-case? As it stands a failure to load _any symbol_ is only handled _after_ `new_custom` completes, meaning `get_instance_proc_addr` is inevitably called leading to the original panic. In other words, in its current form this PR makes no difference to the original issue you attempt to solve.

As stated before just move `load` to `entry.rs` and return a `Result` there, that can subsequently be forwarded in `new_custom`. Then you can also use a `&'static str` again, though holding on to the original `libloading` (and pairing it with the symbol name) is even better.

Rua:
Did you mean move `StaticFn` to entry.rs, or just an `impl` block with `load` in it?

MarijnS95:
> Did you mean move `StaticFn` to entry.rs, or just an `impl` block with `load` in it?

Only the `impl` block with `fn load`; you can either create a second `load` function and leave the existing one for what it is, or make an edgecase to not generate it here:

https://github.com/MaikKlein/ash/blob/b3a010a3154d0cd9d4943e794a0e90916da93936/generator/src/lib.rs#L2210-L2217

Rua:
Unfortunately, it's proving impossible to create a `&'static CStr` because the functions needed for this aren't `const`. To do this would require `lazy_static` as an additional dependency. I can either make `MissingEntryPoint` contain a `&'static [u8]`, or go back to owned `CString`.

MarijnS95:
Since the name is hardcoded (and where possible autogenerated may you want to extend this error to other places in the future) I'm not opposed to `&'static str` and a simple `"vkGetInstanceProcAddr"`. A `&'static [u8]` with `b"vkGetInstanceProcAddr\0"` isn't too bad either. That is, in fact, what I had locally when implementing your feature request initially.

Rua:
Ok, another attempt...

MarijnS95:
Should be good now, thanks :)

ash/src/entry.rs
```diff
@@ -163,11 +164,27 @@ impl<L> EntryV1_2 for EntryCustom<L> {
...
+    ) -> std::result::Result<Self, MissingEntryPoint>
     where
         Load: FnMut(&mut L, &::std::ffi::CStr) -> *const c_void,
     {
-        let static_fn = vk::StaticFn::load(|name| load(&mut lib, name));
+        // TODO: Make this a &'static CStr once CStr::from_bytes_with_nul_unchecked is const
+        static ENTRY_POINT: &'static [u8] = b"vkGetInstanceProcAddr\0";
+
+        // Bypass the normal StaticFn::load so we can return an error
+        let static_fn = vk::StaticFn {
```
MarijnS95, line 178: Perhaps cleaner to move this to an `impl vk::StaticFn` block in this file so that it can be used by external applications as well? `vk::StaticFn::load` is after all public as well.

ash/src/entry.rs
```diff
@@ -226,3 +231,40 @@ impl<L> EntryCustom<L> {
...
+                } else {
+                    ::std::mem::transmute(val)
+                }
+            },
+        })
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct MissingEntryPoint(&'static [u8]); // Slice without the NUL terminator
```
MaikKlein, line 258: Including the bytes there is a bit odd. This will always be `vkGetInstanceProcAddr`. Any reason this is not an empty struct?

MarijnS95: This should probably be a more generic `MissingSymbol` if the name is kept.

Rua: I've changed it to an empty struct.

ash/src/entry.rs
```diff
@@ -255,16 +255,10 @@ impl vk::StaticFn {
...
 impl std::fmt::Display for MissingEntryPoint {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
-        unsafe {
-            write!(
-                f,
-                "missing Vulkan entry point: {}",
-                std::str::from_utf8_unchecked(self.0)
-            )
-        }
+        write!(f, "missing Vulkan entry point")
```
MarijnS95, line 261: Perhaps be more specific here: ``"Cannot load `vkGetInstanceProcAddr` symbol from library"``?

MarijnS95 (changes requested):

Rua:
Does anything else need to be done here?

MaikKlein (approved):

MaikKlein:
> Does anything else need to be done here?

Sorry I forgot about it, looks good. LGTM

ash/src/entry.rs
```diff
@@ -226,3 +231,34 @@ impl<L> EntryCustom<L> {
...
+    pub fn load_checked<F>(mut _f: F) -> Result<Self, MissingEntryPoint>
+    where
+        F: FnMut(&::std::ffi::CStr) -> *const c_void,
+    {
+        // TODO: Make this a &'static CStr once CStr::from_bytes_with_nul_unchecked is const
+        static ENTRY_POINT: &[u8] = b"vkGetInstanceProcAddr\0";
+
+        Ok(vk::StaticFn {
+            get_instance_proc_addr: unsafe {
+                let cname = CStr::from_bytes_with_nul_unchecked(ENTRY_POINT);
```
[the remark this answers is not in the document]
Ralith: I'd expect all of the above to compile down the same, yeah. Might as well favor clarity.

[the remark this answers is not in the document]
MarijnS95: @Ralith What would you consider more clear for these one-off inline named bindings, `const` or `let`?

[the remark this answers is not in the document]
Ralith: I think it's best practice to favor `let` when it's applicable.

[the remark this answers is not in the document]
MarijnS95: Fair enough, I can't imagine the compiler treating a fully-`const` path any different no matter whether it's via a `let` binding, inlined completely, or assigned to a `const`.

PR #392, opened 2021-03-08
ci: Test docs in addition to `cargo t --all-targets`

MarijnS95:
Unfortunately docs are not explicitly (build-)tested as part of `--all-targets` allowing broken code to slip in as shown by #390.

Also remove the `rust` listing type which is the default, leaving only `no_run` (until the CI has a loadable Vulkan library).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,10 +28,16 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
-      - uses: actions-rs/cargo@v1
+      - name: Test all targets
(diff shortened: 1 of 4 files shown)


MarijnS95:
https://github.com/MaikKlein/ash/pull/392/checks?check_run_id=2060979768 Great, CI works, let's push the fixes :)

MarijnS95:
For future reference `cargo test` without `--all-targets` should have also tested the docs. Apparently it tests more without than with the `--all-targets`, except benches: https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection

PR #394, opened 2021-03-12
external_memory_fd: Initialize output struct with proper sType

MarijnS95:
`MemoryFdPropertiesKHR` is a struct that, despite being used as output, hsa to be initialized properly with a valid `sType` set, otherwise:

    Validation Error: [ VUID-VkMemoryFdPropertiesKHR-sType-sType ] Object 0: handle = 0x7f035464d648, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xdc6b3a0d | vkGetMemoryFdPropertiesKHR: parameter pMemoryFdProperties->sType must be VK_STRUCTURE_TYPE_MEMORY_FD_PROPERTIES_KHR. The Vulkan spec states: sType must be VK_STRUCTURE_TYPE_MEMORY_FD_PROPERTIES_KHR (https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-VkMemoryFdPropertiesKHR-sType-sType)

### DIFF
--- a/ash/src/extensions/khr/external_memory_fd.rs
+++ b/ash/src/extensions/khr/external_memory_fd.rs
@@ -40,7 +40,7 @@ impl ExternalMemoryFd {
         handle_type: vk::ExternalMemoryHandleTypeFlags,
         fd: i32,
     ) -> VkResult<vk::MemoryFdPropertiesKHR> {
-        let mut memory_fd_properties = mem::zeroed();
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Thanks!

Ralith (approved):
Thanks!

PR #395, opened 2021-03-12
extensions/khr: Add ExternalSemaphoreFd safe wrapper

MarijnS95:
Implement the two small import and export and functions, part of the `VK_KHR_external_semaphore_fd` extension.


ash/src/extensions/khr/external_semaphore_fd.rs
```diff
@@ -0,0 +1,54 @@
...
+        import_info: &vk::ImportSemaphoreFdInfoKHR,
+    ) -> VkResult<()> {
+        self.external_semaphore_fd_fn
+            .import_semaphore_fd_khr(self.handle, import_info)
+            .result()
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetSemaphoreFdKHR.html>"]
+    pub unsafe fn get_semaphore_fd(&self, get_info: &vk::SemaphoreGetFdInfoKHR) -> VkResult<i32> {
+        let mut fd = -1;
```
MaikKlein, line 40: Is there a reason this is -1?

MarijnS95: Because `-1` is an invalid FD, `0` is not. It should anyway be extremely unlikely (if not against the spec) to return `VK_SUCCESS` without changing this value. Otherwise we should add an extra check that returns `Err` when the FD is invalid?

See also its `external_memory_fd` counterpart:

https://github.com/MaikKlein/ash/blob/d8d742348e7580e74a8385a2c0ff34a4b9db60e7/ash/src/extensions/khr/external_memory_fd.rs#L30

MaikKlein: Ah its a posix fd, okay then this should be fine as is. I don't think we should go out of our ways guard against incorrect drivers.

MaikKlein (approved):

ash/src/extensions/khr/external_semaphore_fd.rs
MarijnS95: Yup:https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_external_semaphore_fd.html

> An application using external memory may wish to synchronize access to that memory using semaphores. This extension enables an application to export semaphore payload to and import semaphore payload from **POSIX file descriptors**.

PR #397, opened 2021-03-14
Add 3 missing DeviceV1_0 functions

Rua:
This adds `merge_pipeline_caches`, `queue_bind_sparse` and `get_render_area_granularity` to the `DeviceV1_0` trait. I hope I didn't accidentally edit an autogenerated file again.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -1593,6 +1593,22 @@ pub trait DeviceV1_0 {
         err_code.result_with_success(data)
     }
 
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkMergePipelineCaches.html>"]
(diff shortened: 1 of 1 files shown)


MarijnS95 (approved):
This file ("safer" bindings over function-pointer calls, or at least accepting Rust types like slices instead of raw pointers) is still written by hand, all good here :+1: 

You could possibly improve the commit title or at least commit description to make it clear what missing functions are added. This is useful when reading the commit log (here on GH or on the commandline) without attached diff.

Rua:
Done.

MarijnS95:
There are generally specific length requirements (50/72) to prevent the title rolling over into the description:

image

But this is good enough, thanks!

MaikKlein (approved):

PR #398, opened 2021-03-14
Add VK_NN_vi_surface extension

Rua:
I followed the examples of other surface extensions for this one, so hopefully it's good as-is. I was not sure whether the function names should have `_nn` at the end or not. `create_xlib_surface` doesn't end in `_khr` but `create_mac_os_surface_mvk` does have the `_mvk`. My own preference would be without the suffix, since it's already implicit in the module and struct name.

### DIFF
--- a/ash/src/extensions/mod.rs
+++ b/ash/src/extensions/mod.rs
@@ -2,4 +2,5 @@ pub mod experimental;
 pub mod ext;
 pub mod khr;
 pub mod mvk;
+pub mod nn;
 pub mod nv;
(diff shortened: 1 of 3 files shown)


MarijnS95 (approved):
> > I was not sure whether the function names should have `_nn` at the end or not. `create_xlib_surface` doesn't end in `_khr` but `create_mac_os_surface_mvk` does have the `_mvk`. My own preference would be without the suffix, since it's already implicit in the module and struct name.

Fully agree; perhaps a followup patch should remove it from `macos_surface` and `ios_surface`. We're allowed to version-wise, though some projects might not like us for it :D

MaikKlein (approved):

PR #400, opened 2021-03-14
Add VK_KHR_get_physical_device_properties2 extension

Rua:
This duplicates functions that were added to Vulkan 1.1, but in 1.0 they are only available as an extension so for completeness (and Vulkano!) I've added it.

The 1.1 functions in Ash have different signatures, namely they take an extra `&mut` argument to place the output of the function. For this version I went with using the return value (possibly in a `Vec`), which is also used by many Ash functions and feels more idiomatic. However, if it's desirable for both functions in a pair to have the same signature, then I can change either these new extension functions or the existing 1.1 functions to match, in an additional commit to this PR.


Ralith (changes requested):
The entire purpose of these functions is to support extensions via the `p_next` mechanism, which is impossible with a `Vec` return. If that's not needed, then the non-extension variants of the functions are sufficient.

Rua:
Hmm, I can follow the original Ash functions then. But I wonder if there's more idiomatic ways to do this.

Rua:
I noticed now that the functions that allow querying for an item count have a `_len` variant which does this, so I copied that example. Still not sure if this is the best design, and also whether other functions that _do_ return a `Vec` should be modified then.

Rua:
Does anything else need to be done here?

MaikKlein (approved):

PR #402, opened 2021-03-15
Update to 1.2.171 + VK_KHR_synchronization2

Caradhrass:
I made some crude manipulations to work with Flag64 dont know if its good enough.
For VK_KHR_synchronization2 I followed the  general layout of the other extensions.


MarijnS95:
Funnily enough I have right about the same change locally, with changes pending to the xml parser/wrapper to propagate `bitwidth`. More than 32 used bits seems like a good heuristic but might not be true in every case, where a 64-bit flags type would be used up front in preparation for more flags appearing in the future. Not sure if such a case exists yet.

Please reopen this PR _before_ force-pushing with the missing files.

I could possibly review this PR as is, but it is perhaps easier to share my attempt: https://github.com/MaikKlein/ash/commit/1125d2cdaa9a4ec1255b5cb5ec651c0006ee9238. Forgot about this for almost two full weeks :see_no_evil:

PR #403, opened 2021-03-15
Add extension wrapper for VK_KHR_synchronization2

Caradhrass:
I tried my best to implement Flag64 but i am not a rust god.


Caradhrass:
Thanks for taking the time really helpful.
I cant think of an easy way to get the correct flag type for formatting in the `const KNOWN[(Flag,&str)]` array.

For the present the `flag>u32::max` check seems save there are only the two flag64 types from synchronization2.

MarijnS95:
@Caradhrass Thanks for your patience, as mentioned in https://github.com/MaikKlein/ash/pull/403#discussion_r594607379 / https://github.com/MaikKlein/ash/pull/402#issuecomment-799658326 I have already been working on extending `vk-parse` with the appropriate field to distinguish the 32- and 64-bit variant, even though the `all_bits > u32::MAX` hack works for now; see #410 / #411.

As non-collaborator I am not allowed to mark _my own_ review comments as resolved; feel free to close everything that has to be fixed so that we have a clear overview of what is done.

It remains to be seen what the best order is to merge all these PRs in; we can possibly get your solution in first including the wrapper for `Synchronization2`, then I'll have to do some conflict resolution to get the proper `bitwidth` variant on top.

MarijnS95:
@Caradhrass are you available to discuss how to proceed? My plan is to take in #410 / #411 first to use updated `vk-parse` for accurately distinguishing 32- and 64-bit fields now and in the future. Feel free to update this PR afterwards to resolve conflicts, but I can do that for you while maintaining authorship.

Sounds good?

Caradhrass:
Sure, just keep the synchronization2 part of this and I am happy. 
That should stay the same regardless of the Flag64 generator used.

MarijnS95:
You've done an excellent job on the non-raw bindings so there's no reason to drop them nor change authorship :grinning: 

I'll take care of rebasing this when the generator and flags are in, thanks!

MarijnS95 (approved):

MaikKlein (approved):

PR #407, opened 2021-03-16
Remove the _mvk suffix from the two extensions' functions

Rua:
Per the discussion at #398, this brings these names in line with the other extensions, which lack the suffix.


MarijnS95 (approved):
Thanks!

Issue #344, opened 2020-12-13
Complete rewrite of the generator


MaikKlein:
The current design of the generator has become almost unmaintainable. There have been lots of changes to vk.xml and how we use it in ash in the past 5 years.

_I started a very early WIP [branch](https://github.com/MaikKlein/ash/tree/wip-generator/generator2) a couple of months ago, but haven't had much time to work on it so far. But I now have more time to spare and hopefully I can finish this up in the next couple of months._

I am going to list some of the changes/additions I have been thinking about:

* Fully switch to vk-parse, don't use the outdated vkxml anymore
* Don't use syn and quote anymore. They don't really have many benefits for pure codegen. Instead we will write directly to the files ourselves.
* The generator needs to be easily understandable and documented. Need to make it easier for people to contribute.
* Feature flags all the things: `ash` gets slower to compile every release.
* * Feature flag for every extension: By default everything will be active. But you can choose to opt-in into only the extensions are you want to use.
* * Feature flag for builders: While builders are nice, not everyone is using them and they add quite a bit of build time.
* * Feature flag provisional apis: Some extensions in vulkan are experimental, instead of generating those by default, you will have to opt-in into those
* * (Possibly?) Feature flags for the high level interface. For people that only want the low level bindings. Otherwise we still might split ash into `ash` and `ash-sys`.

* Generate **everything**: Currently only the low level part is generated. Every high level function is written by hand. Instead we will now generate 95% and only handwrite functions that are particularly difficult to generate correctly.
* * Extensions: We still can **not** automatically generate high level extensions by default because the output has to be verified by a human. Otherwise it would be too easy to run into breaking changes. 
* * * To add a new extension: It should be as easy as to add the name of the extension to a list, re-run the generator, verify the output, and potentially write some custom overrides for some functions.
* * Feature flags would have to be generated as well. (in .rs and .toml)
* * Make it easy to manually override functions that are hard to generate.
* No build.rs: I have been thinking about this for a while. Initially I wanted to move the generation into a build.rs file, but has some limitations. We can't autogenerate the feature flags for extensions, and it will introduce dev-depencies. I'd rather have `ash` remain a 0 dependency library for now.
* No huge files anymore: We need to split everything into multiple files. Github already can't show the diffs anymore for some files, and rustc/rustdoc struggle as well with files that are 50k loc.
* ...

MarijnS95:
[...]

PR #410, opened 2021-03-20
generator: Generate enums from vk_parse representation

MarijnS95:
~Depends on #409~ merged already.

This change prepares for future additions in vk_parse fields ([1]) by converting over the enum generation path from vkxml.  Most of the conversion is easy by repurposing the existing `EnumSpec` parsing logic from extension constants for all enumeration variants, with slight modifications to not bail when `extends` is not set which is specific to extension constants.

As an (unintended) added bonus this unification of the `EnumSpec` codepath allows aliases (for backwards-compatible names) to be generated as discussed earlier in [2].

[1]: https://github.com/krolli/vk-parse/pull/17
[2]: https://github.com/MaikKlein/ash/pull/384#discussion_r588693967

---

@MaikKlein you're probably wondering by now why I'm not helping pushing generator2 over the finish line. Would you mind providing a status/progress update in #344 and clarify the reason behind removing `quote!()`/`syn`, instead opting for manual string manipulation?

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -404,6 +404,9 @@ pub trait ConstantExt {
     fn constant(&self, enum_name: &str) -> Constant;
     fn variant_ident(&self, enum_name: &str) -> Ident;
     fn notation(&self) -> Option<&str>;
+    fn is_alias(&self) -> bool {
+        false
+    }
 }
 
 impl ConstantExt for vkxml::ExtensionEnum {
@@ -429,6 +432,9 @@ impl ConstantExt for vk_parse::Enum {
     }
     fn notation(&self) -> Option<&str> {
         self.comment.as_deref()
+    }
+    fn is_alias(&self) -> bool {
+        matches!(self.spec, vk_parse::EnumSpec::Alias { .. })
     }
 }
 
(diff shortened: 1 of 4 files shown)


generator/src/lib.rs
```diff
@@ -418,8 +421,25 @@ impl ConstantExt for vkxml::ExtensionEnum {
     }
 }
 
+impl ConstantExt for vk_parse::Enum {
+    fn constant(&self, enum_name: &str) -> Constant {
+        Constant::from_vk_parse_enum_spec(&self.spec, Some(enum_name), -1)
```
MaikKlein, line 426: Why is there a `-1` here?

MarijnS95: That's the default for `extension_number: i64,`, it is apparently used by things like `<enum extends="VkStructureType" extnumber="157" offset="0"          name="VK_STRUCTURE_TYPE_SAMPLER_YCBCR_CONVERSION_CREATE_INFO"/>`, where the actual value turns into `1_000_000_000 + (extnumber-1) * 1000 + offset`.

Good catch, this should be turned into an `Option<>` and unwrapped so that we won't get hidden invalid values!

MaikKlein (approved):

PR #411, opened 2021-03-20
generator: Add support for vkFlags64 and update to 1.2.170

MarijnS95:
Depends on ~#409, #410~, https://github.com/krolli/vk-parse/pull/17

This is my initial take on upgrading to .170 and beyond (see also #403), unfortunately belated due to the extra changes needed to get the `bitwidth` field in `vk-parse` and refactor the generator to use `vk_parse::Enums`. Compared to #403 this approach uses the `num` crate instead of converting all debug numbers to 64-bit, though a better solution might be to simply duplicate the function in a 32- and 64-bit version. We'll have to see how and in what order to merge them, as this PR doesn't provide a safe wrapper for the `Synchronization2` extension either. I can rebase this on top if necessary.

### DIFF
--- a/ash/src/vk/const_debugs.rs
+++ b/ash/src/vk/const_debugs.rs
@@ -2,15 +2,16 @@ use crate::vk::bitflags::*;
 use crate::vk::definitions::*;
 use crate::vk::enums::*;
 use std::fmt;
-pub(crate) fn debug_flags(
+pub(crate) fn debug_flags<Value: Into<u64> + Copy>(
     f: &mut fmt::Formatter,
-    known: &[(Flags, &'static str)],
-    value: Flags,
+    known: &[(Value, &'static str)],
+    value: Value,
 ) -> fmt::Result {
     let mut first = true;
-    let mut accum = value;
-    for (bit, name) in known {
-        if *bit != 0 && accum & *bit == *bit {
+    let mut accum = value.into();
+    for &(bit, name) in known {
+        let bit = bit.into();
+        if bit != 0 && accum & bit == bit {
             if !first {
                 f.write_str(" | ")?;
             }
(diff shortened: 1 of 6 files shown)


MarijnS95:
> Compared to #403 this approach uses the `num` crate instead of converting all debug numbers to 64-bit, though a better solution might be to simply duplicate the function in a 32- and 64-bit version.

This is the only question that needs answering (besides some code review). I'll take care of rebasing `.171` from #403 with proper attribution and probably force-push the result to that MR :)

generator/Cargo.toml
```diff
@@ -5,7 +5,9 @@ authors = ["Maik Klein <maikklein@googlemail.com>"]
 edition = "2018"
 
 [dependencies]
-vk-parse = { version = "0.5.0", features = ["vkxml-convert"] }
+# https://github.com/krolli/vk-parse/pull/17
+# vk-parse = { git = "https://github.com/krolli/vk-parse", rev = "873860cbcacd4a35f489d0d6b6dd7eb24156136a", features = ["vkxml-convert"] }
+vk-parse = { git = "https://github.com/MarijnS95/vk-parse", branch = "enums-bitwidth", features = ["vkxml-convert"] }
```
MarijnS95, line 10: I'll have to maintain this branch until either the PR to `vk-parse` is merged or we fork `vk-parse` elsewhere, perhaps directly under Maik's account.

Ralith:
IMO pulling in a dependency, especially a nontrivial one, for this case is poor style. Duplicating the function shouldn't be necessary either; just run everything through the 64 bits, casting at callsites as necessary.

generator/src/lib.rs
```diff
@@ -2334,10 +2348,16 @@ pub struct ConstantMatchInfo {
...
 }
 
-pub fn generate_const_debugs(
-    const_values: &BTreeMap<Ident, Vec<ConstantMatchInfo>>,
-) -> TokenStream {
+#[derive(Default)]
+pub struct ConstantTypeInfo {
+    values: Vec<ConstantMatchInfo>,
+    // TODO: Could also be `type: Option<Ident>,`
+    bitwidth: Option<u32>,
```
MarijnS95, line 2355: I don't really remember why I wanted this, using an integer is probably more consistent. We can always change it to contain the type in the future, necessary at all.

MarijnS95: In hindisght, the type name is the _key_ in `const_values: &BTreeMap<Ident, ConstantTypeInfo>`.

MarijnS95:
> IMO pulling in a dependency, especially a nontrivial one, for this case is poor style. Duplicating the function shouldn't be necessary either; just run everything through the 64 bits, casting at callsites as necessary.

Not a great fan of casting, neither of `num` (code turned out worse than anticipated) nor duplicating the function... Guess casting is easiest, that we should go with.

MarijnS95:
@Ralith By the way, note that we'd have to cast _every entry_ in the `const KNOWN` arrays to 64-bit, that doesn't look nearly as pretty as just duplicating a `debug_flags` and `debug_flags64` function.

generator/Cargo.toml
MarijnS95: Branch got merged, currently pointing this to `krolli/vk-parse` directly until a release is made :)

Caradhrass:
@MarijnS95 I just noticed that naming for the Flags is wrong.
vk::PipelineStageFlags2KHR::PIPELINE_STAGE_2_DRAW_INDIRECT should just be vk::PipelineStageFlags2KHR::DRAW_INDIRECT for consistency.

MarijnS95:
@Caradhrass Good catch, the generator is not able to deal with these. When using Heck to convert the stripped "type" name `VkPipelineStage2KHR` to `SHOUTY_SNAKE_CASE` it turns it into `VK_PIPELINE_STAGE2`, whereas the enum variants have an extra underscore before the number `2`, like `PIPELINE_STAGE_2`. That fails the prefix match and is hence not stripped from the variant name.

I'll think about how to solve this cleanly and consistently instead of making some sort of ugly edgecase.

MarijnS95:
@Caradhrass This is "fixed" with an ugly regex for now - and I should perhaps make sure the string starts with a word and not an underscore too.

I do intend to replace:

https://github.com/MaikKlein/ash/blob/52de90e91861c400e8a70ffe9cfeecf01a2c3ac7/generator/src/lib.rs#L1343

With an "allow"list for the few remaining offenders (afaik all backwards-compatible aliases which should not be used anymore - they make little sense within Rust enum variants _anyway_), and get future insight (read: immediate panic) on new offenders.

MarijnS95:
Allowlist implemented in https://github.com/MaikKlein/ash/pull/426, and added `_INTEL` which was previously missing :grimacing:

ash/src/vk/const_debugs.rs
```diff
@@ -2,24 +2,32 @@ use crate::vk::bitflags::*;
...
-pub(crate) fn debug_flags(
+pub(crate) fn debug_flags<
+    T: Copy
+        + fmt::Binary
+        + num::Num
+        + PartialEq
+        + std::ops::BitAnd<Output = T>
+        + std::ops::BitAndAssign
+        + std::ops::Not<Output = T>,
+>(
```
MaikKlein, line 13: Arg I don't like this very much. Especially bringing a new dependency in just for this case. I guess you could do `Flag: Into<u64>`, although you would have to cast inside the loop here I think.

MaikKlein (changes requested):

ash/src/vk/const_debugs.rs
MarijnS95: Yes, I agree. `Into<u64>` works just as well and requires less code nor the num dependency :grinning:

generator/Cargo.toml
```diff
@@ -5,13 +5,15 @@ authors = ["Maik Klein <maikklein@googlemail.com>"]
...
-vkxml = "0.3"
-nom = "6.0"
 heck = "0.3"
-proc-macro2 = "1.0"
 itertools = "0.10"
+nom = "6.0"
+once_cell = "1.7"
+proc-macro2 = "1.0"
 quote = "1.0"
+regex = "1.4"
```
MaikKlein, line 14: I am sort of ok with this in the generator. Just a bit odd because we also have nom.

generator/Cargo.toml
```diff
@@ -5,13 +5,15 @@ authors = ["Maik Klein <maikklein@googlemail.com>"]
...
+vk-parse = { git = "https://github.com/krolli/vk-parse", features = ["vkxml-convert"] }
```
MaikKlein, line 15: Do you know the status for a new release of vk-parse? Otherwise we should target a specific commit here.

MaikKlein (changes requested):

generator/src/lib.rs
```diff
@@ -2401,19 +2433,28 @@ pub fn generate_const_debugs(
         }
     });
     quote! {
-        pub(crate) fn debug_flags(f: &mut fmt::Formatter, known: &[(Flags, &'static str)], value: Flags) -> fmt::Result {
+        pub(crate) fn debug_flags(
+            f: &mut fmt::Formatter,
+            known: &[(impl Into<u64> + Copy, &'static str)],
+            value: impl Into<u64>,
+        ) -> fmt::Result {
```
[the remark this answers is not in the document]
MarijnS95: True, then we can re-use it for `value: Value` too.

generator/Cargo.toml
MarijnS95: I don't like the regex itself, there should be a simpler way to consume some trailing digits and insert an `_` before there (if it wasn't already). At least thanks to #426 any errors caused by it will be caught.

`nom` is more useful for syntax parsing, but yes, after experimenting with it in https://github.com/MaikKlein/ash/tree/autgenerate-define-exprs we could use it to split out type names in a name and optional trailing version. Perhaps we can write a cleaner parser that takes care of separating out the `vk` prefix, type name, version, "type" (bits/flags), and vendor.

generator/Cargo.toml
MarijnS95: @krolli has been actively updating his repo over the past days, and the CI seems to succeed again. Best to ask him directly how soon a crate release is possible :)

We can also tag a specific commit, yes.

MarijnS95: I've explicitly specified the latest master commit now, at https://github.com/krolli/vk-parse/commit/a133cbb330deeb5c8cea63574cb309075e43a231.

PR #412, opened 2021-03-21
Impl trait functions directly on EntryCustom/Instance/Device

### LINKED ISSUE #408: Implement extensions directly on Instance/Device instead?
Currently, Ash has separate structures for each extension, which means each one has to be loaded and kept around by the user. This is a bit awkward if there's many extensions involved, so what if the extension functions were implemented directly on `Instance` and `Device` instead? The user would only have to keep these two structures around for all their use cases.

These two objects would initialise the function pointers of all extensions internally when they are created. This is allowed by the spec; the spec dictates that asking for the pointer to an extension function that has not been requested at create time should return a null pointer. Ash itself already does this for the Vulkan core functions, which are all loaded regardless of whether the user created their Instance for that version.

If this is done, I also suggest getting rid of the `DeviceV1_0` trait and others like it, and implementing functions (both core and extension) as inherent methods on `Instance` and `Device`. These traits add extra flexibility but I'm not sure if that's actually needed for most people. And if each extension is going to be represented as a trait on `Instance` or `Device`, those two types end up with a _very_ long list of trait implementations. You could keep the functions for returning the raw function pointer structs as they are now, or expose each function struct member of `Instance` and `Device` to the user directly; the latter has my preference.

Rua:
Depends on #400.

A possible improvement to the code structure could be to create a `core` module to sit alongside `extensions`, containing `core/v1_0.rs`, `core/v1_1.rs` and `core/v1_2.rs`. This would better match the way extension code is currently structured. I think it could also be good to re-export Vulkan types from each of these.


MaikKlein (commented):
Sorry for the late reply.

Moving the core traits into methods seems good, I am not so sure about moving the extensions into core as well. Basically this tries to load every extensions when creating a instance/device. It is also a pretty big breaking change as now all the extension structs are gone + the extension functions have a vendor extension in their name.

I am going to request a few reviews to get more opinions.

Ralith:
I'm in favor of dropping the traits despite the breakage; I've never derived any benefit from them, it's just another thing to import, and empirically a recurring source of confusion for new users.

I'm ambivalent about folding the extensions in as well. It's a bit more convenient, but also a bit more error-prone. For extensions that I don't require (e.g. debug stuff) I quite like having an `Option<Extension>` that makes it very easy to ensure I don't forget to handle both cases somewhere. At minimum, we shouldn't conflate that change with the removal of traits; @Rua, if you're strongly interested in advocating for it, please split it into a separate PR.

MarijnS95:
Dropping the traits is fine by me - we're only using `1_2` and as @Ralith says it's just an extra thing to import, `ash::Device` is already a fat struct holding on to them all. If this struct could be more lightweight for embedded purposes we might want to revisit that, but that's more likely a case of some macro magic to define and load just the functions one needs _anyway_.

Moving the extensions in there should definitely be done in a separate PR, if only for cleaner history and easier review/revertability - and it's a significant breaking change that should be considered carefully while allowing "normal" PRs (introducing new features) to be released first.
This struct is already pretty large (we're holding it by value in quite a few locations in our code, something that needs to be refactored on our end) and this only makes it bigger. I don't like the idea of `Option<>` as that means we'd have to check whether the extension is available _every time it is used through the device_, instead of just once at startup.
Not that that really means anything currently since any function that fails to load is substituted with a panicking stub - I'd rather have more explicit control or at least feedback over what did and did not load - but that's also for another issue/PR.
(ie. with extensions you're AFAIK supposed to get none or all of the functions, to catch errors up-front in a Rusty way through `Result` without panicking).

MaikKlein:
> e. with extensions you're AFAIK supposed to get none or all of the functions, to catch errors up-front in a Rusty way through Result without panicking).

That is not fully true IIRC. Some functions from extensions might require other extensions to be enabled. So it is possible for some extension function pointers to not load.

MarijnS95:
> That is not fully true IIRC. Some functions from extensions might require other extensions to be enabled. So it is possible for some extension function pointers to not load.

Right that's true, we can't have a single `Result` catch-all for extensions. The XML seems to have `requires=` declarations for extensions as a whole but not individual functions though? Not sure how we should implement more strict function loading up-front then. Perhaps just bail if _all_ functions are nullpointers?

Ralith:
> That is not fully true IIRC. Some functions from extensions might require other extensions to be enabled. So it is possible for some extension function pointers to not load.

Can you give a specific example of an extension that is valid without another extension loaded, but which defines additional functions only if it's present?

Rua:
Did I misinterpret MaikKlein's reply in https://github.com/MaikKlein/ash/issues/408#issuecomment-803312457? I thought that was giving me the go-ahead for this PR as a whole.

MaikKlein (approved):
> Did I misinterpret MaikKlein's reply in #408 (comment)? I thought that was giving me the go-ahead for this PR as a whole.

Sorry I should have been clearer. 

Thank you for this PR, a much needed simplification <3. @MarijnS95 can you have a quick peek?

MarijnS95:
@MaikKlein I'll queue this up for tomorrow-morning, together with some other PRs. Will also check how it fares on our existing usecases.

MarijnS95 (approved):
Apologies, it took some time (a full work-week, to be precise) to get to this.

Overall this is a net-win for our codebase. All we had to do was remove the trait imports and bounds, it's actually great we're now acknowledging that all (core) function pointers were stored in `ash::Device` _anyway_, and there wasn't any extra compile/runtime checking to ensure that the (mandatory) functions could be loaded in the first place.

All in all the only thing that disappears with this PR is static visibility in what core functions are actually used by a code-base. That might be a big deal. @MaikKlein Do you perhaps remember why traits were used in the first place?

@Rua You should probably remove `Fixes #408` as this only removes trait bounds but does not take care of the main issue described there: unconditionally dumping all extension function pointers in `ash::Device`.

MarijnS95 (changes requested):
@Rua I do think that you need to maintain the 1.2 - 1.1 1.0 ordering in `device.rs` like you did in `instance.rs`. That way the diffstat is more bearable, and the resulting diff is much easier to read and `git blame` is still useful.

Also, conflict resolution with `master` seems to have gone wrong, we replaced all redundant type names with `Self` and _fortunately_ enabled the `clippy::use_self` warning for the entire `extensions` module.

MarijnS95:
@Rua `cargo test` runs application-defined tests, it doesn't perform a linter (nor formatting) pass as far as I'm aware - and there's no way to configure that either.

MarijnS95 (approved):

MarijnS95:
Looking at this again, it seems `Ash-Molten` cannot easily override `static_fn` in `EntryCustom` anymore: https://github.com/EmbarkStudios/ash-molten/blob/main/src/lib.rs

However, that's quite a bit of duplicated code and we might easily get away with by returning a pointer to the static `vkGetInstanceProcAddr` from a loader closure, problem solved: https://github.com/EmbarkStudios/ash-molten/pull/56

PR #415, opened 2021-03-26
Allow upper case acronyms for now

MaikKlein:
With 1.51 upper case acronyms are not allowed by default in clippy. Lets temporarily disable the lint.


MarijnS95 (approved):
Yeah that is the most sensible thing to do for autogenerated bindings, okay for us :)

PR #416, opened 2021-03-30
added missing functions to V1_0 traits

alpacahaircut:
Closes #422.

Added missing functions to DeviceV1_0 and InstanceV1_0.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -2090,4 +2090,38 @@ impl Device {
         );
         granularity
     }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetDeviceMemoryCommitment.html>"]
+    pub unsafe fn get_device_memory_commitment(&self, memory: vk::DeviceMemory) -> vk::DeviceSize {
+        let mut committed_memory_in_bytes = 0;
+        self.device_fn_1_0.get_device_memory_commitment(
+            self.handle(),
+            memory,
+            &mut committed_memory_in_bytes,
+        );
+        committed_memory_in_bytes
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetImageSparseMemoryRequirements.html>"]
+    pub unsafe fn get_image_sparse_memory_requirements(
+        &self,
+        image: vk::Image,
+    ) -> Vec<vk::SparseImageMemoryRequirements> {
+        let mut count = 0;
+        self.device_fn_1_0.get_image_sparse_memory_requirements(
+            self.handle(),
+            image,
+            &mut count,
+            ptr::null_mut(),
+        );
+        let mut data = Vec::with_capacity(count as usize);
+        self.device_fn_1_0.get_image_sparse_memory_requirements(
+            self.handle(),
+            image,
+            &mut count,
+            data.as_mut_ptr(),
+        );
+        data.set_len(count as usize);
+        data
+    }
 }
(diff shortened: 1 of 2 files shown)


ash/src/device.rs
```diff
@@ -2008,6 +2008,29 @@ pub trait DeviceV1_0 {
...
+    ) -> vk::DeviceSize {
+        let mut committed_memory_in_bytes = mem::zeroed();
+        self.fp_v1_0()
+            .get_device_memory_commitment(self.handle(), memory, &mut committed_memory_in_bytes);
+        committed_memory_in_bytes
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetImageSparseMemoryRequirements.html>"]
+    unsafe fn get_image_sparse_memory_requirements(&self, image: vk::Image) -> Vec<vk::SparseImageMemoryRequirements> {
+        let mut num = 0;
```
filnet, line 2025: should be sparse_memory_requirement_count ?

ash/src/device.rs
```diff
@@ -2008,6 +2008,29 @@ pub trait DeviceV1_0 {
             .get_render_area_granularity(self.handle(), render_pass, &mut granularity);
         granularity
     }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetDeviceMemoryCommitment.html>"]
+    unsafe fn get_device_memory_commitment(
+        &self,
+        memory: vk::DeviceMemory,
+    ) -> vk::DeviceSize {
+        let mut committed_memory_in_bytes = mem::zeroed();
```
filnet, line 2017: the mem::zeroed() seems unnecessary ?

ash/src/device.rs
alpacahaircut: Potentially. I tried to keep consistent with existing function bodies, for example `enumerate_device_extension_properties` and `enumerate_physical_devices`.

ash/src/device.rs
alpacahaircut: Yes, but does it really hurt? It is consistent with other functions that return more elaborate types and allows for boiler-plate code that is return type agnostic.

MarijnS95: Other functions using "more elaborate types" have actual _elaborate_ structs with multiple fields. Zero-initialization [is not always correct there either](https://github.com/MaikKlein/ash/pull/394) and I'd _personally_ prefer to just use `Default::default()` everywhere.

In this case however `vk::DeviceSize` is not elaborate at all; it is merely a typedef to `u64` (a 64-bit device pointer). Why don't we treat it that way and make it clear from the get-go, just use `0` like everywhere.

MarijnS95: @alpacahaircut Can you update this with `0` so that we can merge this PR?

ash/src/device.rs
MarijnS95: `get_image_sparse_memory_requirements2` uses `count` as well so there's that. Since this is based on the Vulkan headers where `pSparseMemoryRequirementCount` is used, let's go for `count` as the rest of the name is already implied by the function name.

MarijnS95: By the same rules it looks like the other `num` variables below should be renamed to `count` too.

MaikKlein:
You might have to rebase/merge from main and rerun the generator to make the CI happy. (You can ignore all the conflicts from the generated code)

MarijnS95:
No generated code was changed here so that shouldn't be a problem. But do run it through `rustfmt` to satisfy the CI :smile: 

(The generation job failing is merely an artifact of that, it _also_ performs a global `cargo fmt` and diffs the result. We might need to find a better solution to that, though it's not the end of the world)

MarijnS95 (approved):

MarijnS95:
Closes #422.

PR #420, opened 2021-04-18
generator: Use lifetime borrows instead of raw pointers in pp_geometries

### LINKED ISSUE #301: AccelerationStructureBuildGeometryInfoKHRBuilder::geometries ergonomics
[AccelerationStructureBuildGeometryInfoKHRBuilder::geometries](https://docs.rs/ash/0.31.0/ash/vk/struct.AccelerationStructureBuildGeometryInfoKHRBuilder.html#method.geometries) should accept `&'a &'a AccelerationStructureGeometryKHR` instead of `&'a *const AccelerationStructureGeometryKHR`.

Loosing lifetime here cost me half a hour of wondering what is going on because I accidentally assigned reference to temporary pointer here.
This was especially bad because just an hour ago I found bug in validation layer that caused similar crash.

MarijnS95:
Fixes #301

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1797,9 +1797,9 @@ pub fn derive_setters(
 
             if name == "ppGeometries" {
                 return Some(quote!{
-                    pub fn geometries_ptrs(mut self, geometries: &'a [*const AccelerationStructureGeometryKHR]) -> Self {
+                    pub fn geometries_ptrs(mut self, geometries: &'a [&'a AccelerationStructureGeometryKHR]) -> Self {
                         self.inner.geometry_count = geometries.len() as _;
-                        self.inner.pp_geometries = geometries.as_ptr();
+                        self.inner.pp_geometries = geometries.as_ptr() as *const *const _;
                         self
                     }
                 });
--- a/ash/src/vk/definitions.rs
+++ b/ash/src/vk/definitions.rs
@@ -41669,10 +41669,10 @@ impl<'a> AccelerationStructureBuildGeometryInfoKHRBuilder<'a> {
     }
     pub fn geometries_ptrs(
         mut self,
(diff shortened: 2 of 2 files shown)


generator/src/lib.rs
```diff
@@ -1797,9 +1797,13 @@ pub fn derive_setters(
...
             if name == "ppGeometries" {
                 return Some(quote!{
-                    pub fn geometries_ptrs(mut self, geometries: &'a [*const AccelerationStructureGeometryKHR]) -> Self {
+                    pub fn geometries_ptrs(mut self, geometries: &'a [&'a AccelerationStructureGeometryKHR]) -> Self {
                         self.inner.geometry_count = geometries.len() as _;
-                        self.inner.pp_geometries = geometries.as_ptr();
+                        assert_eq!(
+                            std::mem::size_of::<*const AccelerationStructureGeometryKHR>(),
+                            std::mem::size_of::<&'a AccelerationStructureGeometryKHR>()
+                        );
```
MarijnS95, line 1805: If only we had compile-time asserts without an external crate...

Ralith: I'm not sure this assert is justified at all, really. `AccelerationStructureGeometryKHR` is never going to be a DST, and I don't think rust references will ever not be ABI-compatible with dereffable pointers.

MarijnS95: Apologies for the paranoia. This won't be my first time debugging raw bindings where a DST (slice or dynamic-dispatch object) slips in and ruins the data going across. If only `as *const` wouldn't compile when converting a 16-byte fat pointer to a raw pointer that'd be so much better. Removed it again since this pattern is pretty well used across Ash anyway.

Ralith (approved):
Not familiar with this extension, but the change looks very straightforward.

MaikKlein (approved):

PR #421, opened 2021-04-25
Add extended dynamic state extension

filnet:


ash/src/extensions/ext/extended_dynamic_state.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdSetViewportWithCountEXT.html>"]
+    pub unsafe fn cmd_set_viewport_with_count_ext(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        viewport_count: u32,
+        p_viewports: *const vk::Viewport,
+    ) {
+        self.extended_dynamic_state_fn
+            .cmd_set_viewport_with_count_ext(command_buffer, viewport_count, p_viewports)
```
MarijnS95, line 66: These hand-written wrappers accept safe types and convert them to their unsafe variant. This should take a slice of `viewports: &[vk::Viewport]` and pass the `.len()` and `.as_ptr()` into the unsafe function.

ash/src/extensions/ext/extended_dynamic_state.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdSetScissorWithCountEXT.html>"]
+    pub unsafe fn cmd_set_scissor_with_count_ext(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        scissor_count: u32,
+        p_scissors: *const vk::Rect2D,
```
MarijnS95, line 74: Same here, should be a slice.

ash/src/extensions/ext/extended_dynamic_state.rs
```diff
@@ -0,0 +1,179 @@
...
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdBindVertexBuffers2EXT.html>"]
+    pub unsafe fn cmd_bind_vertex_buffers2_ext(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        first_binding: u32,
+        binding_count: u32,
+        p_buffers: *const vk::Buffer,
+        p_offsets: *const vk::DeviceSize,
+        p_sizes: *const vk::DeviceSize,
+        p_strides: *const vk::DeviceSize,
```
MarijnS95, line 89: These should also take references or slices, where possible with proper validation that all slices share the same length. Note that `p_sizes` and `p_strides` can be optional.

ash/src/extensions/ext/debug_marker.rs
```diff
@@ -11,7 +11,7 @@ pub struct DebugMarker {
 }
 
 impl DebugMarker {
-    pub fn new<I: InstanceV1_0, D: DeviceV1_0>(instance: &I, device: &D) -> DebugMarker {
+    pub fn new<I: InstanceV1_0, D: DeviceV1_0>(instance: &I, device: &D) -> Self {
```
MarijnS95, line 14: Thanks for these, but we should probably do them in a separate PR with `clippy::use_self` globally disabled.

MarijnS95 (changes requested):

ash/src/extensions/ext/debug_marker.rs
MarijnS95: Addendum: Also notice that you forgot to remove these from the struct constructor just below, last expression of these functions.

ash/src/extensions/ext/extended_dynamic_state.rs
filnet: Will do. I did a copy/paste of all the functions and forgot to make them safer...

ash/src/extensions/ext/extended_dynamic_state.rs
```diff
@@ -0,0 +1,200 @@
...
+            sizes.as_ptr()
+        } else {
+            ptr::null()
+        };
+        let p_strides = if let Some(strides) = strides {
+            assert!(strides.len() >= binding_count as usize);
+            strides.as_ptr()
+        } else {
+            ptr::null()
+        };
```
MarijnS95, line 111: I'd probably go for an exact match (`assert_eq!`), these are slices anyway that are very easy to construct and that way the user can't make any accidental mistakes. Then I'd just remove the `binding_count` argument altogether and derive that from the slices (that should all have equivalent length).

MarijnS95: Perhaps worth checking the other extensions, IIRC I've done it this way in the wrappers written by myself, not sure about the rest.

filnet: Done.

ash/src/extensions/ext/extended_dynamic_state.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdSetCullModeEXT.html>"]
+    pub unsafe fn cmd_set_cull_mode(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        cull_mode: vk::CullModeFlags,
+    ) {
+        self.extended_dynamic_state_fn
+            .cmd_set_cull_mode_ext(command_buffer, cull_mode);
```
MarijnS95, line 36: To be consistent with the functions below, drop the trailing `;` (that immediately enforces `cmd_set_cull_mode_ext` to return the same - in this case `()` - as the surrounding `fn cmd_set_cull_mode`).

MarijnS95 (approved):
Looks good :)

ash/src/extensions/ext/extended_dynamic_state.rs
filnet: You've got an eye for these things :)

ash/src/extensions/ext/extended_dynamic_state.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdSetDepthTestEnableEXT.html>"]
+    pub unsafe fn cmd_set_depth_test_enable(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        depth_test_enable: bool,
+    ) {
+        self.extended_dynamic_state_fn
+            .cmd_set_depth_test_enable_ext(command_buffer, depth_test_enable.into())
```
MarijnS95, line 128: Heh, I wasn't even aware we used `.into()` for `bool -> Bool32` conversion (struct builders use it) but I like it, as long as this results in `vk::FALSE` and `vk::TRUE`. Perhaps we should clean up `if update { vk::TRUE } else { vk::FALSE }` in `nv/ray_tracing.rs` some time.

ash/src/extensions/ext/extended_dynamic_state.rs
MarijnS95: Sometimes ^^

MarijnS95:
@filnet Thanks! Merged!

Not sure why the commit title reads `Add VK_KHR/EXT_extended_dynamic_state extension`, there's no `KHR` extension with the same name right? If there was it should probably have its own struct in the right place anyway.

filnet:
The title is wrong : it is a copy/paste from a similar title.
Sorry about that...

PR #422, opened 2021-04-25
added missing enumerate_device_layer_properties function to instance

acfsmith:
Hi!  Added wrapper for vkEnumerateDeviceLayerProperties(3)

### DIFF
--- a/ash/src/instance.rs
+++ b/ash/src/instance.rs
@@ -435,4 +435,23 @@ pub trait InstanceV1_0 {
         data.set_len(num as usize);
         err_code.result_with_success(data)
     }
+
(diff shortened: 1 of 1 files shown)


MarijnS95:
Wasn't this added by #416 as well?

acfsmith:
> Wasn't this added by #416 as well?

Yes, it was.  #416 has wasn't accepted probably because it also contains other changes which are subject of discussion.  While discussion in progress we can use this fix because vkEnumerateDeviceLayerProperties is required for some projects.

MarijnS95:
@acfsmith Those are very minor discussion points that sometimes keep waiting on the contributor to follow up on or a maintainer to chime in for a final say, feel free to ping a maintainer in an open PR in that case as it's usually quicker than re-reviewing a subset extracted from another PR - even if small.

I'll update+merge that PR tomorrow if it hasn't been updated yet, thanks for your contribution anyway!

PR #423, opened 2021-04-25
Fix clippy::use_self in extension modules

filnet:


ash/src/extensions/mod.rs
```diff
@@ -1,3 +1,4 @@
+#![deny(clippy::use_self)]
 pub mod experimental;
```
MarijnS95, line 2: Doesn't `cargo fmt` add a newline between these?

ash/src/extensions/experimental/amd.rs
```diff
@@ -40,6 +40,10 @@ vk_bitflags_wrapped!(
     Flags
 );
 impl fmt::Debug for GpaSqShaderStageFlags {
+    // ignore clippy::use_self false positives
+    // changing GpaSqShaderStageFlags::PS.0 to Self::PS.0 as suggested by clippy generates:
+    // error[E0401]: can't use generic parameters from outer function
+    #![allow(clippy::use_self)]
```
MarijnS95, line 46: Yeah, that's a bit unfortunate. I don't think you intended to use `#!` to apply it to the surrounding `impl fmt::Debug for GpaSqShaderStageFlags` scope? Unfortunately it doesn't seem like `#[allow(clippy::use_self)]` can be applied to just the statement or the `fn fmt` function, it has to be applied to the `impl` block as a whole.

```suggestion
// ignore clippy::use_self false positives
// changing GpaSqShaderStageFlags::PS.0 to Self::PS.0 as suggested by clippy generates:
// error[E0401]: can't use generic parameters from outer function
#[allow(clippy::use_self)]
impl fmt::Debug for GpaSqShaderStageFlags {
```

MarijnS95 (approved):
Thanks, that's exaclty what I wanted to see in https://github.com/MaikKlein/ash/pull/421#discussion_r619842077!

ash/src/extensions/mod.rs
filnet: nope

ash/src/extensions/experimental/amd.rs
filnet: Done.

PR #424, opened 2021-04-26
Improve Result's Display impl for extension values

theo3:
Previously, the `Display` impl for `vk::Result` did not include handling for extension values. For example, VK_ERROR_OUT_OF_DATE_KHR would display simply as "-1000001004".

With this change, we fall back to the `Debug` impl of the `Result` if the value is unknown (i.e. from an extension). This preserves the current nice messages for non-extension values, and the numeric output for truly unknown values, but displays the enum variant name (e.g. "ERROR_OUT_OF_DATE_KHR")  for extension-provided enum variants.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1535,7 +1535,9 @@ pub fn generate_result(ident: Ident, enum_: &vk_parse::Enums) -> TokenStream {
                 if let Some(x) = name {
                     fmt.write_str(x)
                 } else {
-                    self.0.fmt(fmt)
+                    // If we don't have a nice message to show, call the generated `Debug` impl
+                    // which includes *all* enum variants, including those from extensions.
+                    <Self as fmt::Debug>::fmt(&self, fmt)
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


MarijnS95 (approved):
Yeah, that's a much smarter way to go about this! Would you mind adding a comment clarifying that this forwards the call to the `Debug` printer generated for _all_ enum variants in `const_debugs.rs`? Thanks!

(And indeed, if that printer doesn't know the enum variant either it'll finally fall back to `self.0.fmt(fmt)` again :+1: )

generator/src/lib.rs
```diff
@@ -1535,7 +1535,7 @@ pub fn generate_result(ident: Ident, enum_: &vk_parse::Enums) -> TokenStream {
                 if let Some(x) = name {
                     fmt.write_str(x)
                 } else {
-                    self.0.fmt(fmt)
+                    <#ident as fmt::Debug>::fmt(&self, fmt)
```
MarijnS95, line 1538: `#ident` can be `Self`, to help us combat `clippy::use_self` :)

MarijnS95:
@theo3 Awesome, thanks! Merging as soon as the CI succeeds.

PR #427, opened 2021-04-30
Do not initialize simple integer types with mem::zeroed()

MarijnS95:
Following our little discussion in [1] to just use `0` instead.

This also renames the remaining few `num` variables to `count` to more closely match the rest of the codebase and the name of the parameter in Vulkan [2].

[1]: https://github.com/MaikKlein/ash/pull/416#discussion_r604260529
[2]: https://github.com/MaikKlein/ash/pull/416#discussion_r604258864

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -301,7 +301,7 @@ impl Device {
         &self,
         info: &vk::ImageSparseMemoryRequirementsInfo2,
     ) -> usize {
-        let mut count = mem::zeroed();
+        let mut count = 0;
         self.device_fn_1_1.get_image_sparse_memory_requirements2(
             self.handle(),
             info,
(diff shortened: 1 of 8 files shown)


MarijnS95:
Not sure if we should also look into initializing simple handle types in a different way (if possible at all), and just replace `mem::zeroed()` with either `MaybeUninit::zeroed()` or `Default::default()` because of the few edgecases we had where `sType` wasn't set properly. Though that is mostly to combat future copy-paste issues, assuming no current instance zero-initializes an output struct with `sType` field.

Ralith:
We do implement `Default` for handle types. What else would be necessary?

Ralith (approved):

MarijnS95:
> We do implement `Default` for handle types. What else would be necessary?

We only implement this for dispatchable handles, of which there are 5 users in total. Either handle is a newtype so we can't assign `= 0;` to them anyway. `fn null()` is generated but that requires specifying the type which I prefer to be inferred from the called native function and return type.

All I really want to achieve with this is highlighting the difference between zero-initialized structs (`mem::zeroed()`) and integer-like types such as pointers/handles, but there's really no need for that. Let's leave this as it is, thanks!

Ralith:
> We only implement this for dispatchable handles

False: https://github.com/MaikKlein/ash/blob/5bc52dcd0a707340275733a0c89261176be32053/ash/src/vk/macros.rs#L128

PR #430, opened 2021-05-08
Globally remove all allow(dead_code) exceptions and make extensions public

MarijnS95:
`allow(dead_code)` is hiding the fact that some extension functions are inadvertently not public and hence unusable by crate users, nor does it enforce clean coding practices.

In addition remove `ash/src/allocator.rs` which does not appear to be used (anymore?) and isn't in a working state anyway.

---

CC @Rua, I'm curious how you intended to use these extensions in vulkano :)


Rua:
Vulkano already uses them, so any switch to Ash would require support for them, that's all.

MarijnS95:
@Rua I meant that the functions were not accessible outside of the Ash crate before this PR :wink:

Rua:
Oops! Well, the Vulkano conversion hasn't happened yet. Still some implementation details to consider, as well as the suggestion to use Erupt instead that someone brought up.

MaikKlein (approved):

Issue #428, opened 2021-05-02
ash::vk::make_version should be deprecated and ash::vk::make_api_version should take its place


AlexApps99:
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_MAKE_VERSION.html
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_MAKE_API_VERSION.html

MarijnS95:
This only becomes relevant when Vulkan 1.2.175 is supported by Ash, where this deprecation happened. The current open PR for that [silently makes the deprecated ones disappear](https://github.com/MaikKlein/ash/pull/417/files#diff-de84e0b739496d075bbbbe18e9ff5c7b444f0b8bfe2160b07c998ead1b200844L9-R10).

I've been playing around with autogenerating `#define`s including proper handling of links and deprecation warnings, and the result is pretty neat IMO: https://github.com/MaikKlein/ash/commits/autgenerate-define-exprs. The code may be a bit brittle, but that should be okay given that this generator is on its way out again. More interesting, regenerating `.175` with this method puts everything in the right spot automatically: https://github.com/MaikKlein/ash/commit/0c2906680bed81f99401b267105c2e0525a76200#diff-de84e0b739496d075bbbbe18e9ff5c7b444f0b8bfe2160b07c998ead1b200844R9-R57

PR #431, opened 2021-05-08
generator: Parse and autogenerate version-related defines

MarijnS95:
Depends on #429 

With this PR I want to leverage existing parsing and generation infrastructure to completely automate the generation of define macros. It doesn't change much right now but will address #428 as if by magic when generating for `.175` :partying_face: 

Things we should discuss:
- [x] The change to `SHOUTY_SNAKE_CASE` - this isn't standard for `const fn`. We previously had a map to re-map `VK_MAKE_VERSION` to `make_version` but to do this consistently in this new system we'll have to collect a set of `SHOUTY_SNAKE_CASE` identifiers that were found to be `const fn` (`!define.parameters.is_empty()`) intead and map those to `snake_case`;
  EDIT: This is now taken care of by a map, bit verbose but we do that for many other types too;
- [x] Unify and normalize `cexpr` vs `hexadecimal_number`. Both can probably emit a `Literal` or `TokenStream` at this point.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1208,15 +1208,24 @@ pub fn generate_define(define: &vkxml::Define) -> TokenStream {
         .as_ref()
         .map_or(false, |c| c.contains("DEPRECATED"));
 
-    if name == "NULL_HANDLE" || deprecated {
+    if name == "NULL_HANDLE" {
         quote!()
     } else if let Some(value) = &define.value {
         str::parse::<u32>(value).map_or(quote!(), |v| quote!(pub const #ident: u32 = #v;))
     } else if let Some(c_expr) = &define.c_expression {
         if define.defref.contains(&"VK_MAKE_VERSION".to_string()) {
+            let link = khronos_link(&define.name);
             let c_expr = convert_c_expression(c_expr);
+            let deprecated = deprecated.then(|| {
+                let comment = &define.comment;
+                quote!(#[deprecated = #comment])
+            });
 
-            quote!(pub const #ident: u32 = #c_expr;)
+            quote!(
+                #deprecated
+                #[doc = #link]
+                pub const #ident: u32 = #c_expr;
+            )
         } else {
             quote!()
         }
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -3,8 +3,9 @@
 use heck::{CamelCase, ShoutySnakeCase, SnakeCase};
 use itertools::Itertools;
 use nom::{
-    alt, character::complete::digit1, delimited, do_parse, map, named, one_of, opt, pair, preceded,
-    tag, terminated,
+    alt,
+    character::complete::{digit1, hex_digit1},
+    complete, delimited, do_parse, map, named, one_of, opt, pair, preceded, tag, terminated, value,
 };
 use once_cell::sync::Lazy;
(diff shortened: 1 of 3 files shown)


generator/src/lib.rs
```diff
@@ -38,33 +42,58 @@ impl quote::ToTokens for CType {
 
 named!(ctype<&str, CType>,
     alt!(
-        tag!("ULL") => { |_| CType::U64 } |
-        tag!("U") => { |_| CType::U32  }
+        value!(CType::U64, complete!(tag!("ULL"))) |
+        value!(CType::U32, complete!(tag!("U")))
```
MarijnS95, line 46: This was never working before: the longest prefix needs to be matched first but at the same time that sets the minimum length requirement for the rest of the `alt!`, so `tag!("U")` never got matched.

Documented here, perhaps this should be a comment in-line:
https://github.com/Geal/nom/blob/38bb94e7bc45c2b4ef0c473e0e5c03ab134cdff3/src/branch/macros.rs#L93-L142

generator/src/lib.rs
```diff
@@ -38,33 +42,58 @@ impl quote::ToTokens for CType {
...
+        ),
+        |(num, typ)| (typ, {
+            // This is how lit::SynInt reconstructs an integer
+            // because proc_macros2::Literal lacks a way to construct
+            // hexadecimal literals.
+            let stream = format!("0x{}", num).parse::<TokenStream>().unwrap();
+            match stream.into_iter().next().unwrap() {
+                TokenTree::Literal(l) => l,
+                _ => unreachable!(),
+            }
```
MarijnS95, line 80: We do the same hack in `cexpr`, IMO we should just combine this all and return a `Literal` ro `TokenStream` directly from the macro.

generator/src/lib.rs
```diff
@@ -751,38 +714,47 @@ fn name_to_tokens(type_name: &str) -> Ident {
...
-                    rewrite_token_stream(group.stream()),
+                    // Discard outmost parenthesis
+                    if depth == 0 && i == 0 && group.delimiter() == Delimiter::Parenthesis {
+                        // TODO: Emit `Delimiter::Brace` instead, and remove them
+                        // where constants are emitted?
+                        Delimiter::None
+                    } else {
+                        group.delimiter()
+                    },
+                    rewrite_token_stream(group.stream(), depth + 1),
```
MarijnS95, line 738: This really is a gross hack to get rid of the outmost parenthesis in an expression to please the linter (because defines are generally of the form `#define FOO(x) ( ... the expression ... )` and we put that expression including parenthesis inside a function - that's already guarded by brackets).

generator/src/lib.rs
```diff
@@ -751,38 +714,47 @@ fn name_to_tokens(type_name: &str) -> Ident {
...
+                TokenTree::Ident(term) => {
+                    format_ident!("{}", constant_name(&term.to_string())).into()
+                }
+                TokenTree::Literal(lit) => TokenTree::Literal(
+                    if let Ok((_, (_, res))) = hexadecimal_number(&lit.to_string()) {
+                        res
+                    } else {
+                        lit
+                    },
+                ),
```
MarijnS95, line 749: Same generalization point as above. This is mainly to catch and rewrite `0xFFU` to `0xFF`(`u32`).

MaikKlein (approved):

PR #435, opened 2021-05-16
update winit dependency in examples

ManBearPig0:
Winit 0.25.0 was just released so I thought it could be used in the examples.


ManBearPig0:
Consider https://github.com/Traverse-Research/ash/compare/examples-winit-24 and https://github.com/hgallagher1993/ash/tree/examples_update instead. They are more comprehensive and use a winit version newer than 0.20.0, where most breaking changes were anyway.

PR #437, opened 2021-05-17
generator: Untangle mismatched parameter/return fn signatures in types

MarijnS95:
~Depends on #439, #436~ Merged!

With function typedefs for commands having some elements filtered out (by name) if they were previously defined already, combined with unfiltered arrays containing the parameter sets and return type for _all_ commands expanded together in `quote!` macros the wrong array elements get combined resulting in incorrect signatures.  No-one seems to use function pointer types (but we should!) which is why this has gone unnoticed for some time.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -841,48 +841,37 @@ fn generate_function_pointers<'a>(
         .iter()
         .unique_by(|cmd| cmd.name.as_str())
         .collect::<Vec<_>>();
-    // PFN function pointers are global and can not have duplicates. This can happen because there
-    // are aliases to commands
-    let commands_pfn: Vec<_> = commands
-        .iter()
-        .filter(|cmd| fn_cache.insert(cmd.name.as_str()))
-        .collect();
 
-    let function_name_raw = |name: &str| -> String {
-        if let Some(alias_name) = aliases.get(name) {
-            alias_name.to_string()
-        } else {
-            name.to_string()
-        }
-    };
-    let function_name = |name: &str| -> Ident {
-        let fn_name = function_name_raw(name);
-        format_ident!(
-            "{}",
-            fn_name.strip_prefix("vk").unwrap().to_snake_case().as_str()
-        )
-    };
-    let names: Vec<_> = commands
-        .iter()
-        .map(|cmd| function_name(&cmd.name))
-        .collect();
-    let names_ref = &names;
-    let raw_names: Vec<_> = commands
+    struct Command {
+        type_needs_defining: bool,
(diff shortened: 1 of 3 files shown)


generator/src/lib.rs
```diff
@@ -964,7 +963,20 @@ fn generate_function_pointers<'a>(
...
+        .iter()
+        .map(|cmd| {
+            if cmd.return_type.is_void() {
+                quote!()
+            } else {
+                let ret_ty_tokens = cmd.return_type.type_tokens(true);
+                quote!(-> #ret_ty_tokens)
+            }
+        })
+        .collect();
```
MarijnS95, line 978: I really hate having to duplicate this (just like the rest of the code here). Perhaps we should pre-generate a struct with all bits, and a boolean whether it needs a typedef created for it.

MaikKlein:
What exactly was broken? I assume something about the return types in fps?

MarijnS95:
> What exactly was broken? I assume something about the return types in fps?

@MaikKlein Oh yeah the dependency on previous PRs makes the diff impossible to read - the actual change in this PR is in the last commit: aeb120ffea436495b361c3d933027e653203d4a5

Both the parameters and return type were wrong for some of these.

MarijnS95:
And now that dependencies are in the total diff is usable again :)

MarijnS95:
@MaikKlein Reworded the commit/PR description, hope this more accurately represents what is going on - writing/wording good commits is extremely tough :smile:

generator/src/lib.rs
MarijnS95: Had a go at this in 0d0d221, let me know what you think! That refactor prevents this class of array-content-mismatches in generation entirely.

MarijnS95: Applied and pushed the refactor to this PR where it belongs - can also do it as a separate PR if you prefer (before or after #438).

Any reason why we were manually generating a `Clone` implementation instead of deriving it? The latter is easier and saves some code, and does not appear to be slower within margin of error:

```
   Completed ash v0.32.0 in 5.9s
```
Versus manual `impl Clone` for every pointer-struct:
```
   Completed ash v0.32.0 in 5.8s
```

MaikKlein: > Had a go at this in 0d0d221, let me know what you think! That refactor prevents this class of array-content-mismatches in generation entirely.

Looks good

> Any reason why we were manually generating a Clone implementation instead of deriving it? The latter is easier and saves some code, and does not appear to be slower within margin of error:

I am not sure actually, lets just autoderive it.

PR #446, opened 2021-06-06
Update Vulkan-Headers to 1.2.181

MarijnS95:
Couple new extensions, not much interesting.


MaikKlein:
Quick question: Was the thing with disabled extensions fixed in this release?

MarijnS95:
> Quick question: Was the thing with disabled extensions fixed in this release?

Yes, the missing type for that disabled extension was added in .181 :). I don't really know in what order we want to merge these patches, if we'd like to keep "pretty-much-useless" and usually meaningless disabled extensions in-tree at all.

PR #447, opened 2021-06-06
Use intradoc-links and explain the various loading methods

MarijnS95:
Still had this in stash locally, might be nice for the next release.

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -10,7 +10,12 @@ use std::os::raw::c_char;
 use std::os::raw::c_void;
 use std::ptr;
 
-/// Function loader
+/// Holds a custom type `L` to load symbols from (usually a handle to a `dlopen`ed library),
+/// the [`vkGetInstanceProcAddr`][vk::StaticFn::get_instance_proc_addr()] loader function from
+/// this library (in [`vk::StaticFn`]), and Vulkan's "entry point" functions (resolved with `NULL`
+/// `instance`) as listed in [`vkGetInstanceProcAddr`'s description].
+///
+/// [`vkGetInstanceProcAddr`'s description]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetInstanceProcAddr.html#_description
 #[derive(Clone)]
 pub struct EntryCustom<L> {
     static_fn: vk::StaticFn,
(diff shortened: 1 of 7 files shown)


ash/src/lib.rs
```diff
@@ -25,6 +25,12 @@
 //! # Ok(()) }
 //! ```
 //!
+//! ## Getting started
+//! Load the Vulkan library at the default location using [`Entry::new()`][EntryCustom<_>::new()],
+//! or at a custom location using [`Entry::with_library("path/to/vulkan")`][EntryCustom<_>::with_library()].
```
MarijnS95, line 30: Rustdoc doesn't seem capable of finding functions on type aliases yet :/

```rust
29 | //! Load the Vulkan library at the default location using [`Entry::new()`],
   |                                                            ^^^^^^^^^^^^^^ the type alias `Entry` has no function named `new`
```
So we're forced to specify `EntryCustom<_>::` directly.

ash/src/entry.rs
```diff
@@ -168,8 +168,10 @@ impl<L> EntryCustom<L> {
         &self.entry_fn_1_1
     }
 
-    #[deprecated = "This function is unavailable and therefore panics on Vulkan 1.0, please use `try_enumerate_instance_version` instead"]
+    #[deprecated = "This function is unavailable and therefore panics on Vulkan 1.0"]
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumerateInstanceVersion.html>"]
+    ///
+    /// Please use [`Self::try_enumerate_instance_version`] instead.
```
MarijnS95, line 174: Intradoc links don't work in deprecation warnings - perhaps we should leave the original function name in the deprecation message?

ash/src/entry.rs
```diff
@@ -10,7 +10,7 @@ use std::os::raw::c_char;
 use std::os::raw::c_void;
 use std::ptr;
 
-/// Function loader
+/// Hold `static` and `entry` functions
```
Ralith, line 13:
```suggestion
/// Holds `static` and `entry` functions
```

I don't think "`static` and `entry`" is very helpful to someone who doesn't already understand this type. Maybe a bit more verbosity is justified?

ash/src/entry.rs
Ralith: I'd favor a helpful deprecation message over a link, especially since the link doesn't go very far in this case.

Ralith (approved):
Nice improvement!

ash/src/entry.rs
MarijnS95: Reverted the deprecation message, but left the link in docs in-place.

ash/src/entry.rs
MarijnS95: Added a bit more verbosity. Library users probably don't - should not - look at the type directly, but it's useful for those implementing their own loader. Couldn't find a proper definition of "entry" functions in Vulkan docs, but the table in `vkGetInstanceProcAddr` should be enough.

Ralith (approved):

PR #451, opened 2021-06-19
Update ash-window version in README.md

al-jshen:
Update from 0.4 to 0.6. 

I read the README.md and copied the Cargo.toml line because that is just what I saw, instead of using the latest version. So I was using 0.4, and had trouble getting `create_surface` to work. Worked after updating to 0.6 with no other changes. (Very) small change but hopefully it helps others avoid the same problem in the future.

### DIFF
--- a/ash-window/README.md
+++ b/ash-window/README.md
@@ -22,7 +22,7 @@
 Interoperability between [`ash`](https://github.com/MaikKlein/ash) and [`raw-window-handle`](https://github.com/rust-windowing/raw-window-handle) for surface creation.
 
 ```toml
-ash-window = "0.4"
+ash-window = "0.6"
 ```
 
 ## Usage


MarijnS95:
We could fix this now and "document" it to not forget about updating such things in the future, or switch to a `cargo-release` process with `pre-release-replacements` like done in https://github.com/gfx-rs/rspirv/pull/204/commits/52bbe3db43f9473a9e1b73bb810af933c61dbc5a. We have a similar issue under `## Versions` in this readme, with `ash = "0.31"`.

@MaikKlein Thoughts?

MarijnS95:
This fix was addressed in #455 but another desync happened, requiring #473. We should look into `cargo-release` to auto-update this version or remove it entirely.

PR #452, opened 2021-06-22
Adds LICENSE-* files to crate subdirectories

jmacnak:
... so that the license files appear in downloaded crates.

### DIFF
--- a/ash-window/LICENSE-APACHE
+++ b/ash-window/LICENSE-APACHE
@@ -1,0 +1,1 @@
+../LICENSE-APACHE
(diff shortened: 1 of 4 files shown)


jmacnak:
For #453 .

Examples of this in other crates:
- https://github.com/vulkano-rs/vulkano/blob/master/vulkano/LICENSE-MIT
- https://github.com/rust-lang/futures-rs/blob/master/futures-core/LICENSE-MIT

MarijnS95 (approved):
Thanks!

We'll rebase this on #454 once it's in to make the CI happy, before being allowed to merge :)

PR #454, opened 2021-06-29
Update Vulkan-Headers to 1.2.184

MarijnS95:


MarijnS95:
We might want to add extra checks for [VkPipelineCacheHeaderVersionOne](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkPipelineCacheHeaderVersionOne.html#_description), as per the documentation it is packed (not `repr(C)`) and should total 32-bytes. That's currently true but might not be a given on all platforms. Should we introduce a dependency for static assertions?

MaikKlein:
> We might want to add extra checks for VkPipelineCacheHeaderVersionOne, as per the documentation it is packed (not repr(C)) and should total 32-bytes. That's currently true but might not be a given on all platforms. Should we introduce a dependency for static assertions?

So it should really be `repr(c,packed)` I guess, but it doesn't matter for this type. So not sure why it is marked as packed. This seems currently to be unused.

Hm I feel like the vk.xml should have information which types are packed.

MarijnS95:
As per the documentation:

> The C language specification does not define the packing of structure members.

So they (Khronos) don't perform any packing or checks in the C(++) header either (duh, it's generated from the same `vk.xml`). I don't think it's worth our time dealing with this now until someone shows up with a platform where this results in the wrong layout, or Khronos adds the relevant annotations to `vk.xml` for their own C bindings. Until then this discrepancy is properly documented in the Vulkan Documentation.

PR #456, opened 2021-07-20
Update Vulkan-Headers to 1.2.185

MarijnS95:


Rua:
I wonder if it would be possible to make a patch release each time the headers are updated?

MarijnS95:
@Rua That's something I plan on discussing with @MaikKlein after hiatus, same for the smaller PRs we do over time. Right now releases seem to be too much of a "big thing", and delaying them makes its contents - and the annoyance having to wait for new features and fixes to drop - even worse.

MaikKlein:
> I wonder if it would be possible to make a patch release each time the headers are updated?

Yeah that sounds reasonable. 
> after hiatus, same for the smaller PRs we do over time

Also sorry needed a quick break from open source.

MarijnS95:
> > after hiatus, same for the smaller PRs we do over time
> 
> Also sorry needed a quick break from open source.

And that's totally fine, it's what a hiatus or little holiday is for! Especially working on and switching between many different (open and closed) projects can be very taxing.

However, we should figure something out outside breaks too. It seems we're saving up and cramming too much in a single release, making it harder to release it in the end. We still have to write the release for all patches landed since the last release, it might make sense to enforce such additions for every PR where it's relevant, up front. Then we can possibly make a patch release ever "1-2 weeks" when `Vulkan-Headers` update, and/or after every few PRs to make contributions immediately visible.

Maybe it helps to have more maintainers with publish rights, or an automated publishing system like what we do at [`android-ndk-rs`](https://github.com/rust-windowing/android-ndk-rs/blob/master/.github/workflows/publish.yml): Crates are automatically published as soon as `Cargo.toml` has version bumps. We can automate version updates using `cargo-release`, but let's keep that discussion separate in https://github.com/MaikKlein/ash/pull/451#issuecomment-877086851.

PR #457, opened 2021-07-30
Support linking Vulkan directly

Ralith:
This is the preferred pattern in most environments when an application cannot function without Vulkan, as it saves the libloading dependency, eliminates an error case, and makes the Vulkan dependency visible to the OS.

Ideally I'd like to move towards this being the default, but that might be more disruptive than is justified.

Draft until I've tested this properly on Windows. The build script might need to explicitly add `VULKAN_SDK/lib` to the linker search path.


MarijnS95 (commented):
Nice, I've been intending to [propose the same](https://github.com/EmbarkStudios/ash-molten/pull/56#issuecomment-877034275) after `0.33` is out assuming we can get permission to absorb detection and static linking of `MoltenVK` through the same mechanism, effectively making `ash-molten` obsolete.

I'm still conflicted on the `linked` name, but guess it makes sense if we were to support both dynamic _and_ static linking at some point in time.

ash/src/entry_linked.rs
```diff
@@ -0,0 +1,33 @@
...
+use crate::{entry::EntryCustom, vk};
+
+/// Marker type for [`EntryLinked`]
+pub struct Linked;
+
+/// Compile-time function loader
+///
+/// Prefer this over [`Entry`](crate::Entry) in code that would otherwise panic on
+/// [`Entry::new`](crate::Entry::new) failing.
+#[cfg_attr(docsrs, doc(cfg(feature = "linked")))]
```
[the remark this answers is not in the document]
MarijnS95: Unfortunately this is still "broken", see https://github.com/rust-lang/rust/issues/83428:

image

Feature requirements are propagated to the functions _within_ types in this module, but not to the type itself. Turn this into `pub mod` and it all starts working.

Friz64:
Please be aware that the current implementation (indirect linking) is recommended in order to achieve optimal performance as per the [Vulkan Loader documentation](https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#best-application-performance-setup).

MarijnS95:
@Friz64 You probably intended to link https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#indirectly-linking-to-the-loader as https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#best-application-performance-setup is specific to linking Vulkan functions **after** having acquired `vkGetInstanceProcAddr`/`vkGetDeviceProcAddr`. This PR only concerns itself with dynamic/runtime loading of the `vkGetInstanceProcAddr` entrypoint, which should theoretically be faster if performed by the library loader instead of having to go through extra hoops in userspace to call `dlopen` + `dlsym` (though this is what `ld.so` performs under the hood...).

Friz64:
Yes, i intended to link that one instead. The documentation is somewhat confusing in this regard, but i understand now.

ash/src/entry_libloading.rs
```diff
@@ -52,7 +52,11 @@ impl From<MissingEntryPoint> for LoadingError {
     }
 }
 
-/// Default function loader
+/// Runtime function loader
+///
+/// Prefer [`EntryLinked`](crate::EntryLinked) in code that would would otherwise panic on
+/// [`Entry::new`] failing.
```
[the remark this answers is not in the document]
Friz64: This is my biggest concern as the Vulkan library will need to be present during linking and thus break CIs. One could indeed install the Vulkan Loader on the CI server somehow but i think that's too much of a hassle.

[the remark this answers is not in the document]
MarijnS95: @Friz64 I'm curious if this isn't the case already. GH Actions - which a lot of Rust crates use - is already loaded to the brim with packages.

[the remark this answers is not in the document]
Friz64: Good thought, but [does not look like that is the case, unfortunately](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md).

ash/src/entry_linked.rs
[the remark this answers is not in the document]
Ralith: So, no action needed for now then?

ash/src/entry_linked.rs
```diff
@@ -0,0 +1,33 @@
+use std::os::raw::c_char;
+
+use crate::{entry::EntryCustom, vk};
+
+/// Marker type for [`EntryLinked`]
+pub struct Linked;
+
+/// Compile-time function loader
```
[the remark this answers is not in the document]
Ralith: Do you have a specific proposal for a better wording? I'd like to have something reasonably concise that still gives a sense of the advantages.

ash/src/entry_libloading.rs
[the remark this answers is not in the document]
Ralith: > If all these questions are answered with no we should probably seek to move in this direction by default, instead of using libloading by default.

I agree. The only reason I'm aware of to use dlopen for Vulkan is if you have a graceful fallback path, which is an extremely uncommon case, well suited to requiring an opt-in.

> One could indeed install the Vulkan Loader on the CI server somehow but i think that's too much of a hassle.

If this is an issue, fixing it is as simple as `apt-get install libvulkan1` or whatever--trivial. But CI seems to be passing fine with `--all-features` already. The loader is part of the base install of most distros these days, after all.

ash/src/entry_libloading.rs
```diff
@@ -52,7 +52,11 @@ impl From<MissingEntryPoint> for LoadingError {
...
+#[cfg_attr(docsrs, doc(cfg(feature = "libloading")))]
```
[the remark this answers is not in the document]
Ralith: IMO this is a must for feature-gated APIs; it's this or less-visible handwritten warnings. Unless the automagic version is imminent we should do whatever's necessary to have good docs.

ash/src/entry.rs
```diff
@@ -28,7 +28,7 @@ pub struct EntryCustom<L> {
 /// Vulkan core 1.0
 #[allow(non_camel_case_types)]
 impl<L> EntryCustom<L> {
-    pub fn new_custom<Load>(
+    pub unsafe fn new_custom<Load>(
```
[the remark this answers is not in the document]
Ralith: `load` is the relevant interface for this particular function; `StaticFn::get_instance_proc_addr` is not directly exposed, but fetched internally via `load`, which might return garbage.

ash/src/entry_linked.rs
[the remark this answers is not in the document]
MarijnS95: `Function loader from dynamically linked entrypoint` (`s/dynamically linked/runtime-loaded` for `Entry`?

ash/src/entry_libloading.rs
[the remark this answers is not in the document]
MarijnS95: > I agree. The only reason I'm aware of to use dlopen for Vulkan is if you have a graceful fallback path, which is an extremely uncommon case, well suited to requiring an opt-in.

Exactly this :+1: 
 
> > One could indeed install the Vulkan Loader on the CI server somehow but i think that's too much of a hassle.
> 
> If this is an issue, fixing it is as simple as `apt-get install libvulkan1` or whatever--trivial. But CI seems to be passing fine with `--all-features` already. The loader is part of the base install of most distros these days, after all.

I couldn't find it in the same page @Friz64 linked, and `--all-features` on `check` does not perform the final linking step where the library is needed, we should use `cargo build` for that. That's probably a good thing, and for projects needing to spit out built artifacts they can easily add it with `apt` - that's the justification for [GH actions not including `Vulkan` (the SDK) in the first place](https://github.com/actions/virtual-environments/issues/18).

ash/src/entry_linked.rs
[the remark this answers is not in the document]
MarijnS95: Nope, it's unfortunate but it is what it is. I hope to notice the GH link whenever that issue is closed and move it at some point.

Ralith:
> I'm still conflicted on the linked name, but guess it makes sense if we were to support both dynamic and static linking at some point in time.

None of this code is sensitive to whether the linked library is static or dynamic; it's just a matter of what's in your search path.

ash/src/entry_libloading.rs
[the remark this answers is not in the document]
MarijnS95: Let's keep it for now, it's better than nothing at all. Glad docs.rs uses nightly to build at least.

ash/src/entry_libloading.rs
```diff
@@ -52,7 +52,11 @@ impl From<MissingEntryPoint> for LoadingError {
...
 pub type Entry = EntryCustom<Arc<Library>>;
```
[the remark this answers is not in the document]
Ralith: If we make linking the loader the default I'm in favor of this. For simplicity, we might even consider something [like I'm doing for OpenXR](https://github.com/Ralith/openxrs/blob/master/openxr/src/entry.rs#L262-L266), where there's a single monomorphic `Entry` type that has a cfg-guarded library guard.

MarijnS95:
> > I'm still conflicted on the linked name, but guess it makes sense if we were to support both dynamic and static linking at some point in time.
> 
> None of this code is sensitive to whether the linked library is static or dynamic; it's just a matter of what's in your search path.

It feels like there should be something to distinguish it from runtime linking/loading, perhaps renaming `libloading` because that's rather vague if unfamiliar with the particular crate name (and anyway the only feature - and selected by default - so I don't think anyone has that explicitly listed in their `Cargo.toml`).

Ralith:
`dlopen` isn't any kind of linking at all, so I think `linked` is a pretty apt distinction; renaming the other feature makes sense, and shouldn't be very disruptive since it's defaulted on.

ash/src/entry_libloading.rs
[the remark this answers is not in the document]
Ralith: Switched to linking by default.

ash/src/entry.rs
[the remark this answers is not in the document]
Ralith: Added safety docs.

ash/src/entry_libloading.rs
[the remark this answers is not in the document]
Ralith: Shuffled the names around.

ash/src/entry_linked.rs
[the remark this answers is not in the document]
Ralith: Updated in that spirit, thanks!

Ralith:
I've added a commit which folds all the various entry types into a single monomorphic `Entry` struct that contains a lib guard iff runtime loading is enabled, just like I do in the OpenXR bindings. I think this is a big win for API simplicity.

Ralith:
Findings from CI: libvulkan.so.1 (which is what ends up in the binary's RUNPATH) *is* installed by default, but libvulkan.so (which is the symlink the linker finds at compiletime) comes from libvulkan-dev, which is not.

Rua:
The `Entry` API currently assumes that you will be using the "fancy" wrappers, but Vulkano uses only the bare wrappers, `EntryFnV1_0` and so on. Therefore, Vulkano has kept using its original loader for the moment. Would it be possible to change the loading mechanism in such a way that it can be used with the bare wrappers as well? Then Vulkano could use it. Thanks!

Ralith:
What exactly about the current iteration poses a problem for you? You've always been able to access the function pointer table directly via e.g. `Entry::fp_v1_0`.

Rua:
True, but it feels a bit weird to construct an `Entry` when I don't intend to use it at all. Separating the loading mechanism from the rest of the Vulkan API seems cleaner.

Ralith:
How exactly do you not intend to use it? `Entry` loads function pointers, which is what you're asking for access to.

examples/src/lib.rs
```diff
@@ -6,8 +6,9 @@ use ash::extensions::{
     khr::{Surface, Swapchain},
 };
 
-use ash::{vk, Entry};
-pub use ash::{Device, EntryCustom, Instance};
+use ash::vk;
+use ash::Entry;
```
MarijnS95, line 10: That's a strange change.

ash/src/lib.rs
```diff
@@ -26,22 +27,18 @@
...
 //!
 //! ## Getting started
-//! Load the Vulkan library at the default location using [`Entry::new()`][EntryCustom<_>::new()],
-//! or at a custom location using [`Entry::with_library("path/to/vulkan")`][EntryCustom<_>::with_library()].
-//! These loaders use [`libloading`]. If you wish to perform function loading yourself
-//! call [`EntryCustom::new_custom()`] with a closure turning function names
-//! into function pointers.
+//!
+//! Load the Vulkan library linked at compile time using [`Entry::new`], or at load it at runtime
+//! using [`Entry::load`], which uses `libloading`.
```
MarijnS95, line 32:
```suggestion
//! Load the Vulkan library linked at compile time using [`Entry::new()`], or at load it at runtime
//! using [`Entry::load()`], which uses `libloading`.
```

ash/src/entry.rs
```diff
@@ -2,55 +2,153 @@ use crate::instance::Instance;
...
+        })
+    }
+
+    /// Load entry points based on an already-loaded [`vk::StaticFn`]
+    ///
+    /// # Safety
+    ///
+    /// `static_fn` must contain valid function pointers that comply with the semantics specified by
+    /// Vulkan 1.0, which must remain valid for at least the lifetime of the returned [`Entry`].
+    pub unsafe fn from_static_fn(static_fn: vk::StaticFn) -> Self {
```
MarijnS95, line 136: I like this better than the previous implementation forcibly loading `b"vkGetInstanceProcAddr\0"` through the custom loader function; this truly allows a custom function to be passed without hackery like https://github.com/EmbarkStudios/ash-molten/blob/5e30540b75954160c5a62d951a65f0c9816fff15/src/lib.rs#L91-L97.

(But besides manually constructing `vk::StaticFn` an application is still completely free to call `vk::StaticFn::load(_checked)` :tada:)

ash/src/entry.rs
```diff
@@ -2,55 +2,153 @@ use crate::instance::Instance;
...
-        mut lib: L,
-        mut load: Load,
-    ) -> std::result::Result<Self, MissingEntryPoint>
-    where
-        Load: FnMut(&mut L, &::std::ffi::CStr) -> *const c_void,
-    {
-        // Bypass the normal StaticFn::load so we can return an error
-        let static_fn = vk::StaticFn::load_checked(|name| load(&mut lib, name))?;
+impl Entry {
+    /// Load from a linked Vulkan loader
```
MarijnS95, line 32: Maybe

```suggestion
    /// Load entrypoint from a linked Vulkan loader
```

?

This could help clarify the `Note` below.

ash/src/lib.rs
```diff
@@ -26,22 +27,18 @@
 //! ```
 //!
 //! ## Getting started
-//! Load the Vulkan library at the default location using [`Entry::new()`][EntryCustom<_>::new()],
-//! or at a custom location using [`Entry::with_library("path/to/vulkan")`][EntryCustom<_>::with_library()].
-//! These loaders use [`libloading`]. If you wish to perform function loading yourself
-//! call [`EntryCustom::new_custom()`] with a closure turning function names
-//! into function pointers.
```
MarijnS95, line 33: Perhaps leave this last sentence in place, referring to ``[`Entry::from_static_fn()`]`` instead?

.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --workspace --all-targets
+          args: --workspace --all-targets --all-features
```
MarijnS95, line 19: While the features can technically be used in parallel, should we `check` the possible combinations in isolation? (ie `linked`, `loaded`, and `linked,loaded`)

ash/src/entry.rs
```diff
@@ -2,55 +2,153 @@ use crate::instance::Instance;
...
+                get_instance_proc_addr: vkGetInstanceProcAddr,
+            })
+        }
+    }
+
+    /// Load default Vulkan library for the current platform
+    ///
+    /// # Safety
+    /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
+    /// for [`Library::new`] and [`Library::get`] apply here.
```
MarijnS95, line 68:
```suggestion
    /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
    /// for [`Library::new()`] and [`Library::get()`] apply here.
```

(I forgot those initially)

ash/src/entry.rs
```diff
@@ -2,55 +2,153 @@ use crate::instance::Instance;
...
+        const LIB_PATH: &str = "libvulkan.dylib";
+
+        Self::load_from(LIB_PATH)
+    }
+
+    /// Load Vulkan library at `path`
+    ///
+    /// # Safety
+    /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
+    /// for [`Library::new`] and [`Library::get`] apply here.
```
MarijnS95, line 110:
```suggestion
    /// for [`Library::new()`] and [`Library::get()`] apply here.
```

ash/src/entry.rs
```diff
@@ -107,16 +205,15 @@ impl<L> EntryCustom<L> {
         &self,
         create_info: &vk::InstanceCreateInfo,
         allocation_callbacks: Option<&vk::AllocationCallbacks>,
-    ) -> Result<Instance, InstanceError> {
+    ) -> VkResult<Instance> {
```
MarijnS95, line 208: :+1:

ash/src/entry.rs
```diff
@@ -2,55 +2,153 @@ use crate::instance::Instance;
...
+        Ok(Self {
+            _lib_guard: Some(lib),
+            ..Self::from_static_fn(static_fn)
+        })
+    }
+
+    /// Load entry points based on an already-loaded [`vk::StaticFn`]
+    ///
+    /// # Safety
+    ///
```
MarijnS95, line 133: Nit: None of the `# Safety` markdown headers have an empty line after them (this anyway isn't visible in rendered documentation).

```suggestion
    /// # Safety
```

MarijnS95:
> `dlopen` isn't any kind of linking at all, so I think `linked` is a pretty apt distinction; renaming the other feature makes sense, and indeed shouldn't be very disruptive.

While I would technically agree, this isn't the first time I have seen it being called a [`dynamic linking loader`](https://linux.die.net/man/3/dlopen).

> I've added a commit which folds all the various entry types into a single monomorphic `Entry` struct that contains a lib guard iff runtime loading is enabled, just like I do in the OpenXR bindings. I think this is a big win for API simplicity.

Should be fine, it's extremely unlikely for the features to be used together. In the rare event that someone wishes to use multiple Vulkan libraries at the same time (building whatever arbitration and circumventing the ICD) they better use runtime loading to keep the libraries separated.

> Findings from CI: libvulkan.so.1 (which is what ends up in the binary's RUNPATH) _is_ installed by default, but libvulkan.so (which is the symlink the linker finds at compiletime) comes from libvulkan-dev, which is not.

Fortunately we're already using `libvulkan.so.1` in `entry_libloading.rs`, not `libvulkan.so`, but that isn't easily selectable with `println!("cargo:rustc-link-lib=vulkan");` I suppose? Pretty strange of Ubuntu to provide it that way...

> The `Entry` API currently assumes that you will be using the "fancy" wrappers, but Vulkano uses only the bare wrappers, `EntryFnV1_0` and so on. Therefore, Vulkano has kept using its original loader for the moment. Would it be possible to change the loading mechanism in such a way that it can be used with the bare wrappers as well? Then Vulkano could use it. Thanks!

@Ralith It appears @Rua is referring to this shorthand:

https://github.com/MaikKlein/ash/blob/17149bd791cd6e09b145063a238d3e27d855780c/ash/src/entry.rs#L159-L166

Which allows all structs in `ash/src/extensions` to accept an `Entry` to call `get_instance_proc_addr()` on, instead of the raw `vk::StaticFn`. I suppose Vulkano is keeping track of the various `vk::*Fn*` structs on its own?

But that has always been the case. Not sure if it should change towards `StaticFn` as part of this PR, the lines are already being touched though. In the end I'm in favour of this change as Ash should just accept the bare minimum (`StaticFn` in this case) in its API.

MarijnS95 (commented):
This is looking really good so far. Should we let the CI run on it? I'm curious what others think on it, @MaikKlein? It's quite an impactful change, but I'd love to roll this out soon.

.github/workflows/ci.yml
Ralith: They don't really interact, and in general testing every combination of features is combinatorically infeasible, so my inclination is not to bother.

Ralith (commented):
> Should we let the CI run on it?

CI seems to be running on it fine here? Maybe that's only visible to me. Will un-draft it once I've satisfied my concerns about the Windows build.

> @Ralith It appears @Rua is referring to this shorthand:

I don't really follow. What capability is missing from the refactored `Entry`?

> Pretty strange of Ubuntu to provide [libvulkan.so in the -dev package]

I think this is reasonable for them to do, as the unversioned .so symlink is only referenced at linktime. If we somehow hardcoded the .so.1 in the linker args that would just make us fragile and unidiomatic.

> it's extremely unlikely for the features to be used together

To be clear, they can still be used together with the unified type; cargo features must be composable, after all.

> While I would technically agree, this isn't the first time I have seen it being called a dynamic linking loader.

Hah, good cite. I'm open to clearer naming ideas.

Ralith:
Added some magic to build.rs to make this Just Work with a Vulkan SDK install on Windows (and Linux, if for some reason you aren't using the proper distro packages).

Ralith:
If anyone has any ideas what, if any, special measures are needed for good behavior on macOS, let me know!

ash/build.rs
```diff
@@ -0,0 +1,24 @@
...
+                ("windows", "64") => "Lib",
+                _ => "lib",
+            };
+            println!("cargo:rustc-link-search={}/{}", var, suffix);
+        }
+        let lib = match &*target_family {
+            "windows" => "vulkan-1",
+            _ => "vulkan",
+        };
+        println!("cargo:rustc-link-lib={}", lib);
```
MaikKlein, line 22: Can you remind me what kind of linking this is? Will this always be static linking?

MaikKlein: fwiw only having static linking is fine I think, I don't think we need dynamic linking because we have runtime loading which are fairly similar.

MaikKlein (commented):
Awesome, I don't have any concerns here.

> If anyone has any ideas what, if any, special measures are needed for good behavior on macOS, let me know!

Not sure myself. I am also not sure how this will integrate with [ash-molten](https://github.com/EmbarkStudios/ash-molten/).

I'd like to keep ash-molten as a separate crate because it is quite opinionated an builds or downloads moltenvk itself.

I think we can keep ash-molten as is and it can internally do `Entry::from_static_fn`. I think we can use no-default-features there as well and let ash-molten statically link with vulkan.

ash/build.rs
Ralith: Per [rustc docs](https://doc.rust-lang.org/rustc/command-line-arguments.html#option-l-link-lib), this will link whatever type of library is available, favoring dynamic if both are found. Note that the Vulkan loader is normally distributed only as a dynamic library.

> I think we can keep ash-molten as is and it can internally do `Entry::from_static_fn`. I think we can use no-default-features there as well and let ash-molten statically link with vulkan.

This should work great.

.github/workflows/ci.yml
MarijnS95: There are only two features that are intended to be used separately, but fine to leave as-is.

MarijnS95:
> > Should we let the CI run on it?
> 
> CI seems to be running on it fine here? Maybe that's only visible to me. Will un-draft it once I've satisfied my concerns about the Windows build.

Yeah, that. Good to see it run now.

> > @Ralith It appears @Rua is referring to this shorthand:
> 
> I don't really follow. What capability is missing from the refactored `Entry`?

Nothing that this PR changes, afaik. They must not be using `Entry` at all for whatever reason. The only change I can see being useful is to require just `StaticFn` in extension constructors instead of the "full" `Entry` which is not technically required to call `StaticFn::get_instance_proc_addr`.

> > Pretty strange of Ubuntu to provide [libvulkan.so in the -dev package]
> 
> I think this is reasonable for them to do, as the unversioned .so symlink is only referenced at linktime. If we somehow hardcoded the .so.1 in the linker args that would just make us fragile and unidiomatic.

Derp. Yes, that is totally sensible. We're already using `.so.1` for linker args though, right? Looking at `ldd $(which vkcube)` on my machine (Arch Linux) shows links against the versioned variant too. Seems like the most sane thing to do compatibility-wise.

> > it's extremely unlikely for the features to be used together
> 
> To be clear, they can still be used together with the unified type; cargo features must be composable, after all.

Exactly. That's better than making them incompatible and breaking `--all-features` in strange ways, thanks!

> > While I would technically agree, this isn't the first time I have seen it being called a dynamic linking loader.
> 
> Hah, good cite. I'm open to clearer naming ideas.

With both features enabled by default I don't think it matters too much anymore. `dynamic_linked` and `runtime_linked` are a bit too verbose.

ash/src/entry.rs
```diff
@@ -2,55 +2,152 @@ use crate::instance::Instance;
...
+    /// Load default Vulkan library for the current platform
+    ///
+    /// # Safety
+    /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
+    /// for [`Library::new()`] and [`Library::get()`] apply here.
+    ///
+    /// ```no_run
+    /// use ash::{vk, Entry};
+    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+    /// let entry = unsafe { Entry::load() }?;
```
MarijnS95, line 73: Nit: the rest of the examples have the `?` inside the `unsafe{}` block. Should we move them out in every occasion?

MarijnS95 (approved):
Ready to get in. Not sure if we should bite the bullet and accept `StaticFn` instead of `Entry` in all the extension constructors though, this PR is already touching all those lines anyway.

Ralith:
> With both features enabled by default

This PR disables dynamic loading by default, since it's a niche case and brings in an extra dependency. I think the docs still do a reasonably good job of making things clear thanks to `doc(cfg(_))` though.

> Not sure if we should bite the bullet and accept StaticFn instead of Entry in all the extension constructors though, this PR is already touching all those lines anyway.

I'd prefer to leave that for future work, since it's logically independent and you can easily construct an `Entry` given a `StaticFn` already.

ash/src/entry.rs
Ralith: This inconsistency predates this PR, but I'll go ahead and clean it up. IMO it's more readable inside the block.

MarijnS95: Fine by me, it's just an example.

MarijnS95:
Then we'll wait for @MaikKlein to give the final sign-off on this, and get it in! Let's immediately cut a new minor release as well with perhaps some other PRs?

ash/src/lib.rs
```diff
@@ -26,22 +27,19 @@
...
+//!
+//! Load the Vulkan library linked at compile time using [`Entry::new()`], or at load it at runtime
```
filnet, line 31: typo: "[...] or at load it [...]"

Ralith: Fixed, thanks!

Ralith:
@MaikKlein ping?

MaikKlein:
Sorry for the delay, looks good :partying_face:  and I'd like to get this in.

Two high level comments:

- This will default to dynamic/static loading which will most likely break all the CIs with a linker error. I assume most don't actually have a Vulkan environment. So I think I'd prefer either runtime loading by default or no default features at all.  With "no default features " at least they would get a build error.

- Does this PR solve https://github.com/MaikKlein/ash/issues/462? How would one build ash statically wit this PR? What happens if you have both the .so and .a in the path? IIRC you can specify the link type somewhere as a flag.  Just asking, this PR doesn't have to solve this.

MarijnS95:
> Sorry for the delay, looks good  and I'd like to get this in.
> 
> Two high level comments:
> 
> * This will default to dynamic/static loading which will most likely break all the CIs with a linker error. I assume most don't actually have a Vulkan environment. So I think I'd prefer either runtime loading by default or no default features at all.  With "no default features " at least they would get a build error.

According to https://github.com/MaikKlein/ash/pull/457#issuecomment-890278798 this needs `sudo apt-get install libvulkan-dev` as seen in `ci.yaml`.

> * Does this PR solve [Support for a statically linked vulkan library #462](https://github.com/MaikKlein/ash/issues/462)? How would one build ash statically wit this PR? What happens if you have both the .so and .a in the path?

According to https://github.com/MaikKlein/ash/pull/457/files#r689123563 the .so is preferred over the .a. I marked #462 as linked to this PR since the asker only seems to have a `.a` (if they're cross-compiling and it finds the hosts' `.so` that's a different issue), probably something to find out over time?

> IIRC you can specify the link type somewhere as a flag.  Just asking, this PR doesn't have to solve this.

There's the optional `[kind=]`, we can add `cargo:rustc-link-lib=static=vulkan` (and `dylib=` just in case) with some feature flag if this turns out to be a problem.

MaikKlein:
> According to #457 (comment) this needs sudo apt-get install libvulkan-dev as seen in ci.yaml.

Yeah that is what I meant. It will break all the CIs without a vulkan environment and they have to add `sudo apt-get install libvulkan-dev` right?

Ralith:
Yes, the build environment will need to have a libvulkan for the target available. We should probably consider this a breaking change for that reason, but IMO it's nonetheless a reasonable default. For one thing, the linker error you get for a missing foreign library is highly recognizable and easy to google, with the same solution across all programming languages, whereas a build error due to not having default features would be more idiosyncratic.

Leaving the default unchanged wouldn't be a disaster, but there's something to be said for defaulting to the lighter-weight, more robust, and more conventional configuration.

> How would one build ash statically wit this PR?

As @MarijnS95 says, it'll use whatever's available in the search path. If the target environment only has a static library, then that's what'll be used. I expect this will be sufficient in practice; especially when cross-compiling, the user has considerable control over the available libraries.

Jasper-Bekkers:
> Yes, the build environment will need to have a libvulkan for the target available. IMO it's a reasonable default. For one thing, the linker error you get for a missing foreign library is highly recognizable and easy to google, with the same solution across all programming languages, whereas a build error due to not having default features would be more idiosyncratic.

I'll resurrect this thread now that ash 0.34 is out and we're starting to see the first linker errors across our crates. I think this may have been a mistake. Rust users aren't used to fixing (or seeing) linker errors, and I honestly think this is a bad precedent to set from a usability point of view.

Ralith:
FWIW, there is overwhelming precedent for this behavior for FFI crates (e.g. cpal, openssl, openexr, gtk, ...). While I'm sure ash 0.34 will be some peoples' first encounter with link-time dependencies, most ash users should find this familiar and easily diagnosed.

aloucks:
I agree with @Jasper-Bekkers, this should not be the default. This imposes a burden that is 100% avoidable.

Jasper-Bekkers:
> I agree with @Jasper-Bekkers. This imposes a burden that is 100% avoidable. This is not the same as a sys crate where we are building the `C` source. We're linking to an external library. I think it's fine to support compile time linking but it should not be the default. 

Along with this, it's quite unclear what the need for this change was to begin with, dropping a small and extremely common dependency is not really a gain. Neither in build times (only if you look at them in isolation, not when you look at them in the context of a much larger app) nor in runtime performance.

At best it's a dubious default to have, at worst it actively harms usability for this crate. I would suggest doing a fox release that reverts this behavior honestly.

PR #463, opened 2021-08-05
Update Vulkan-Headers to 1.2.191; drop Video enum-variant edgecases

MarijnS95:
Not creating a patch release for this because the changes are limited to the provisional video extension, and breaking changes at that. This should normally require a minor release as we haven't even been able to hide these behind a feature toggle yet :/

If the next updates/PRs are interesting, should we go and push a patch release anyway at risk of breaking someones experimental workflow for a bit?


MarijnS95:
The last commit is now replicating #352, without refactoring type handling in the generator. This is necessary because `vk.xml` in 1.2.188 changed once again how `pVersionData` is represented; that's the only field across the entire API with a pointer to a static-sized array (instead of a dynamic-sized slice with a `latexmath:` expression based on other variables).

MaikKlein:
> If the next updates/PRs are interesting, should we go and push a patch release anyway at risk of breaking someones experimental workflow for a bit?

I think that is fine. Can't expect stability guarantees for provisional extensions. Of course would be better to have it behind a toggle at one point.

MaikKlein (approved):

MarijnS95:
> Of course would be better to have it behind a toggle at one point.

Yup, maybe need to have another crack at that some time soon; it's a known flaw in our API for now.

PR #465, opened 2021-08-07
ash: Repeatedly call enumeration functions when `VK_INCOMPLETE` is returned

### LINKED ISSUE #464: VK_INCOMPLETE returned by enumerate_instance_extension_properties
Downstream - https://github.com/gfx-rs/wgpu/issues/1768
Ash code - https://github.com/MaikKlein/ash/blob/d6e049f340e0b0ed8a17ff5de55a8500bbe0b534/ash/src/instance.rs#L405
Looks like the spec isn't guaranteeing that the `count` is enough:

> Because the list of available layers may change externally between calls to `vkEnumerateInstanceExtensionProperties`, two calls may retrieve different results if a `pLayerName` is available in one call but not in another. The extensions supported by a layer may also change between two calls, e.g. if the layer implementation is replaced by a different version between those calls.

Not sure who'd be responsible to rule that out. Would it be completely nuts to do this query in a loop until it succeeds?

MarijnS95:
Fixes #464

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -1638,24 +1638,14 @@ impl Device {
         &self,
         pipeline_cache: vk::PipelineCache,
     ) -> VkResult<Vec<u8>> {
-        let mut data_size: usize = 0;
-        self.device_fn_1_0
-            .get_pipeline_cache_data(
+        read_into_uninitialized_vector(|count, data| {
+            self.device_fn_1_0.get_pipeline_cache_data(
                 self.handle(),
                 pipeline_cache,
-                &mut data_size,
-                ptr::null_mut(),
+                count,
+                data as _,
             )
-            .result()?;
-        let mut data: Vec<u8> = Vec::with_capacity(data_size);
-        let err_code = self.device_fn_1_0.get_pipeline_cache_data(
-            self.handle(),
-            pipeline_cache,
-            &mut data_size,
-            data.as_mut_ptr() as _,
-        );
-        data.set_len(data_size);
-        err_code.result_with_success(data)
+        })
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkMergePipelineCaches.html>"]
@@ -2107,21 +2097,16 @@ impl Device {
         &self,
         image: vk::Image,
     ) -> Vec<vk::SparseImageMemoryRequirements> {
-        let mut count = 0;
-        self.device_fn_1_0.get_image_sparse_memory_requirements(
-            self.handle(),
-            image,
-            &mut count,
-            ptr::null_mut(),
-        );
-        let mut data = Vec::with_capacity(count as usize);
-        self.device_fn_1_0.get_image_sparse_memory_requirements(
-            self.handle(),
-            image,
-            &mut count,
-            data.as_mut_ptr(),
-        );
-        data.set_len(count as usize);
-        data
+        read_into_uninitialized_vector(|count, data| {
+            self.device_fn_1_0.get_image_sparse_memory_requirements(
+                self.handle(),
+                image,
+                count,
+                data,
+            );
+            vk::Result::SUCCESS
+        })
+        // The closure always returns SUCCESS
+        .unwrap()
     }
 }
--- a/ash/src/extensions/ext/tooling_info.rs
+++ b/ash/src/extensions/ext/tooling_info.rs
@@ -3,7 +3,6 @@ use crate::vk;
 use crate::{EntryCustom, Instance};
 use std::ffi::CStr;
 use std::mem;
-use std::ptr;
 
 #[derive(Clone)]
 pub struct ToolingInfo {
@@ -31,16 +30,10 @@ impl ToolingInfo {
         &self,
         physical_device: vk::PhysicalDevice,
     ) -> VkResult<Vec<vk::PhysicalDeviceToolPropertiesEXT>> {
-        let mut count = 0;
-        self.tooling_info_fn
-            .get_physical_device_tool_properties_ext(physical_device, &mut count, ptr::null_mut())
-            .result()?;
-        let mut v = Vec::with_capacity(count as usize);
-        let err_code = self
-            .tooling_info_fn
-            .get_physical_device_tool_properties_ext(physical_device, &mut count, v.as_mut_ptr());
-        v.set_len(count as usize);
-        err_code.result_with_success(v)
+        read_into_uninitialized_vector(|count, data| {
+            self.tooling_info_fn
+                .get_physical_device_tool_properties_ext(physical_device, count, data)
+        })
     }
 
     pub fn fp(&self) -> &vk::ExtToolingInfoFn {
--- a/ash/src/extensions/ext/tooling_info.rs
+++ b/ash/src/extensions/ext/tooling_info.rs
@@ -30,7 +30,7 @@ impl ToolingInfo {
         &self,
         physical_device: vk::PhysicalDevice,
     ) -> VkResult<Vec<vk::PhysicalDeviceToolPropertiesEXT>> {
-        read_into_uninitialized_vector(|count, data| {
+        read_into_defaulted_vector(|count, data| {
             self.tooling_info_fn
                 .get_physical_device_tool_properties_ext(physical_device, count, data)
         })
--- a/ash/src/extensions/khr/pipeline_executable_properties.rs
+++ b/ash/src/extensions/khr/pipeline_executable_properties.rs
@@ -3,7 +3,6 @@ use crate::vk;
 use crate::{EntryCustom, Instance};
 use std::ffi::CStr;
 use std::mem;
-use std::ptr;
 
 #[derive(Clone)]
 pub struct PipelineExecutableProperties {
@@ -33,27 +32,15 @@ impl PipelineExecutableProperties {
         device: vk::Device,
         executable_info: &vk::PipelineExecutableInfoKHR,
     ) -> VkResult<Vec<vk::PipelineExecutableInternalRepresentationKHR>> {
-        let mut count = 0;
-        let err_code = self
-            .pipeline_executable_properties_fn
-            .get_pipeline_executable_internal_representations_khr(
-                device,
-                executable_info,
-                &mut count,
-                ptr::null_mut(),
-            );
-        if err_code != vk::Result::SUCCESS {
-            return Err(err_code);
-        }
-        let mut v: Vec<_> = vec![Default::default(); count as usize];
-        self.pipeline_executable_properties_fn
-            .get_pipeline_executable_internal_representations_khr(
-                device,
-                executable_info,
-                &mut count,
-                v.as_mut_ptr(),
-            )
-            .result_with_success(v)
+        read_into_defaulted_vector(|count, data| {
+            self.pipeline_executable_properties_fn
+                .get_pipeline_executable_internal_representations_khr(
+                    device,
+                    executable_info,
+                    count,
+                    data,
+                )
+        })
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPipelineExecutablePropertiesKHR.html>"]
@@ -62,27 +49,10 @@ impl PipelineExecutableProperties {
         device: vk::Device,
         pipeline_info: &vk::PipelineInfoKHR,
     ) -> VkResult<Vec<vk::PipelineExecutablePropertiesKHR>> {
-        let mut count = 0;
-        let err_code = self
-            .pipeline_executable_properties_fn
-            .get_pipeline_executable_properties_khr(
-                device,
-                pipeline_info,
-                &mut count,
-                ptr::null_mut(),
-            );
-        if err_code != vk::Result::SUCCESS {
-            return Err(err_code);
-        }
-        let mut v: Vec<_> = vec![Default::default(); count as usize];
-        self.pipeline_executable_properties_fn
-            .get_pipeline_executable_properties_khr(
-                device,
-                pipeline_info,
-                &mut count,
-                v.as_mut_ptr(),
-            )
-            .result_with_success(v)
+        read_into_defaulted_vector(|count, data| {
+            self.pipeline_executable_properties_fn
+                .get_pipeline_executable_properties_khr(device, pipeline_info, count, data)
+        })
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPipelineExecutableStatisticsKHR.html>"]
@@ -91,27 +61,10 @@ impl PipelineExecutableProperties {
         device: vk::Device,
         executable_info: &vk::PipelineExecutableInfoKHR,
     ) -> VkResult<Vec<vk::PipelineExecutableStatisticKHR>> {
-        let mut count = 0;
-        let err_code = self
-            .pipeline_executable_properties_fn
-            .get_pipeline_executable_statistics_khr(
-                device,
-                executable_info,
-                &mut count,
-                ptr::null_mut(),
-            );
-        if err_code != vk::Result::SUCCESS {
-            return Err(err_code);
-        }
-        let mut v: Vec<_> = vec![Default::default(); count as usize];
-        self.pipeline_executable_properties_fn
-            .get_pipeline_executable_statistics_khr(
-                device,
-                executable_info,
-                &mut count,
-                v.as_mut_ptr(),
-            )
-            .result_with_success(v)
+        read_into_defaulted_vector(|count, data| {
+            self.pipeline_executable_properties_fn
+                .get_pipeline_executable_statistics_khr(device, executable_info, count, data)
+        })
     }
 
     pub fn fp(&self) -> &vk::KhrPipelineExecutablePropertiesFn {
--- a/ash/src/prelude.rs
+++ b/ash/src/prelude.rs
@@ -1,3 +1,5 @@
+use std::convert::TryInto;
+
 use crate::vk;
 pub type VkResult<T> = Result<T, vk::Result>;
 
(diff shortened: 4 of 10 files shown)


ash/src/device.rs
```diff
@@ -1638,24 +1638,18 @@ impl Device {
...
-            pipeline_cache,
-            &mut data_size,
-            data.as_mut_ptr() as _,
-        );
-        data.set_len(data_size);
-        err_code.result_with_success(data)
+                data as _,
+            );
+            // TODO: Truncation might be dangerous!
+            *count = data_size as u32;
```
MarijnS95, line 1650: I don't think we can make `read_into_vector` generic with just `Into` and `From`, perhaps this is the time to introduce https://docs.rs/num/0.4.0/num/trait.Zero.html again?

ash/src/device.rs
```diff
@@ -2107,21 +2101,16 @@ impl Device {
...
+            self.device_fn_1_0.get_image_sparse_memory_requirements(
+                self.handle(),
+                image,
+                count,
+                data,
+            );
+            vk::Result::SUCCESS
+        })
+        // The closure always returns SUCCESS
+        .unwrap()
```
MarijnS95, line 2114: This helper function is not ideal for Vulkan functions without error codes, but it's does save quite a few lines.

MarijnS95:
I think this PR can go in with a patch release as no API is modified. However, IMO we should perform a minor (breaking) release afterwards to get rid of explicit `*_len()` functions and make them use the same helper instead:

```rust
get_image_sparse_memory_requirements2_len()
get_physical_device_queue_family_properties2_len()
get_physical_device_sparse_image_format_properties2_len()
enumerate_physical_device_groups_len()
```

The problem here is that their non-`_len` counterpart receives a mutable slice but never updates the resulting size (impossible with slice references anyway, unless having a mutable reference to the mutable slice reference), in case `count` changed between the `_len` and "fill" call.

ash/src/prelude.rs
```diff
@@ -19,3 +19,27 @@ impl vk::Result {
...
+
+/// Repeatedly calls `f` until it does not return [`vk::Result::INCOMPLETE`] anymore,
+/// ensuring all available data has been read into the vector.
+///
+/// See for example [`vkEnumerateInstanceExtensionProperties`]: the number of available
+/// items may change between calls; if it has increased `INCOMPLETE` is returned because
+/// the result-vec is not large enough.
+///
+/// [`vkEnumerateInstanceExtensionProperties]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumerateInstanceExtensionProperties.html
+pub(crate) unsafe fn read_into_vector<T>(
```
MarijnS95, line 31: Naming suggestions are welcome, btw :wink:

ash/src/extensions/khr/pipeline_executable_properties.rs
```diff
@@ -33,27 +32,15 @@ impl PipelineExecutableProperties {
...
-            .get_pipeline_executable_internal_representations_khr(
-                device,
-                executable_info,
-                &mut count,
-                ptr::null_mut(),
-            );
-        if err_code != vk::Result::SUCCESS {
-            return Err(err_code);
-        }
-        let mut v: Vec<_> = vec![Default::default(); count as usize];
```
MarijnS95, line 48: Note that the `sType` in these structs is not initialized anymore which is probably the task of `get_pipeline_executable_internal_representations_khr` (etc. below). This is likely fine to assume but I want to prevent a situation like #394.

ash/src/prelude.rs
```diff
@@ -19,3 +19,27 @@ impl vk::Result {
...
+    f: impl Fn(&mut u32, *mut T) -> vk::Result,
+) -> VkResult<Vec<T>> {
+    loop {
+        let mut count: u32 = 0;
+        f(&mut count, std::ptr::null_mut()).result()?;
+        let mut data = Vec::with_capacity(count as usize);
```
MarijnS95, line 37: Do we want to keep this vector around in subsequent loop iterations and use `.reserve()` instead, likely using just a `realloc` instead of `free` + `malloc`? OTOH we don't need the original memory to be copied, that's probably making it less efficient in the end.

ash/src/prelude.rs
```diff
@@ -19,3 +19,27 @@ impl vk::Result {
...
+
+        let err_code = f(&mut count, data.as_mut_ptr());
+        if err_code != vk::Result::INCOMPLETE {
+            data.set_len(count as usize);
+            break err_code.result_with_success(data);
```
Ralith, line 42: Style nit: no need to be cheeky here
```suggestion
            return err_code.result_with_success(data);
```

ash/src/device.rs
Ralith: How about `N: Default, usize: TryFrom<N>`?

ash/src/device.rs
Ralith: This seems fine to me. It should optimize out trivially.

ash/src/extensions/khr/pipeline_executable_properties.rs
Ralith: > which is probably the task of get_pipeline_executable_internal_representations_khr (etc. below).

Unfortunately, this isn't true. We *must* initialize both `sType` and, even more importantly, `pNext` correctly in every case. The implementation may traverse `pNext` in search of extension output structures. I recommend introducing a variant of `read_into_vector` which takes an initial value, or perhaps which invokes `T::default()` internally.

ash/src/prelude.rs
Ralith: Nothing will be copied since the length is zero, so this is fine to do, though I don't think it'll matter in practice.

Ralith (changes requested):
> However, IMO we should perform a minor (breaking) release afterwards to get rid of explicit *_len() functions and make them use the same helper instead

These variants were introduced for the case where an extension output structure may be needed, in which case each element's `pNext` must be initialized to address a unique output structure. It'd be nice to develop a high-level, abstract interface for that case, but the helper introduced here isn't sufficient; it'll have to be significantly more complex (maybe a HList of vecs?).

ash/src/prelude.rs
MarijnS95: `loop`-`break` is a common Rust pattern I'd like to embrace. If we shouldn't use `break <some_value>;` I don't think we should be using `loop` either and just convert this into `let mut foo = INCOMPLETE; while foo == INCOMPLETE` etc.

ash/src/device.rs
MarijnS95: Just seems ugly to me - but the original duplicated call more so and is more error-prone at the same time.

Will keep it this way, shall I duplicate this comment to every (4 or so, iirc) occurrence of this pattern or leave it out entirely?

ash/src/device.rs
MarijnS95: That's really neat, didn't think of that :). Switched to `N: TryInto<usize>` since it makes the trait bound a bit easier and I'd anyway write `count.try_into()` instead of `usize::try_from(count)`. Thanks!

ash/src/extensions/khr/pipeline_executable_properties.rs
MarijnS95: That's horrible, but true for _every_ function returning its result in a "base" structure (with `sType`/`pNext` fields). Is there anything we can automate with traits here (or at least safeguard), maybe something with a `BaseOutStructure`-like trait the generator could emit?

MarijnS95: We could have an automated (private) trait telling how to initialize the given datastructure were it used in an output position: uninitialized for handles and the like, `Default::default()` for `BaseOutStructure`-like types.

Alternatively we can take the performance "hit" of always `default`-initializing every vector instead of using `with_capacity()` and `unsafe fn set_len()`.

Finally we can have two distinct functions for this, but it's up to authors and reviewers to ensure the right variant is used. If so, I'd make the more-unsafe/unreliable uninitialized variant obvious in the name (`read_into_uninitialized_vector`) instead of the saner/safer variant (`read_into_defaulted_vector`).

ash/src/prelude.rs
MarijnS95: The `Vec<>` len will be zero, but the length of the allocation will not - `realloc` looks at the previous size passed to `malloc` which is based on what was passed to `with_capacity`, it doesn't know about `Vec<>` having a counter determining how much is actually initialized with useful data (leaving detailed optimizations of ie. pages being physically allocated only when they're actually going to be written to aside).

This is just a moot point anyways, let's resolve and not care about it.

MarijnS95:
> > However, IMO we should perform a minor (breaking) release afterwards to get rid of explicit *_len() functions and make them use the same helper instead
> 
> These variants were introduced for the case where an extension output structure may be needed, in which case each element's `pNext` must be initialized to address a unique output structure.

Conveniently enough only `enumerate_physical_device_groups_len` out of the four listed functions actually has a return type, there's no possibility for the others to return `INCOMPLETE`.

We can think about helpers for initializing `pNext` later; I think you'll want to apply that to the functions in `pipeline_executable_properties.rs` later as well, which need the same initialization but currently do not allow `pNext` to be filled beforehand?

ash/src/prelude.rs
Ralith: No strong opinion.

ash/src/extensions/khr/pipeline_executable_properties.rs
Ralith: In openxrs, I generate associated functions of this form:
```rust
    pub fn out(next: *mut BaseOutStructure) -> MaybeUninit<Self> {
        let mut x = MaybeUninit::<Self>::uninit();
        unsafe {
            (x.as_mut_ptr() as *mut BaseOutStructure).write(BaseOutStructure {
                ty: Self::TYPE,
                next,
            });
        }
        x
    }
```
To avoid adding large amounts of code, this could be a single `unsafe` function that assumes a suitable `T`. Alternatively, we could add an empty `unsafe trait OutStructure` and make the function itself safe. I'm not sure if this is really worth the additional complexity over just using `Default` (even C++ Vulkan developers tend to 0-initialize structures wholesale as a relatively sane default) but it should work fine. However, perhaps it would be best to leave that to a followup PR?

Ralith:
> We can think about helpers for initializing pNext later; I think you'll want to apply that to the functions in pipeline_executable_properties.rs later as well, which need the same initialization but currently do not allow pNext to be filled beforehand?

Yeah, I think there's a whole bunch of cases where we haven't exposed `pNext` for output just because nobody's wanted it yet. Agreed that we should leave it to followup work, let's just not take away the existing workarounds until then.

ash/src/extensions/khr/pipeline_executable_properties.rs
MarijnS95: @Ralith I suppose we could generate this function for exactly every structure that receives `.push_next` (unless there are structs used in this out-position with no known extension yet, while obviously still requiring appropriate initialization), and I have some ideas to move all that duplication to a single definition with traits. We can use the same trait bound for `fn push_next()` in `fn out()` as well.

We will have the `MaybeUninit<Self>`'s in vecs/slices to pass them at-large to Vulkan though, how are you converting them back to their initialized form in openxrs? `transmute`?

Either way, let's leave this for a followup PR. The initial default-initialization for `pipeline_executable_properties.rs` has been restored. I went over all the calls to `read_into_uninitialized_vector` and found only `get_physical_device_tool_properties` which was (even before this PR) not performing the appropriate initialization.

PS. I like the idea of keeping everything but `s_type`/`p_next` uninitialized!

ash/src/device.rs
Ralith: No strong preference; I think it's obvious enough from context that the comment isn't absolutely necessary, but it saves a double-take.

Ralith (approved):

kvark (commented):
Is this really needed for all these functions? I'd think that the scenario when VK_INCOMPLETE shows up is very specific, since something must be able to change behind ones back. It doesn't mean that all of the vectored queries have to do it.

MarijnS95:
@kvark The documentation for _every_ function that now uses `read_into_*_vector` mentions that it can possibly return `VK_INCOMPLETE`, even though most are likely only because of not providing enough space in the first place instead of being a change in the background.

Either way I don't want to be inconsistent in our API and only handle `VK_INCOMPLETE` in specific places, especially if Ash is the one querying for and allocating the vector. Unless the Vulkan docs are very clear on where a change behind ones back can possibly happen, I think it's better to be safe than sorry. An extra compare+jump shouldn't hurt that much either, in favour of having simplified code too.

In any case, if we really want to make a distinction between those kinds of functions (again, if the Vulkan API clearly specifies this), the non-looping queries would get an assertion/panic instead as they're not allowed to change the count after initially querying it.

kvark:
Specification says `VK_INCOMPLETE` is possible for a simple reason that your count is too low. I don't think we should be careful "just in case" here. Ash is expected to understand the Vulkan spec in its implementation.

Ralith:
> Specification says VK_INCOMPLETE is possible for a simple reason that your count is too low

Yes, that's the definition of the error. What it *doesn't* say is that the required capacity is guaranteed never to change. We should not rely on assumptions that implementations are under no obligation to comply with, particularly when there's clear precedent for getters that may change at any time.

filnet:
The exact phrasing in the [specs](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkResult.html) is:

> VK_INCOMPLETE A return array was too small for the result

But it does not say when or how this array can be too small.

MarijnS95:
> Specification says `VK_INCOMPLETE` is possible for a simple reason that your count is too low.

This is extremely unlikely (dare I say impossible bar cosmic radiation) when ash just queried the count in the previous call. You yourself suggested to run this in a loop, and since [`vkEnumerateDeviceExtensionProperties`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumerateDeviceExtensionProperties.html) does not explicitly state whether or not a change at runtime is possible, we assume this is the case for every function unless proven otherwise.

If you want to get to the bottom of this I recommend checking if _anything_ on the system of the original reporter is out of the ordinary, and otherwise insert some debug logging to see if the count indeed changes (and if possible, run this in some tight loop that could perhaps diff the list before and after). This is also my first time encountering `VK_INCOMPLETE`.

I'd still like to keep the `read_into_*_vector` with lambda if we decide against this `loop`, it cleans up duplicate count/result handling nicely.

kvark:
ok, I don't think my point got through, but I also don't want to block this fix that would clearly help us. So please land this :)

MarijnS95:
@kvark It's important to get on the same page first as this fix is solving an issue reported by you - I have never seen it before and don't rely on this PR.

If you think there is a specific environment in which a change in `count` occurs, shouldn't it be up to the application to detect and handle this case? At least I think that's what [you're suggesting above](https://github.com/MaikKlein/ash/pull/465#issuecomment-903931972)? Otherwise, don't we all agree that the spec is rather vague in this aspect, and doesn't confirm nor rule out a change of `count` between queries?

kvark:
The problem originates from this clause:
> Because the list of available layers may change externally between calls to vkEnumerateInstanceExtensionProperties, two calls may retrieve different results if a pLayerName is available in one call but not in another. The extensions supported by a layer may also change between two calls, e.g. if the layer implementation is replaced by a different version between those calls.

This is specifically about `vkEnumerateInstanceExtensionProperties`. If it applied to other functions, we'd expect to see similar statements there. So the spec doesn't appear to be vague here.

MarijnS95:
@kvark Now that brings something new to the table. Up to this point it looks like everyone failed to notice that bit in the spec but you.

Are you aware of any other function that has the same or a similar mention in the spec? 

I'll update this PR ASAP to deal with this.

kvark:
It's not new, it's the starting point - https://github.com/MaikKlein/ash/issues/464#issue-962066971
Searching the spec reveals that this is the only function with that kind of obscure behavior. We can cover more later if we discover anything - the changes are backwards compatible anyway.

filnet:
VK_INCOMPLETE can happen :

1/ if the user provides an array that is too small.
But that can't happen with the proposed change.

2/ when using a Vulkan function that explicitly mentions that the list can change between two calls.
Only vkEnumerateInstanceExtensionProperties mentions but I find it a bit dangerous to assume that no other functions has this behavior. And Vulkan might add more such functions or change the behavior of existing functions.

3/ with bad user code ? 
Would it be possible for user code to issue a concurrent function call in between the "two calls" which would cause the list size to change ? I doubt that this is possible since Vulkan is mostly stateless but who knows.

MarijnS95:
> It's not new, it's the starting point - [#464 (comment)](https://github.com/MaikKlein/ash/issues/464#issue-962066971)

New in the sense that I hadn't noticed / forgotten this was mentioned in the issue.

> Searching the spec reveals that this is the only function with that kind of obscure behavior. We can cover more later if we discover anything - the changes are backwards compatible anyway.

Great, only one function to deal with then.

> VK_INCOMPLETE can happen :
> 
> 1/ if the user provides an array that is too small.
> But that can't happen with the proposed change.

Couldn't happen before either, because Ash was querying the size and directly reusing this value in every instance.

> 2/ when using a Vulkan function that explicitly mentions that the list can change between two calls.
> Only vkEnumerateInstanceExtensionProperties mentions but I find it a bit dangerous to assume that no other functions has this behavior. And Vulkan might add more such functions or change the behavior of existing functions.

We'll do this just for `vkEnumerateInstanceExtensionProperties` now. No-one has ever seen this before anyway, and it's no surprise that the only occurrence of a changing value happens on a function that has it clearly documented. Nothing dangerous IMO.

> 3/ with bad user code ?
> Would it be possible for user code to issue a concurrent function call in between the "two calls" which would cause the list size to change ? I doubt that this is possible since Vulkan is mostly stateless but who knows.

Strongly doubt this. Maybe available layers can change at any point since that's an external factor?

filnet:
What about functions related to hardware ?
`vkGetPhysicalDeviceDisplayPropertiesKHR` for example. What if a display device gets connected/disconnected between the two calls.

MarijnS95:
> What about functions related to hardware ?
> `vkGetPhysicalDeviceDisplayPropertiesKHR` for example. What if a display device gets connected/disconnected between the two calls.

Same thoughts here; quick testing shows this doesn't change until restarting the application. Maybe that is the case for similar functions, too.

Ralith:
> We'll do this just for `vkEnumerateInstanceExtensionProperties` now.

We should do this for every function that may return `VK_INCOMPLETE` unless the spec explicitly guarantees that the size is fixed. "It worked on my machine" is not an excuse for UB.

MarijnS95:
> We should do this for every function that may return `VK_INCOMPLETE` unless the spec explicitly guarantees that the size is fixed. "It worked on my machine" is not an excuse for UB. It's not like there's a drawback to being defensive here.

Rather, there's two camps now: those who want to spend one extra cycle for the compare, and those who don't. I'm all for the defensive route, feel free to merge.

kvark:
I don't care about extra cycles for this. I care about Ash knowing what API it abstracts over. So far, Vulkan spec does not explicitly spell out any guarantees about what the count is between different invocations, except for `vkEnumerateInstanceExtensionProperties`. If it's also the case for other functions (like `vkGetPhysicalDeviceDisplayPropertiesKHR`), then I'd say this would be a spec bug. But Ash doesn't need to be smarter than the spec.

Also, I'm no longer sure that the loop is the right solution here. What's the point of ensuring that you got *exactly* all of the elements *at some point* of the program execution? If I got 3 layers reported, that doesn't mean that the next second the configuration isn't going to change, anyway.

Did we consider just doing the following:
  1. do the regular call for getting the count, and then the call to fill the data
  2. if the returned count is less, issue a `log::warn` and continue
  3. if the returned error is VK_INCOMPLETE, issue `log::warn` and continue

This seems marginally better than having a loop. And in a scenario where it happens - your configuration changed between two invocations - there isn't any guarantee that it's not going to change right after the call, anyway.

---

This is all super bike-sheddy, and I understand if you folks just don't consider it worth improving. Landing the patch as is would be fine by me.

filnet:
There's also `vkEnumerateInstanceLayerProperties` ;)

ash/src/extensions/ext/tooling_info.rs
```diff
@@ -30,7 +30,7 @@ impl ToolingInfo {
         &self,
         physical_device: vk::PhysicalDevice,
     ) -> VkResult<Vec<vk::PhysicalDeviceToolPropertiesEXT>> {
-        read_into_uninitialized_vector(|count, data| {
+        read_into_defaulted_vector(|count, data| {
```
filnet, line 33: commit message seems off.

Ralith:
> So far, Vulkan spec does not explicitly spell out any guarantees about what the count is between different invocations, except for vkEnumerateInstanceExtensionProperties. If it's also the case for other functions (like vkGetPhysicalDeviceDisplayPropertiesKHR), then I'd say this would be a spec bug.

If the spec guarantees a thing in one case and doesn't guarantee that thing in other case, that's not a bug, it's the absence of a guarantee. When behavior isn't guaranteed, we cannot assume a single possibility.

> What's the point of ensuring that you got exactly all of the elements at some point of the program execution? If I got 3 layers reported, that doesn't mean that the next second the configuration isn't going to change, anyway.

Good question. Unfortunately, it is necessary, because there's no guarantee *which* results are omitted from an incomplete set. A user can reasonably be expected to tolerate having fractionally out of date information regarding a change in configuration/whatever that just happened, but they should not expect to lose information that hasn't *ever* changed just because another datum in the same category changed.

> `log::warn` and continue

IMO adding `log` to the dependencies and public interface of the crate is strongly contrary to the spirit of ash; user-facing diagnostics are not the responsibility of a Vulkan binding.

ash/src/extensions/ext/tooling_info.rs
Ralith: Seems correct to me; solving this problem is exactly why `read_into_defaulted_vector` exists.

Ralith:
Merging as-is since this is a clear improvement over the status quo, and a code size reduction besides. We can discuss in a new issue if anyone feels further refinement is necessary.

PR #469, opened 2021-08-22
Expose LIB_PATH

zarik5:
Objective: Construct a `EntryCustom<Option<Arc<Library>>>` used to store an either locally or externally managed entry point.

`LIB_PATH` constants are used internally by `Entry::new()`. This PR exposes these constants to be used conditionally with `EntryCustom::new_custom()`, without having to redefine them.


zarik5:
This PR might be superseded by #457 which removes the type parameters from `EntryCustom`/`Entry` and makes the `lib` field itself `Option<>`.

PR #476, opened 2021-09-25
Add `Packed24_8` helper-type for constructing AS Instance bitfields

### LINKED ISSUE #474: `AccelerationStructureInstanceKHR` is missing a builder
[`AccelerationStructureInstanceKHR`](https://docs.rs/ash/0.33.3+1.2.191/ash/vk/struct.AccelerationStructureInstanceKHR.html) is a nasty one to use without a builder as several fields are packed into one:

```c
// Provided by VK_KHR_acceleration_structure
typedef struct VkAccelerationStructureInstanceKHR {
    VkTransformMatrixKHR          transform;
    uint32_t                      instanceCustomIndex:24;
    uint32_t                      mask:8;
    uint32_t                      instanceShaderBindingTableRecordOffset:24;
    VkGeometryInstanceFlagsKHR    flags:8;
    uint64_t                      accelerationStructureReference;
} VkAccelerationStructure
```

See: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureInstanceKHR.html

If there's any way that I can help to fix this issue, please let me know :^)

MarijnS95:
Fixes #474, cc @expenses

Draft because we can possibly improve API design here by making a special type instead of `u32` that has this as a constructor? This doesn't include a full builder yet, but if we do anything more fancy than this we should move these types to a new file directly included in `ash/vk/` instead of having static text (that doesn't depend on anything from the autogeneration process) hardcoded in the generator.


generator/src/lib.rs
```diff
@@ -2128,10 +2128,18 @@ pub fn generate_struct(
...
             pub struct AccelerationStructureInstanceKHR {
                 pub transform: TransformMatrixKHR,
+                /// Use [`pack_24_8(instance_custom_index, mask)`][Self::pack_24_8()] to construct this field
                 pub instance_custom_index_and_mask: u32,
+                /// Use [`pack_24_8(instance_shader_binding_table_record_offset, flags)`][Self::pack_24_8()] to construct this field
                 pub instance_shader_binding_table_record_offset_and_flags: u32,
                 pub acceleration_structure_reference: AccelerationStructureReferenceKHR,
             }
+            impl AccelerationStructureInstanceKHR {
+                pub fn pack_24_8(low_24: u32, high_8: u8) -> u32 {
```
expenses, line 2138: We might want a comment here like 'Pack `low_24` and `high_8` into the same `u32` such that `high_8` occupies the highest 8 bits and `low_24` occupies the lowest 24 bits' or something.

Otherwise, this looks great!

Rua:
I think using a custom type is more idiomatic, and conveys to the user that it's not just a single integer. If the inner `u32` is kept private, it can also prevent incorrect usage.

MarijnS95:
@Rua Yup, got it done :)

Regarding making `vk.rs` a manual file and adding `prelude.rs`, what's everyones view on adopting the `auto` folder/module from GTK where the generator spits its outputs (`src/vk/auto/`), and everything outside that (straight in the root of `src/vk/`) is manual files? Should make it a bit easier to understand which of the files are safe to touch by hand, and which files are left to the generator.

ash/src/vk/prelude.rs
```diff
@@ -0,0 +1,27 @@
+/// Holds 24 bits in the least significant bits of memory,
+/// and 8 bytes in the most significant bits of that memory,
+/// occupying a single [`u32`] in total. This is commonly used in
+/// [acceleration structure instances] such as
+/// [`vk::AccelerationStructureInstanceKHR`][super::AccelerationStructureInstanceKHR],
+/// [`vk::AccelerationStructureSRTMotionInstanceNV`][super::AccelerationStructureSRTMotionInstanceNV] and
+/// [`vk::AccelerationStructureMatrixMotionInstanceNV`][super::AccelerationStructureMatrixMotionInstanceNV].
+///
+/// [acceleration structure instances]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureInstanceKHR.html#_description
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
```
MaikKlein, line 10: Hm that should have `#[repr(transparent)]`.

MaikKlein (changes requested):

ash/src/vk/prelude.rs
MarijnS95: Yup, whoops.

expenses (approved):
Looks good to me!

MaikKlein (approved):

PR #477, opened 2021-09-28
Update Vulkan-Headers to 1.2.198

MarijnS95:


generator/src/lib.rs
```diff
@@ -1390,7 +1390,7 @@ pub fn variant_ident(enum_name: &str, variant_name: &str) -> Ident {
     // TODO: Also needs to be more robust, vendor names can be substrings from itself, id:4
     // like NVX and NV
     let vendors = [
-        "_NVX", "_KHR", "_EXT", "_NV", "_AMD", "_ANDROID", "_GOOGLE", "_INTEL",
+        "_NVX", "_KHR", "_EXT", "_NV", "_AMD", "_ANDROID", "_GOOGLE", "_INTEL", "_FUCHSIA",
```
MarijnS95, line 1393: Implementing the TODO just above (`// TODO: Should be read from vk.xml id:2`) is really ugly:

https://github.com/Traverse-Research/ash/compare/vendor-tags

Makes me want to see in what state generator2 is and continue working on that, in hopes of having a more clear `parsing -> immediate representation with context -> write out to disk` instead of all these `HashSet`s and `HashMap`s passed all over the place.

MarijnS95:
With the addition of `VK_EXT_rgba10x6_formats`, `VK_KHR_format_feature_flags2` and `VK_KHR_maintenance4` this seems like a good moment to update and release ash. After merging this PR I can submit the extension for maintenance4 so that we have the new commands covered from the get-go.

MarijnS95:
Currently blocking on https://github.com/KhronosGroup/Vulkan-Headers/issues/236.

Jasper-Bekkers (approved):

PR #478, opened 2021-10-04
vk: Remove misleading `all()`/`-`/`-=` function/ops from bitflags

### LINKED ISSUE #303: Incorrect generation of bitflags in extensions
It seems we are not correctly generating bitflags that are defined in extensions.

For example: 
```
<extension name="VK_EXT_fragment_density_map" number="219" type="device" requires="VK_KHR_get_physical_device_properties2" author="EXT" contact="Matthew Netsch @mnetsch" supported="vulkan">
    <require>
        <enum bitpos="0"  extends="VkSamplerCreateFlagBits"         name="VK_SAMPLER_CREATE_SUBSAMPLED_BIT_EXT"/>
        <enum bitpos="1"  extends="VkSamplerCreateFlagBits"         name="VK_SAMPLER_CREATE_SUBSAMPLED_COARSE_RECONSTRUCTION_BIT_EXT"/>
    </require>
```
```Rust
#[repr(transparent)]
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
#[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkSamplerCreateFlagBits.html>"]
pub struct SamplerCreateFlags(pub(crate) Flags);
vk_bitflags_wrapped!(SamplerCreateFlags, 0b0, Flags);
```

`SamplerCreateFlags::all()` should not be `0b0`

### LINKED ISSUE #396: use bitflags crate
Ash uses a [custom macro vk_bitflags_wrapped!](https://github.com/MaikKlein/ash/blob/master/ash/src/vk/macros.rs#L18), that is not specific to Vulkan.
It could instead use bitflags! macro for [generating a more complete rust code](https://docs.rs/bitflags/1.2.1/bitflags/example_generated/struct.Flags.html).
This crate is used by [many other rust projects](https://crates.io/crates/bitflags/reverse_dependencies).

MarijnS95:
Fixes #303, fixes #396

The `all()` function only represents bitflags known in the core of Vulkan; it omits all bits added by extensions making this function unrepresentative and has hence been scheduled for removal for quite some time to get rid of the confusion it causes.

Alternatively the generator could be taught to collect bitflags added by extensions, but new extensions get added over time skewing available values in ash versus the current driver/environment.  This makes the value from `all()` unreliable and fragile at best.

---

Keeping this draft so that we can all give a final sign-off on removing `fn all()`. Alternatively we can collect all possible values in the generator, but in that case I prefer to just migrate over to `bitflags!` and provide all extension constants within the same block so that it's immediately visible what the possible values are, instead of relying on "hidden" `impl` blocks in different files.

### DIFF
--- a/ash/src/vk/macros.rs
+++ b/ash/src/vk/macros.rs
@@ -1,6 +1,6 @@
 #[macro_export]
 macro_rules! vk_bitflags_wrapped {
-    ($ name : ident , $ all : expr , $ flag_type : ty) => {
+    ($ name : ident , $ flag_type : ty) => {
         impl Default for $name {
             fn default() -> Self {
                 Self(0)
@@ -10,10 +10,6 @@ macro_rules! vk_bitflags_wrapped {
             #[inline]
             pub const fn empty() -> Self {
                 Self(0)
-            }
-            #[inline]
-            pub const fn all() -> Self {
-                Self($all)
             }
             #[inline]
             pub const fn from_raw(x: $flag_type) -> Self {
@@ -26,10 +22,6 @@ macro_rules! vk_bitflags_wrapped {
             #[inline]
             pub fn is_empty(self) -> bool {
                 self == Self::empty()
-            }
-            #[inline]
-            pub fn is_all(self) -> bool {
-                self & Self::all() == Self::all()
             }
             #[inline]
             pub fn intersects(self, other: Self) -> bool {
@@ -80,24 +72,11 @@ macro_rules! vk_bitflags_wrapped {
                 *self = *self ^ rhs
             }
         }
-        impl ::std::ops::Sub for $name {
-            type Output = Self;
-            #[inline]
-            fn sub(self, rhs: Self) -> Self {
-                self & !rhs
-            }
-        }
-        impl ::std::ops::SubAssign for $name {
-            #[inline]
-            fn sub_assign(&mut self, rhs: Self) {
-                *self = *self - rhs
-            }
-        }
         impl ::std::ops::Not for $name {
             type Output = Self;
             #[inline]
             fn not(self) -> Self {
-                self ^ Self::all()
+                Self(!self.0)
             }
         }
     };
--- a/ash/src/extensions/experimental/amd.rs
+++ b/ash/src/extensions/experimental/amd.rs
@@ -34,11 +34,7 @@ use std::os::raw::*;
 #[repr(transparent)]
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct GpaSqShaderStageFlags(pub(crate) Flags);
-vk_bitflags_wrapped!(
-    GpaSqShaderStageFlags,
-    0b1111111111111111111111111111111,
-    Flags
-);
+vk_bitflags_wrapped!(GpaSqShaderStageFlags, Flags);
 impl fmt::Debug for GpaSqShaderStageFlags {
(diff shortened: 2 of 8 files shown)


ash/src/vk/macros.rs
```diff
@@ -1,103 +1,95 @@
...
             }
         }
         impl ::std::ops::Not for $name {
-            type Output = $name;
+            type Output = Self;
             #[inline]
-            fn not(self) -> $name {
-                self ^ $name::all()
+            fn not(self) -> Self {
+                Self(!self.0)
```
MarijnS95, line 92: The best we can do here is fully negate the bitflags since we never really know what "all" bitflags are - even if we decide to keep and improve `fn all()` in the end.

(Sidenote: should I create a separate PR for te `$name` -> `Self` conversion so that these details pop out better?)

Ralith (approved):

ash/src/vk/macros.rs
Ralith: > The best we can do here is fully negate the bitflags

Seems reasonable to me. I think the only case where you'd be negating bitflags is for purposes of constructing masks or something, in which case this is exactly the desired behavior.

> should I create a separate PR for te $name -> Self conversion

This would make it easier to carefully review, yeah.

MarijnS95:
@Ralith What do you think, shall we just get this in?

Ralith (approved):
Absolutely!

MarijnS95:
@Ralith We're currently discussing whether `ops::Not`, and by extension `ops::Sub` make any sense or are better removed as well... Thoughts?

MarijnS95:
Alternatively we could hide this behind a default-disabled cfg - gives people the ability to opt in, as long as we name it something like `bitflag_broken_ops_let_us_know_if_you_are_relying_on_this_or_it_will_be_removed_indefinitely`.

Ralith:
I'm not a bit-twiddling wizard, but it's difficult for me to imagine what `Sub` could possibly be useful for, so I'm totally on board there. I think `Not` is more plausible for masking reasons so I'm ambivalent there, but IMO the cost to removing it is low since people can easily and safely operate on the raw representation to address niche requirements.

MarijnS95:
I can only imagine both being useful for turning off/removing/masking out a bunch of flags that you already know about, and just seeing if the result includes anything new.

Perhaps we should just remove them so that users remain absolutely in control if they wish to use the internal bit representation, which as you say is easily accessible. If anyone relies on this and has a strong argument for either implementation we'll hear about it soon I guess.

Ralith:
SGTM.

MarijnS95:
Actually, we just found a use of this in our own code:

```rust
required_mem_flags &= !vk::MemoryPropertyFlags::DEVICE_LOCAL;
```

For a very simple allocator that tries to create device local and host visible memory for an upload, and otherwise falls back to just host visible. We'd have to change this to `= HOST_VISIBLE` which is not really pretty.

Note also that this could have been written as:

```rust
required_mem_flags -= vk::MemoryPropertyFlags::DEVICE_LOCAL;
```
Which is nasty and not how integer math works.

`Sub`/`SubAssign` should absolutely go, still not entirely convinced about `not()`.

Ralith:
That's about the sort of thing I was imagining, yeah. I vote we leave `not` in for now and just get this merged.

Jasper-Bekkers (approved):

Jasper-Bekkers:
Nice to see these oddities get cleaned up.

PR #481, opened 2021-10-14
Update Winit and Fix crash on x11 Linux

hYdos:
Fixes `thread 'main' panicked at 'attempted to leave type platform::platform::x11::util::input::PointerState uninitialized, which is invalid', /rustc/c8dfcfe046a7680554bf4eb612bad840e7631c4b/library/core/src/mem/mod.rs:663:9` on X11 Linux devices.

TODO:
- [x] Fix Triangle Example
- [x] Fix Texture Example


MarijnS95:
Keep in mind that #356 also performs this upgrade, and improves the examples at the same time.

PR #482, opened 2021-10-16
Make Debug impls optional

Ralith:
When disabled, this buys us a ~12% reduction in buildtime in my environment

I'd like some discussion on whether it should be enabled by default. It's less breaking that way, but I dunno how often people are going to scan the feature list for opportunities to improve buildtime.


Ralith:
Since nobody seems to feel strongly, I've updated this to disable the feature by default.

MarijnS95:
> I'm tracking the windows-rs repository, and users seem rather disgruntled by the removal of `Debug`. Not sure if they'd be as unhappy if there was a `debug` feature, even id disabled by default.

Ralith:
I don't have enough experience with winapi to judge how the usage patterns might vary. I'd rather have this in but enabled by default than not in at all, if that's your preference? I'll ask around and see if anyone I know would be impacted; hard to predict in general.

MarijnS95:
`windows-rs`, the replacement for `winapi` - gauging the few replies and thumbs up [here](https://github.com/microsoft/windows-rs/issues/1392), and the Rust API guidelines suggesting to implement it for public types: https://rust-lang.github.io/api-guidelines/debuggability.html#all-public-types-implement-debug-c-debug it seems better to leave it on by default? I'm still fine with having it as `feature` of course.

OTOH I'm not sure how useful our limited debugging implementation is: we don't even (and for good reason) dereference and follow pointers. Still, at least being able to see pointer values and everything else that's not a pointer already goes a long way. Still on the fence about default-enabling or disabling it...

Ralith:
I think our current `Debug` impls are trivially useful for output structures, and potentially useful for input structures that might get handled by intermediate layers of abstractions in user code. Sure, a lot of stuff needs pointer context for completeness, but lots of stuff doesn't as well; e.g. being able to quickly and easily inspect memory barriers seems convenient.

Given the uncertainty, I think it's better for us to merge this with the feature enabled by default than agonize over it, so I've turned it back on. We can always disable in the future.

MarijnS95:
@Ralith On my machine I can win another 400-500ms (7.9s down to 7.4s) by removing `const_debugs`: https://github.com/Ralith/ash/compare/optional-debug...MaikKlein:optional-debug

We may want to preserve just the debug for `Result` though?

Ralith:
I preserved `Result` (for `unwrap()`) and `ObjectType` (for debug messenger purposes).

ash/src/vk/enums.rs
```diff
@@ -2242,3 +2242,117 @@ impl QueryResultStatusKHR {
     pub const NOT_READY: Self = Self(0);
     pub const COMPLETE: Self = Self(1);
 }
+impl fmt::Debug for ObjectType {
```
MarijnS95, line 2245: If possible I would have snuck these in directly below where the type is defined - we already have an edgecase for `Result` in the generator where we can call `generate_const_debugs`? Seems ugly to pass the `const_values` map down into that though, to be able to extract a specific `Ident`.

Ralith: Really doesn't seem worth complicating the generator to me, especially with that code due to be scrapped en masse soon anyway.

MarijnS95: Fair enough :)

> soon

I've been telling this a lot to myself as well :sweat_smile:

MarijnS95 (approved):
LGTM!

(I hope nobody notices the `cfg_attr` being below instead of above the `derive`s like in `vk/definition.rs` - I did :grimacing:)

PR #486, opened 2021-10-30
ash: Regenerate negative constants with `syn 1.0.81`

MarijnS95:
`syn` has been updated to not remove the `_` again from negative numbers, which previously disappeared in the 1.0 upgrade in 43b5058 ("Upgrade proc_macro2/syn/quote to v1.0 (#351)").

See also https://github.com/dtolnay/syn/releases/tag/1.0.81:

    Support arbitrary precision negative literal tokens on rustc 1.56+

### DIFF
--- a/ash/src/vk/extensions.rs
+++ b/ash/src/vk/extensions.rs
@@ -236,11 +236,11 @@ impl KhrSurfaceFn {
 }
 #[doc = "Generated from 'VK_KHR_surface'"]
 impl Result {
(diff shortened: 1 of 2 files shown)


MarijnS95:
This one has a bit of prio to fix the unsuspecting build failure on `master` (merged PR was submitted before `1.0.81`, and everything was fine).

MaikKlein:
We also might want to add a lockfile for the generator.

MarijnS95:
Perhaps, though I prefer keeping everything (automatically) updated. It's the first time we run into this, nothing too bad IMO :)

PR #487, opened 2021-11-08
examples: Upgrade to winit 0.25

neurotok:
Current examples are based on old version of winint that was causing  thread panic on my system, this is well know issue that was already fixed by winint team. I updated examples to use most up to date 0.25.0 version

### DIFF
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -147,7 +147,7 @@ pub struct ExampleBase {
     pub swapchain_loader: Swapchain,
     pub debug_utils_loader: DebugUtils,
     pub window: winit::window::Window,
-    pub events_loop: RefCell<EventLoop<()>>,
+    pub event_loop: RefCell<EventLoop<()>>,
     pub debug_call_back: vk::DebugUtilsMessengerEXT,
 
     pub pdevice: vk::PhysicalDevice,
@@ -180,7 +180,7 @@ pub struct ExampleBase {
 
 impl ExampleBase {
     pub fn render_loop<F: Fn()>(&self, f: F) {
-        self.events_loop
+        self.event_loop
             .borrow_mut()
             .run_return(|event, _, control_flow| {
                 *control_flow = ControlFlow::Wait;
@@ -207,14 +207,14 @@ impl ExampleBase {
 
     pub fn new(window_width: u32, window_height: u32) -> Self {
         unsafe {
-            let events_loop = EventLoop::new();
+            let event_loop = EventLoop::new();
             let window = WindowBuilder::new()
                 .with_title("Ash - Example")
                 .with_inner_size(winit::dpi::LogicalSize::new(
                     f64::from(window_width),
                     f64::from(window_height),
                 ))
-                .build(&events_loop)
+                .build(&event_loop)
                 .unwrap();
             let entry = Entry::new();
             let app_name = CString::new("VulkanTriangle").unwrap();
@@ -528,7 +528,7 @@ impl ExampleBase {
                 .unwrap();
 
             ExampleBase {
-                events_loop: RefCell::new(events_loop),
+                event_loop: RefCell::new(event_loop),
                 entry,
                 instance,
                 device,
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -15,7 +15,7 @@ use std::ffi::{CStr, CString};
 use std::ops::Drop;
 
 use winit::{
-    event::{DeviceEvent, ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent},
+    event::{ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent},
     event_loop::{ControlFlow, EventLoop},
     platform::run_return::EventLoopExtRunReturn,
     window::WindowBuilder,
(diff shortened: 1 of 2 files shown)


examples/src/lib.rs
```diff
@@ -173,32 +180,42 @@ pub struct ExampleBase {
...
+                            state,
+                            ..
+                        }) => match (keycode, state) {
+                            (VirtualKeyCode::Escape, ElementState::Released) => {
+                                *control_flow = ControlFlow::Exit
+                            }
+                            _ => (),
+                        },
+                        _ => (),
+                    },
```
Ralith, line 206: We should only respond to keyboard inputs on the window, not globally, so this should be under `WindowEvent`, not `DeviceEvent`.

Ralith (changes requested):
Thanks!

examples/src/lib.rs
MarijnS95: You can also simplify these quite a bit by writing the expected value for `keycode` and `state` directly in the match statement. Both arms are the same, so you can combine those with `|` too. That's how I did it (without checking for `Release` state) 10 months ago:

https://github.com/Traverse-Research/ash/blob/65c6442d917cb599b18d54c5e852d5cf35aefc22/ash-window/examples/winit.rs#L37-L57

MarijnS95:
Note that there have been quite a few attempts at fixing this before, see for example #356 and #481 and all the linked issues/PRs. The main problem is that they all try to refactor the entire example to clean up resources inside the event loop because `fn run()` doesn't return, leading to massive hard-to-review PRs that just get procrastinated.

This PR on the other hand takes the "smart" approach of using `EventLoopExtRunReturn` which is supported on every platform _except_ iOS. At this point I'm inclined to just accept that and get rid of this issue for the majority of users, and hope someone (not me) will get around to reviewing and accepting the refactor to re-enable that platform once again.

examples/src/lib.rs
neurotok: I found this nifty I OR at https://sotrh.github.io/learn-wgpu/beginner/tutorial1-window/#the-code I hope that this will calm down clippy

examples/src/lib.rs
```diff
@@ -173,32 +180,39 @@ pub struct ExampleBase {
...
+        self.events_loop
+            .borrow_mut()
+            .run_return(|event, _, control_flow| {
+                *control_flow = ControlFlow::Wait;
+                f();
+                match event {
+                    Event::WindowEvent {
+                        ref event,
+                        window_id,
+                    } if window_id == self.window.id() => match event {
```
Ralith, line 192: Is `if window_id == self.window.id()` actually necessary here? This isn't a multi-window application.

MarijnS95: It's good practice to showcase it in an example, though.

examples/src/lib.rs
MarijnS95: I think it's going to ask you to combine `Event::WindowEvent` and the `WindowEvent::CloseRequested | WindowEvent::KeyboardInput` though (yes, nested `|` is supported in Rust now).

examples/src/lib.rs
Ralith: I don't think I agree; including it suggests to a naive reader that it does something meaningful in this case, which is misleading, and multi-window applications are rare regardless.

hYdos:
I'm able to open my PR again if you want to use #481

examples/src/lib.rs
MarijnS95: @Ralith What do you think? Shall we just remove this `if` and get on with it? One downside of this PR is that it uses `EventLoopExtRunReturn` which isn't supported on at least IOS (maybe also MacOS?) but at least it is _way_ less intrusive than all the other PRs that refactor the entire example while at it (necessary evil to have a place _inside_ the eventloop to free all resources :disappointed:).

Will need to do some title/commitbody curating though before this is mergeable.

Ralith: I think we should merge this, and I'd prefer that we do so with the guard removed (but it's a big improvement regardless).

MarijnS95: Fair enough. Fortunately @neurotok left this branch push-able for maintainers so I've just removed that, recombined the nested `matches` into one big one, and turned it into an `if-let` as per clippy's request. Should be ready to merge then, and we can worry about the other PRs when someone really wants to run this example on iOS. I think that's a necessary tradeoff to make for the gross of issues out there running older winit.

MarijnS95 (approved):

examples/src/lib.rs
Ralith: Yeah, I'm a big fan of incremental refactoring in general--helps manage reviewer bandwidth if nothing else :D

Ralith (approved):

examples/src/lib.rs
MarijnS95: Definitely, small isolated PRs are _a lot_ easier to process and the few extra minutes taken to create multiple PRs (not mentioning the act of separating ones changes during development, that's a whole different rabbit hole!) or clicking multiple merge buttons completely outweigh the _hours_ otherwise spent comprehending massive changesets :D

MarijnS95:
@Ralith I noticed you don't seem to (usually?) squash merge requests - appreciated for my PRs which are generally multiple commits that all have their own purpose, though I create separate PRs for each of them now - but might have been desired for this PR otherwise I'd have squashed the whole loose bunch before force-pushing :D

Ralith:
Whoops, my bad! I had a roll going with all your PRs and didn't look closely enough here. I'm a big fan of the rebase/rewrite oriented workflow in general, since it optimizes for readability of the final product, which is what matters most.

MarijnS95:
@Ralith I'm a little surprised since for me "squash and merge" is shown as default for this repository (and I don't recall that being a per user, per repository setting). We can probably disable the other options too now  though.

Ralith:
I think it's per-user per-repo, and I probably used it on some of my own PRs in the past.

MarijnS95:
Looks it simply remembers the last merge method that was used, indeed. I don't have access to a settings page for this repo so there's no way for me to configure it.

Ralith:
Something to hassle @MaikKlein about, I suppose. In the mean time I'll exercise extra caution.

MaikKlein:
> I don't have access to a settings page for this repo so there's no way for me to configure it.

I'll have a look.

> Something to hassle @MaikKlein about, I suppose. In the mean time I'll exercise extra caution.
Personally I prefer squash and merge a.

PR #489, opened 2021-11-10
extensions: Add VK_KHR_maintenance4

MarijnS95:
Trying a new approach for getter functions with `pNext` structures.  seems like this could work for the single item in `get_device_image_memory_requirements`, ~but it is probably a bit too weird for the array in `get_device_image_sparse_memory_requirements`:~

- ~Length of `next` array provides the count, so you'll have to fill that with `None` (ie. `vec![None; get_device_image_sparse_memory_requirements_len()]`) if no `pNext` pointers are used;~
- ~Slice is an array of borrows, so you'll need to create your (list of) next object(s) in one place, and create a new vector or slice just to hold the borrows to it;~
- ~Slice could perhaps be an iterator instead.~

Never mind, `SparseImageMemoryRequirements2` currently doesn't have anything that extends it. Should we keep this function similar to `get_image_sparse_memory_requirements2` (Vulkan 1.1) with just an `out: &mut []` array and separate `_len` function?


ash/src/extensions/khr/maintenance4.rs
```diff
@@ -0,0 +1,98 @@
...
+        );
+        let mut data = Vec::with_capacity(count as usize);
+        self.fns.get_device_image_sparse_memory_requirements_khr(
+            self.handle,
+            create_info,
+            &mut count,
+            data.as_mut_ptr(),
+        );
+        data.set_len(count as usize);
+        data
```
Ralith, line 84: This should probably be using `read_into_defaulted_vector`.

ash/src/extensions/khr/maintenance4.rs
```diff
@@ -0,0 +1,98 @@
...
+            handle: device.handle(),
+            fns,
+        }
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetDeviceBufferMemoryRequirementsKHR.html>"]
+    pub unsafe fn get_device_buffer_memory_requirements<'a>(
+        &self,
+        create_info: &vk::DeviceBufferMemoryRequirementsKHR,
+        next: Option<&'a mut impl vk::ExtendsMemoryRequirements2>,
```
Ralith, line 28: Unfortunately I think this sucks to call, because passing `None` (the overwhelmingly common case) leaves the implied type variable ambiguous. I wonder if we could get away with `dyn` instead of `impl` here? Not using (the existing form of) `push_next`, of course, but maybe by hand or with changes to the trait interface?

Ralith (changes requested):
> Should we keep this function similar to get_image_sparse_memory_requirements2 (Vulkan 1.1) with just an out: &mut [] array and separate _len function?

I think it's best to be consistent there, yeah.

ash/src/extensions/khr/maintenance4.rs
MarijnS95: That was omitted when quickly fixing up the wrong trait type - I want to split this back out anyway to be consistent with the other `_len()` functions.

ash/src/extensions/khr/maintenance4.rs
MarijnS95: Fair point; `dyn` just works but then again I'm not sure if we shouldn't also keep this consistent with just an `out: &mut vk::MemoryRequirements2` like the slice case.

Though, the idea of this PR is to see if we could/should also change the others. It won't work for the slice case (first commit with the wrong trait bound, before this was force-pushed) since you anyway need to query `_len()` before knowing how many `None`/`Some(&next)` elements to pass in the first place.

MarijnS95: EDIT: I'm pretty sure I compile-tested this on a different machine to see `impl` fail and `dyn` succeed, but see `error[E0277]: the size for values of type `(dyn definitions::ExtendsMemoryRequirements2 + 'a)` cannot be known at compilation time` now which wasn't an issue before...

Ralith: Right, `push_next` currently requires `Sized` but could be modified not to, or replaced with handwritten code to illustrate the concept.

I think I lean towards mundane consistency here regardless, in part because it's a bit weird to get some of your result data via output argument and some via return value rather than all via output argument. Also keeps things simple. Not a very strong opinion, though.

MarijnS95: Let's go for simple and consistency then. We can always revisit this in a later PR for the entire API surface at once.

Ralith (approved):

ash/src/extensions/khr/maintenance4.rs
```diff
@@ -0,0 +1,88 @@
...
+        out: &mut [vk::SparseImageMemoryRequirements2],
+    ) {
+        let mut count = out.len() as u32;
+        self.fns.get_device_image_sparse_memory_requirements_khr(
+            self.handle,
+            create_info,
+            &mut count,
+            out.as_mut_ptr(),
+        );
+        assert_eq!(count, out.len() as u32);
```
MarijnS95, line 74: @Ralith I slipped in an `assert_eq` here without mentioning. I have a change locally to apply this to the other functions of this sort - do you think it's worth adding that everywhere consistently?

It seems very unlikely for this to ever change, hence an assert might be a good idea in the strange event that a driver happens to overwrite it after all (ie. playing around with beta drivers etc).

ash/src/extensions/khr/maintenance4.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[derive(Clone)]
+pub struct Maintenance4 {
+    handle: vk::Device,
+    fns: vk::KhrMaintenance4Fn,
+}
+
+impl Maintenance4 {
+    pub fn new(instance: &Instance, device: &Device) -> Self {
+        let fns = vk::KhrMaintenance4Fn::load(|name| unsafe {
```
filnet, line 14: please factorize  `device.handle()` like here : https://github.com/MaikKlein/ash/pull/494/files#diff-6eb43a992bec48c8676fdb7916ad74eb423692b90b465dfa8f44717e081d47d8

MarijnS95: This has been merged long before your PR - in fact you even factorized this out in https://github.com/MaikKlein/ash/pull/494/files#diff-42685c200f5a13446e740816683a82dc8df0dcad950ed9a1b859295870f1b2c1R14-R18 already :)

filnet: Oops. Sorry for the noise.

PR #492, opened 2021-11-11
Fix broken vulkan-tutorial link in README.md

GlowingUmbreon:
Fixes the vulkan-tutorial.com link inside of the readme.


MarijnS95 (approved):
Good catch!

PR #493, opened 2021-11-12
extensions: Add VK_KHR_present_wait

filnet:


ash/src/extensions/khr/present_wait.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkWaitForPresentKHR.html>"]
+    pub unsafe fn wait_for_present(
+        &self,
+        swapchain: vk::SwapchainKHR,
+        present_id: u64,
+        timeout: u64,
+    ) -> VkResult<()> {
+        self.fns
+            .wait_for_present_khr(self.handle, swapchain, present_id, timeout)
+            .result()
```
MarijnS95, line 33: Heh I should probably clean this up in the rest of the code as I've been using `.into()` here for the most part whereas some extension code is using `.result()`.

ash/src/extensions/khr/present_wait.rs
```diff
@@ -0,0 +1,47 @@
...
+            fns,
+        }
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkWaitForPresentKHR.html>"]
+    pub unsafe fn wait_for_present(
+        &self,
+        swapchain: vk::SwapchainKHR,
+        present_id: u64,
+        timeout: u64,
```
MarijnS95, line 29: Please replace this with a `Duration`, so that there's no debate what the format is - milliseconds? microseconds?

ash/src/extensions/khr/present_wait.rs
filnet: I'll do a cleanup pass in all extensions.

MarijnS95: Separate PR please.

MarijnS95 (approved):

ash/src/extensions/khr/present_wait.rs
filnet: Sure.
The cleanup will include:
- calling `into()` instead of `result()`,
- renaming the function pointer members to `fns`,
- moving `fn name()` down with the `fp()` and all.

filnet: Not sure about moving `fn name()` down.
The recently added extension `dynamic_rendering` has the `fn name()` and `fn fp()` near the top.
What is the preferred layout ?

MarijnS95: > * calling `into()` instead of `result()`,

I think calling `.result()` everywhere is more clear (and perhaps removing the `From<vk::Result> for VkResult<()>` entirely). Afaik Rust can't deduce the type when using this with a `?` operator in `.into()?;` cases.

CC @Ralith I presume you prefer `.result()` too instead of having to guess what `From`/`Into` has been implemented?

> * renaming the function pointer members to `fns`,

Lovely, that annoyed me when writing `maintenance4` and `dynamic_rendering`

> * moving `fn name()` down with `fn fp()` and all.

Yes please, my eye also fell on that.

> Not sure about moving `fn name()` down.
> The recently added extension `dynamic_rendering` has the `fn name()` and `fn fp()` near the top.

Yeah that annoyed me when writing `maintenance4`, especially after noticing I put it in the wrong place myself in `dynamic_rendering`. Let's have:

- `fn new()`;
- All the commands;
- `fn name()`;
- `fn fp();
- `fn instance/device()`.

filnet:
Please merge this PR so I can include this new extension in the upcoming cleanup PR.

ash/src/extensions/khr/present_wait.rs
MarijnS95: Exactly.

MarijnS95:
Feel free to fix that in this PR instead, so that this extension is introduced with the clean way of doing things from the get-go.

ash/src/extensions/khr/present_wait.rs
filnet: > I think calling .result() everywhere is more clear (and perhaps removing the From<vk::Result> for VkResult<()> entirely). Afaik Rust can't deduce the type when using this with a ? operator in .into()?; cases.

There are 26 locations, outside of extensions, where `into()` is called (24 in device.rs and 2 in instance.rs).

filnet:
> 
> 
> Feel free to fix that in this PR instead, so that this extension is introduced with the clean way of doing things from the get-go.

Will do.

filnet:
What is preferred :
``` rust
self.fns
    .wait_for_present_khr(
        self.handle,
        swapchain,
        present_id,
        timeout.as_nanos().try_into().unwrap(),
    )
    .result()
```
or
``` rust
let timeout = timeout.as_nanos().try_into().unwrap();
self.fns
    .wait_for_present_khr(self.handle, swapchain, present_id, timeout)
    .result()
```
?

filnet:
I updated the 2nd choice to not alias timeout...

MarijnS95:
@filnet IMO aliasing is preferred, as is calling the function right in the expression - but following the comments in #496 I'm not sure if we should use `Duration` after all.

ash/src/extensions/khr/present_wait.rs
MarijnS95: > outside of extensions

Those files are also hand-written helpers, just like the extensions.

filnet: > Those files are also hand-written helpers, just like the extensions.

Yes, they are handled here : https://github.com/MaikKlein/ash/pull/495.
Should have linked the PRs earlier...

MarijnS95 (approved):
Thanks, sorry for the wait!

PR #494, opened 2021-11-12
extensions: make naming and layout consistent accross all extensions

filnet:
breaking change:
- some extensions were exposing instance() instead of device()

includes:
- renaming function pointer member to fns
- moving name(), fp(), device()/instance() functions at end of file
- adding missing device()/instance() functions
- ~~using result() instead of into()~~

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,11 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `VK_KHR_present_wait` extension (#493)
+- Added `VK_KHR_present_wait` device extension (#493)
 
 ### Changed
 
-- Removed `instance` function from `extensions::ext::ExtendedDynamicState` and `extensions::khr::PushDescriptor` (#494)
+- Device extension `ext::ExtendedDynamicState`, `khr::PushDescriptor`, `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` now expose `fn device()` instead of `fn instance()` (#494, #499)
+- Changed `khr::PipelineExecutableProperties::new()` and `khr::TimelineSemaphore::new()` to take `instance` and `device` as arguments (#499)
+
+### Removed
+
+- Removed `From<vk::Result>` trait for `VkResult` (#495)
+- Removed `device` argument from `ext::DebugMarkers::debug_marker_set_object_name` function (#494)
+- Removed `instance` argument from `ext::DebugUtils::submit_debug_utils_message` function (#499)
 
 ## [0.33.3] - 2021-09-08
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `VK_KHR_present_wait` extension (#493)
 
+### Changed
+
+- Removed `instance` function from `extensions::ext::ExtendedDynamicState` and `extensions::khr::PushDescriptor` (#494)
+
 ## [0.33.3] - 2021-09-08
 
 ### Added
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,13 +12,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Device extension `ext::ExtendedDynamicState`, `khr::PushDescriptor`, `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` now expose `fn device()` instead of `fn instance()` (#494, #499)
+- Device extension `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` now expose `fn device()` instead of `fn instance()` (#499)
 - Changed `khr::PipelineExecutableProperties::new()` and `khr::TimelineSemaphore::new()` to take `instance` and `device` as arguments (#499)
 
 ### Removed
 
-- Removed `From<vk::Result>` trait for `VkResult` (#495)
+- Removed `device()` function from `khr::Synchronization2` device extension (#494)
+- Removed `instance()` function from `ext::ExtendedDynamicState`, `khr::PushDescriptor`, `ext::ToolingInfo` and `khr::GetPhysicalDeviceProperties2` instance extensions (#494)
 - Removed `device` argument from `ext::DebugMarkers::debug_marker_set_object_name` function (#494)
+- Removed `From<vk::Result>` trait for `VkResult` (#495)
 - Removed `instance` argument from `ext::DebugUtils::submit_debug_utils_message` function (#499)
 
 ## [0.33.3] - 2021-09-08
--- a/ash/src/extensions/ext/debug_marker.rs
+++ b/ash/src/extensions/ext/debug_marker.rs
@@ -7,16 +7,16 @@ use std::mem;
 #[derive(Clone)]
 pub struct DebugMarker {
     handle: vk::Device,
-    fns: vk::ExtDebugMarkerFn,
+    fp: vk::ExtDebugMarkerFn,
 }
 
 impl DebugMarker {
     pub fn new(instance: &Instance, device: &Device) -> Self {
         let handle = device.handle();
-        let fns = vk::ExtDebugMarkerFn::load(|name| unsafe {
+        let fp = vk::ExtDebugMarkerFn::load(|name| unsafe {
             mem::transmute(instance.get_device_proc_addr(handle, name.as_ptr()))
         });
-        Self { handle, fns }
+        Self { handle, fp }
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkDebugMarkerSetObjectNameEXT.html>"]
@@ -24,7 +24,7 @@ impl DebugMarker {
         &self,
         name_info: &vk::DebugMarkerObjectNameInfoEXT,
     ) -> VkResult<()> {
-        self.fns
+        self.fp
             .debug_marker_set_object_name_ext(self.handle, name_info)
             .result()
     }
@@ -35,13 +35,13 @@ impl DebugMarker {
         command_buffer: vk::CommandBuffer,
         marker_info: &vk::DebugMarkerMarkerInfoEXT,
     ) {
-        self.fns
+        self.fp
             .cmd_debug_marker_begin_ext(command_buffer, marker_info);
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdDebugMarkerEndEXT.html>"]
     pub unsafe fn cmd_debug_marker_end(&self, command_buffer: vk::CommandBuffer) {
-        self.fns.cmd_debug_marker_end_ext(command_buffer);
+        self.fp.cmd_debug_marker_end_ext(command_buffer);
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCmdDebugMarkerInsertEXT.html>"]
@@ -50,7 +50,7 @@ impl DebugMarker {
         command_buffer: vk::CommandBuffer,
         marker_info: &vk::DebugMarkerMarkerInfoEXT,
     ) {
-        self.fns
+        self.fp
             .cmd_debug_marker_insert_ext(command_buffer, marker_info);
     }
 
@@ -59,7 +59,7 @@ impl DebugMarker {
     }
 
     pub fn fp(&self) -> &vk::ExtDebugMarkerFn {
-        &self.fns
+        &self.fp
     }
 
     pub fn device(&self) -> vk::Device {
--- a/ash/src/extensions/ext/debug_marker.rs
+++ b/ash/src/extensions/ext/debug_marker.rs
@@ -22,11 +22,10 @@ impl DebugMarker {
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkDebugMarkerSetObjectNameEXT.html>"]
     pub unsafe fn debug_marker_set_object_name(
         &self,
-        device: vk::Device,
         name_info: &vk::DebugMarkerObjectNameInfoEXT,
     ) -> VkResult<()> {
         self.fns
-            .debug_marker_set_object_name_ext(device, name_info)
+            .debug_marker_set_object_name_ext(self.handle, name_info)
             .result()
     }
 
--- a/ash/src/extensions/ext/debug_marker.rs
+++ b/ash/src/extensions/ext/debug_marker.rs
@@ -6,19 +6,17 @@ use std::mem;
 
 #[derive(Clone)]
 pub struct DebugMarker {
-    debug_marker_fn: vk::ExtDebugMarkerFn,
+    handle: vk::Device,
+    fns: vk::ExtDebugMarkerFn,
 }
 
 impl DebugMarker {
     pub fn new(instance: &Instance, device: &Device) -> Self {
-        let debug_marker_fn = vk::ExtDebugMarkerFn::load(|name| unsafe {
-            mem::transmute(instance.get_device_proc_addr(device.handle(), name.as_ptr()))
+        let handle = device.handle();
+        let fns = vk::ExtDebugMarkerFn::load(|name| unsafe {
+            mem::transmute(instance.get_device_proc_addr(handle, name.as_ptr()))
         });
-        Self { debug_marker_fn }
-    }
-
-    pub fn name() -> &'static CStr {
-        vk::ExtDebugMarkerFn::name()
+        Self { handle, fns }
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkDebugMarkerSetObjectNameEXT.html>"]
@@ -27,7 +25,7 @@ impl DebugMarker {
         device: vk::Device,
         name_info: &vk::DebugMarkerObjectNameInfoEXT,
     ) -> VkResult<()> {
-        self.debug_marker_fn
+        self.fns
             .debug_marker_set_object_name_ext(device, name_info)
             .result()
     }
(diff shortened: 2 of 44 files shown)


ash/src/extensions/ext/extended_dynamic_state.rs
```diff
@@ -6,23 +6,17 @@ use std::ptr;
 
 #[derive(Clone)]
 pub struct ExtendedDynamicState {
-    handle: vk::Instance,
-    extended_dynamic_state_fn: vk::ExtExtendedDynamicStateFn,
+    handle: vk::Device,
```
filnet, line 9: breaking change here...

ash/src/extensions/ext/extended_dynamic_state.rs
```diff
@@ -187,11 +177,15 @@ impl ExtendedDynamicState {
...
+    pub fn name() -> &'static CStr {
+        vk::ExtExtendedDynamicStateFn::name()
+    }
+
     pub fn fp(&self) -> &vk::ExtExtendedDynamicStateFn {
-        &self.extended_dynamic_state_fn
+        &self.fns
     }
 
-    pub fn instance(&self) -> vk::Instance {
```
filnet, line 194: Breaking change !

ash/src/extensions/khr/push_descriptor.rs
```diff
@@ -53,21 +47,24 @@ impl PushDescriptor {
...
+    pub fn name() -> &'static CStr {
+        vk::KhrPushDescriptorFn::name()
     }
 
     pub fn fp(&self) -> &vk::KhrPushDescriptorFn {
-        &self.push_descriptors_fn
+        &self.fns
     }
 
-    pub fn instance(&self) -> vk::Instance {
```
filnet, line 70: Breaking change !

MarijnS95:
I guess we want to get this in after #495?

filnet:
> I guess we want to get this in after #495?

The two PRs are independent AFAIK.

MarijnS95:
This PR also converts some of the `.into()` calls to `.result()`. Those should disappear as soon as #495 is merged, hence the suggestion to merge that one in first (and reduce the size of this diff a little).

filnet:
> This PR also converts some of the `.into()` calls to `.result()`. Those should disappear as soon as #495 is merged, hence the suggestion to merge that one in first (and reduce the size of this diff a little).

Ok. Makes sense.

ash/src/extensions/ext/debug_marker.rs
```diff
@@ -6,19 +6,17 @@ use std::mem;
 
 #[derive(Clone)]
 pub struct DebugMarker {
-    debug_marker_fn: vk::ExtDebugMarkerFn,
+    handle: vk::Device,
```
MarijnS95, line 9: The general consensus was is that we only hold on to the instance/device `handle` if it is actively used in the functions.

You'll have to change the function signatures of ie. `debug_marker_set_object_name` to remove `device: vk::Device,` from it since this is now available as `self.handle`.

More importantly, it prevents callers from using these functions with a different device than it was originally retrieved with through `get_device_proc_addr` and simplifies the calling some, too.

ash/src/extensions/ext/debug_utils.rs
```diff
@@ -7,22 +7,16 @@ use std::mem;
...
+    fns: vk::ExtDebugUtilsFn,
 }
 
 impl DebugUtils {
     pub fn new(entry: &Entry, instance: &Instance) -> Self {
-        let debug_utils_fn = vk::ExtDebugUtilsFn::load(|name| unsafe {
-            mem::transmute(entry.get_instance_proc_addr(instance.handle(), name.as_ptr()))
+        let handle = instance.handle();
+        let fns = vk::ExtDebugUtilsFn::load(|name| unsafe {
+            mem::transmute(entry.get_instance_proc_addr(handle, name.as_ptr()))
```
MarijnS95, line 17: Almost all the calls here take a `device` or a child of the `device` (`CommandBuffer`, `Queue`) so we might as well change this to `get_device_proc_addr` and hold on to `device` in addition to `instance`. However, the `_messenger` and `_message` calls are _not_ device specific and AFAIK it is not allowed to query these through `get_device_proc_addr`. We could think something up for these (in a future PR) or accept the fact that these might wrap a jump table and that the device needs to be passed to those specific functions, as we cannot easily rework the `load` function.

A future PR may also try to split the `load` up in functions that receive a `device` or child of the `device`, so that we can load them separately.

ash/src/extensions/khr/pipeline_executable_properties.rs
```diff
@@ -7,23 +7,16 @@ use std::mem;
...
-            handle: instance.handle(),
-            pipeline_executable_properties_fn,
-        }
-    }
-
-    pub fn name() -> &'static CStr {
-        vk::KhrPipelineExecutablePropertiesFn::name()
+        let handle = instance.handle();
+        let fns = vk::KhrPipelineExecutablePropertiesFn::load(|name| unsafe {
+            mem::transmute(entry.get_instance_proc_addr(handle, name.as_ptr()))
```
MarijnS95, line 17: It looks like `PipelineExecutableProperties` uses the `device` in every call so feel free to switch this all to use the device. Fine to do it in a separate PR in a way that gives the least amount of work/conflicts for you with this PR :)

ash/src/extensions/khr/timeline_semaphore.rs
```diff
@@ -7,22 +7,16 @@ use std::mem;
...
+    fns: vk::KhrTimelineSemaphoreFn,
 }
 
 impl TimelineSemaphore {
     pub fn new(entry: &Entry, instance: &Instance) -> Self {
-        let timeline_semaphore_fn = vk::KhrTimelineSemaphoreFn::load(|name| unsafe {
-            mem::transmute(entry.get_instance_proc_addr(instance.handle(), name.as_ptr()))
+        let handle = instance.handle();
+        let fns = vk::KhrTimelineSemaphoreFn::load(|name| unsafe {
+            mem::transmute(entry.get_instance_proc_addr(handle, name.as_ptr()))
```
MarijnS95, line 17: Same comment here, `TimelineSemaphore` is a good contender for getting `device`-ified. I can take care of these changes as well if you prefer :)

ash/src/extensions/khr/pipeline_executable_properties.rs
MarijnS95: I decided to document this and take care of it (also `TimelineSemaphore`) in #499 :wink:

ash/src/extensions/khr/timeline_semaphore.rs
MarijnS95: #499

ash/src/extensions/ext/debug_marker.rs
filnet: Done

MarijnS95 (approved):
Not seeing any red flags with this PR, looks good overall. Thanks for the hard work!

filnet:
> Not seeing any red flags with this PR, looks good overall. Thanks for the hard work!

You're welcome. 

Just pushed some commit log rewording.

Changelog.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `VK_KHR_present_wait` extension (#493)
 
+### Changed
+
+- Removed `instance` function from `extensions::ext::ExtendedDynamicState` and `extensions::khr::PushDescriptor` (#494)
```
MarijnS95, line 15: Should we mention that these were replaced with `device` instead?

Also, theoretically we should list the removal of the `instance` and `device` parameters from the functions in the last two commits.

filnet:
There are two "extensions" I left untouched:
- `ext::PhysicalDeviceDrm`
- `experimental::amd`

Changelog.md
filnet: I'll cleanup the Changelog.

MarijnS95: Fyi https://keepachangelog.com/en/1.0.0/ says to have a special `### Removed` header for these :)

(Though in the most cases we're just moving arguments because they're better, or already, provided elsewhere)

filnet: Yes, and I can't find a good wording (precise and succinct).

filnet: Giving it another try...

MarijnS95: Not saying it any different than it is may be a good start. For example, nobody really knows what "normalize" means in this context. Just stating that `Extension X and Y are now exposing `fn device()` instead of `fn instance()` because they are device extensions` seems simple and good enough to me?

Changelog.md
```diff
@@ -8,7 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add `VK_KHR_present_wait` extension (#493)
+- Added `VK_KHR_present_wait` device extension (#493)
+
+### Changed
+
+- Extensions `ext::ExtendedDynamicState` and `khr::PushDescriptor` are now exposing `fn device()` instead of `fn instance()` because they are device extensions.
+- Extensions `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` are now exposing `fn device()` instead of `fn instance()` because they are device extensions.
```
MarijnS95, line 16: `s/./(#494)`?

Changelog.md
```diff
@@ -8,7 +8,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Device extension `khr::TimelineSemaphore` now exposes `fn device()` instead of `fn instance()` (#499)
+- Changed `khr::TimelineSemaphore::new()` to take `instance` and `device` as arguments (#499)
+
+### Removed
+
+- Removed `device` argument from `ext::DebugMarkers::debug_marker_set_object_name` function (#494)
+- Removed `From<vk::Result>` trait for `VkResult` (#495)
+- Removed `instance` argument from `ext::DebugUtils::submit_debug_utils_message` function (#499)
+- Removed `device` argument from `khr::PipelineExecutableProperties` functions (#499)
+- Removed `device` argument from `khr::TimelineSemaphore` functions (#499)
```
MarijnS95, line 28: Thanks, also for taking care of #499. Just not sure if we should mention some single PRs this often in the changelog.

For example, the `### Removed` section is afaik for removed API - changing the arguments to remove ie. `device` that's already available in the struct is already documented under `### Changed`, no need to repeat that again.

Furthermore, we can probably make lists, ie turn this:

```markdown
- Device extension `ext::ExtendedDynamicState` now exposes `fn device()` instead of `fn instance()` (#494)
- Device extension `khr::PushDescriptor` now exposes `fn device()` instead of `fn instance()` (#494)
- Device extension `khr::PipelineExecutableProperties` now exposes `fn device()` instead of `fn instance()` (#499)
```

into:

```markdown
- Device extensions `ext::ExtendedDynamicState`, `khr::PushDescriptor` and `khr::PipelineExecutableProperties` now expose `fn device()` instead of `fn instance()` (#494, #499)
```

I think that's much more more concise (more readable).

MarijnS95: @Ralith What are your thoughts on this, and this PR in general? Seems to be pretty much ready to go in besides some changelog bikeshedding, unless you have some burning comments?

Would like to get this in and do a release soon, it has been way too long.

ash/src/extensions/ext/extended_dynamic_state.rs
Ralith, line 9: Should we be storing handles that aren't used by any of the functions? This seems to occur for a number of extensions.

Changelog.md
Ralith: Your proposed tweaks to the changelog SGTM, and overall this looks like great cleanup, though I haven't reviewed every single change in detail.

ash/src/extensions/ext/extended_dynamic_state.rs
MarijnS95: Indeed, I don't think we should store it (beyond using it in the constructor to get device-specific functions) if none of the commands need it. IMO it's fine to just remove `fn instance()` in this case without replacing it with `fn device()` unnecessarily - no need to be "consistent" for that across extension helpers.

After all, it seems `fn instance()` and `fn device()` were added for convenience some time ago, but I would be totally fine to just remove these getters. I doubt they add much, the instance/device is kept for internal use and if it is needed outside of that the caller probably wants to have have the full `ash::Device` (or another extension - which undoubtedly also carries the instance/device where necessary) instead.

Ralith: That sounds good to me as well, though in the interest of controlling scope creep, it'd be fine with me to leave for future work the removal of the getter from cases which do actually need to store a handle.

Changelog.md
filnet: The changes are pretty boring. It's mostly code gardening.

ash/src/extensions/ext/extended_dynamic_state.rs
filnet: Then `fn fp()` can go to ?

MarijnS95: Unlike `instance` and `device` the caller doesn't have access to the function pointers which are loaded in the constructor, I'd keep `fn fp()` to access those.

That said though these function pointers are only used circumvent the "overhead" induced by the headers, and eventual functions that did not have their helper counterpart implemented (which should be unlikely / not present at all).

I'd keep them for now and discuss the removal of `device()`/`instance()`/`fp()` in a future issue/PR. The only thing I'd remove in this PR is the `device` and `instance` members (and their matching getters, of course) if they are not used by any of the helper functions.

Changelog.md
MarijnS95: @filnet Yes, curating the changelog (usually when preparing for a release) is pretty tough. We usually end up going for just the commit titles since the last release, but that's usually not fully representative nor are all the titles good enough for direct use :)

ash/src/extensions/ext/extended_dynamic_state.rs
filnet: I think it does not make sense to expose `device()`/`instance()` based on wether they are used internally or not.
They should either be always exposed for all extensions or never but not an in between.

MarijnS95: This is not about exposing those getter functions but about unnecessarily storing these fields in the first place, increasing struct size (concerns have been raised over this in the past, specifically for embedded applications).

filnet: I'll remove them then.

ash/src/extensions/ext/full_screen_exclusive.rs
```diff
@@ -77,8 +70,12 @@ impl FullScreenExclusive {
             .result_with_success(present_modes)
     }
 
+    pub fn name() -> &'static CStr {
+        vk::ExtFullScreenExclusiveFn::name()
+    }
+
     pub fn fp(&self) -> &vk::ExtFullScreenExclusiveFn {
```
aloucks, line 77: Perhaps you should unify the method name and the field name if you're renaming it anyway.

filnet: Should I rename the function to `fns()` or the field name to `fp` ?

Ralith: The field name is purely internal, so I don't think it matters much what it is. The functions should not be renamed as that would be gratuitous breakage.

MarijnS95: @Ralith I think `fp` works then, time to sign-off and merge this PR?

Ralith: That particular change seems fine to me, yeah. I don't know if I'll have the energy to intensively review this whole PR in the short term, but considering it's straightforward refactoring I see no reason for you to wait on me if you've already done so.

Changelog.md
```diff
@@ -8,7 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Added `VK_KHR_present_wait` device extension (#493)
+
+### Changed
+
+- Device extension `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` now expose `fn device()` instead of `fn instance()` (#499)
+- Changed `khr::PipelineExecutableProperties::new()` and `khr::TimelineSemaphore::new()` to take `instance` and `device` as arguments (#499)
+
+### Removed
+
+- Removed `device()` function from `ext::ExtendedDynamicState`, `khr::DrawIndirectCount`, `khr::DynamicRendering`, `khr::PushDescriptor`, `khr::Synchronization2`, `nv::DeviceDiagnosticCheckpoints` and `nv::MeshShader` device extensions (#494)
```
Ralith, line 20: `ExtendedDynamicState` originally had `instance()`, not `device()`. Perhaps this is true of others as well.

Ralith (changes requested):

ash/src/extensions/ext/full_screen_exclusive.rs
MarijnS95: I have previously gone through the entire PR and checked that there are no unintended changes besides what we discussed and what's listed in the changelog. Will do so again with the `fp` changes and new changelog, validate how this all works against one of our Vulkan applications that heavily uses a bunch of extensions, and perhaps just merge and perform a release after that.

Changelog.md
filnet: Fixed.
`PushDescriptor` was also wrong.

Ralith (approved):
Approving wrt. the issues I've previously specifically highlighted.

Changelog.md
```diff
@@ -8,7 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Added `VK_KHR_present_wait` device extension (#493)
+
+### Changed
+
+- Device extension `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` now expose `fn device()` instead of `fn instance()` (#499)
+- Changed `khr::PipelineExecutableProperties::new()` and `khr::TimelineSemaphore::new()` to take `instance` and `device` as arguments (#499)
+
+### Removed
+
+- Removed `device()` function from `khr::DrawIndirectCount`, `khr::DynamicRendering`, `khr::Synchronization2`, `nv::DeviceDiagnosticCheckpoints` and `nv::MeshShader` device extensions (#494)
```
MarijnS95, line 20: `DrawIndirectCount`, `DynamicRendering`, `DeviceDiagnosticCheckpoints`, `MeshShader` did not have a `device()` function before this PR. They were introduced here earlier, and removed later.

MarijnS95 (commented):
Changelog seems fine besides this - will re-re-re-review the rest of the code later on :)

Changelog.md
filnet: Fixed

MarijnS95 (approved):
This all looks good now. Thanks for your perseverance in dealing with all the review comments and repeated large changes!

PR #495, opened 2021-11-12
remove From<vk::Result> for VkResult

filnet:
this is a breaking change

see https://github.com/MaikKlein/ash/pull/493


MarijnS95 (approved):
Yeah I think this makes the code a bit more sensible and readable.

filnet:
Should I update the Changelog.md ?

If yes, please suggest a wording for the change.

Ralith (approved):

PR #496, opened 2021-11-15
timeouts as std::time::Duration instead of u64

filnet:


aloucks:
I guess I'd like to eventually see the wrapper also fully generated from `vk.xml` in addition to the raw function pointers and loaders. We can derive things like slices, enums, and primitives from the spec, but it becomes more difficult with richer APIs like `Duration`. I think it would be better to just rename the variable `timeout_ns` and have the user do `acquire_next_image(swapchain, duration.as_nanos(), semaphore, fence)` when they call the API.

Ralith:
I also weakly favor the simplicity and consistency of a u64. Taking a value that has a different range than can actually be passed down is weird, even if it's mostly an academic issue here.

MarijnS95:
This suggestion came from https://github.com/MaikKlein/ash/pull/493#discussion_r748219210, but the downside is that `as_nanos()` returns a `u128`. Upon consideration, it's perhaps better to rename the variables to `timeout_ns` instead - for this low-level case that's a better way of conveying the desired unit (without having to open the Vulkan spec).

Storing a `u64::MAX` in a `u128` `Duration` seems odd, too.

filnet:
Note that we'd have to change [device::wait_semaphores()](https://github.com/MaikKlein/ash/blob/e120dd78387dd44cf98f19a7be4e5eaa9dab7ece/ash/src/device.rs#L1798) too.

Anyways, I'm closing this PR and will update the related one.

filnet:
PS: Vulkan might or will have to switch to `u128` for durations in the future.

filnet:
Sorry for continuing the discussion about timeout here. 

On second thought, renaming `timeout` to `timeout_ns` might be counterproductive if this code gets generated some day.

The `_ns` suffix might help clarify things but I'll will still open the Vulkan spec (or have it open already).

Ralith:
> PS: Vulkan might or will have to switch to u128 for durations in the future.

Why? 2^64 nanoseconds is almost 600 years.

filnet:
> 
> 
> > PS: Vulkan might or will have to switch to u128 for durations in the future.
> 
> Why? 2^64 nanoseconds is almost 600 years.

For the hopeful ?

I get your point and forgot that we lived in a 60 Hz world. 
But who knows... and why u64 then and not u32 ?

Ralith:
> I get your point and forgot that we lived in a 60 Hz world.

That wasn't my point. 2^64 *nanoseconds*, not 1/60ths of a second, is ~600 years.

>  But who knows... and why u64 then and not u32 ?

2^32 nanoseconds is 4 seconds.

filnet:
> That wasn't my point. 2^64 _nanoseconds_, not 1/60ths of a second, is ~600 years.

I was just goofing around and I understand that there is no point in waiting for more than 600 years for an operation to finish.

But one day u64 might be regarded as we regard u8 nowadays. It is starting to happen to u32.

PR #499, opened 2021-11-19
Use device in pipeline_executable_properties and timeline_semaphore

MarijnS95:
Of all the extensions calling get_instance_proc_addr only these two remain that should use the "optimized" device-specific function pointers, since all functions take the device as argument (a child of the device such as a command buffer or queue is also possible, but not applicable here) and may otherwise have to go through a dispatch function [1].

Only VK_EXT_debug_utils remains where all but three of the functions are device (or device-child) specific.  This however requires the autogenerated loader to be separated out into two stages (and debug utils are generally initialized before creating a logical device), making it worth to accept the dispatch function unless this extension struct is split, too.

[1]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetDeviceProcAddr.html

### DIFF
--- a/ash/src/extensions/ext/debug_utils.rs
+++ b/ash/src/extensions/ext/debug_utils.rs
@@ -131,13 +131,12 @@ impl DebugUtils {
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkSubmitDebugUtilsMessageEXT.html>"]
     pub unsafe fn submit_debug_utils_message(
         &self,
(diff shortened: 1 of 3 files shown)


Rua:
Not directly related to the PR, but how do you know which functions have such "optimised" versions?

MarijnS95:
@Rua guessing and interpreting [1], there's no actual doc stating what functions can go through this mechanism.

Ralith (approved):

PR #501, opened 2021-11-23
generator: Add missing documentation to bitflag extension variants

MarijnS95:


generator/src/lib.rs
```diff
@@ -2374,12 +2377,23 @@ pub fn generate_constant<'a>(
...
     let ident = format_ident!("{}", name);
+
+    let notation = constant.notation().map(|n| {
+        let lowercase = n.to_lowercase();
+        if lowercase.contains("backwards") || lowercase.starts_with("alias") {
+            quote!(#[deprecated = #n])
+        } else {
+            quote!(#[doc = #n])
+        }
+    });
```
MarijnS95, line 2388: Some duplicated code, perhaps we should move this as a default-implementation to `trait ConstantExt`.

Ralith: Did we ever work out what was going on with @MaikKlein's generator rewrite?

Ralith (approved):

generator/src/lib.rs
MarijnS95: There are no direct plans to finish it that I'm aware of. I haven't personally looked through it enough to see if I/we could continue on and finish it, nor how far along it is in the first place.

PR #502, opened 2021-11-23
generator: Use "Backwards-compatible" comment to detect deprecated alias

MarijnS95:
vk.xml now contains the comment text "Backwards-compatible alias containing a typo" which we can use to detect intentional renames, without needing to specify explicit overrides/exceptions in the generator anymore.

These deprecated constants exist for the sole reason of backwards compatibility which Vulkan cannot permit itself to remove in the C headers, but are unreasonable for crate authors to use anyway due to their `#[deprecated]` annotation whose cargo-check warnings are easy to fix by just using the non-deprecated variant instead.  Furthermore, Ash is still allowing itself to perform breaking changes in its releases making this the perfect time to get rid of all these useless variants and the generator support code that comes with it.  No need to come up with a "more proper" variant name if we don't generate those that "intentionally" fail to adhere to the "enum variant name" specification in the first place.

---

~Perhaps we should reach a point where we just stop generating any variants with this discrepancy at all, given that we're releasing ash where breaking changes are allowed and `#[deprecated]` triggers check/clippy warnings already anyway.  I doubt anyone uses these (EDIT: Bar all the missing `#[deprecated]` annotations from https://github.com/MaikKlein/ash/pull/501/files... :sweat_smile:).~ Done!

### DIFF
--- a/ash/src/vk/extensions.rs
+++ b/ash/src/vk/extensions.rs
@@ -1972,11 +1972,6 @@ impl StructureType {
     pub const DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT: Self = Self(1_000_011_000);
 }
 #[doc = "Generated from 'VK_EXT_debug_report'"]
-impl StructureType {
-    #[deprecated = "Backwards-compatible alias containing a typo"]
-    pub const DEBUG_REPORT_CREATE_INFO_EXT: Self = Self::DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT;
-}
-#[doc = "Generated from 'VK_EXT_debug_report'"]
 impl Result {
     pub const ERROR_VALIDATION_FAILED_EXT: Self = Self(-1_000_011_001);
 }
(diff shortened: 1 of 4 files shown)


ash/src/vk/extensions.rs
```diff
@@ -4481,7 +4481,7 @@ impl PipelineCreateFlags {
 #[doc = "Generated from 'VK_KHR_dynamic_rendering'"]
 impl PipelineCreateFlags {
     #[deprecated = "Backwards-compatible alias containing a typo"]
-    pub const RASTERIZATION_STATE_FRAGMENT_SHADING_RATE_ATTACHMENT_KHR: Self =
+    pub const PIPELINE_RASTERIZATION_STATE_CREATE_FRAGMENT_SHADING_RATE_ATTACHMENT_KHR: Self =
```
Ralith, line 4484: Why is the name changing as well?

Ralith (commented):
Dropping the backwards-compatibility aliases entirely sounds good to me.

ash/src/vk/extensions.rs
MarijnS95: Because I came up with this "more correct" prefix and added explicit fallback names back when these were the only enums in the source (before being turned into fallback-aliases), instead of throwing them in `is_enum_variant_with_typo` and calling it a day.

MarijnS95:
> Dropping the backwards-compatibility aliases entirely sounds good to me.

They also don't make much sense anyway with - just the `VK_` prefix stripped from them. Shall we do that as part of this PR immediately already?

Ralith:
If you're up for it, I see no reason to delay!

generator/src/lib.rs
```diff
@@ -1080,7 +1092,11 @@ impl<'a> ConstantExt for ExtensionConstant<'a> {
         self.constant.clone()
     }
     fn variant_ident(&self, enum_name: &str) -> Ident {
-        variant_ident(enum_name, self.name)
+        variant_ident(
+            enum_name,
+            self.name,
+            self.notation == Some("Backwards-compatible alias containing a typo"),
```
filnet, line 1098: The string constant appears four times.
A Java coder would factorize it into a private static constant.

MarijnS95: Fortunately I'm completely oblivious to Java (coders), nor do I think anyone cares for a generator that's on the brink of being rewritten.

MarijnS95:
@Ralith Done!

Ralith (approved):
Very tidy!

PR #505, opened 2021-12-01
ash-window: Upgrade to raw-window-handle 0.4.2

MarijnS95:
Note that this depends on one of the many winit 0.25 example upgrades (though we'll require 0.26 to make `raw-window-handle` 0.4 work) in order to succeed compilation on the examples.

---

Not sure about the timing on this one: raw-window-handle just came out, and winit 0.26 too.  There's a compatibility in 0.3.4 that allows consumers like this ash-window crate to consume the 0.4 handle (because the trait is implemented for it) produced by ie. winit, so there's a good reason to stay on 0.3.4 for longer to have wider adoption before migrating the ash-window crate - which is unfortunately tied in to the ash version!  This just provides a git branch/rev for crate authors to use if they desire to deduplicate their raw-window-handle dependency in the crate graph.

I wonder if we can use a lower and upper version bound on ash and just perform minor releases of ash-window to increase that upper bound whenever a new, compatible version of ash is released.  A bit like the [proposed fixes for vma](https://github.com/gwihlidal/vk-mem-rs/pull/57/files).  We can keep two ash-window major releases in parallel, but that's going to get hairy as soon as there's a breaking change in ash (in fact the unreleased "Support linking vulkan directly" PR changes our API surface already).

---

The match arms are not guarded by `cfg` anymore, allowing us to compile-test these simple arms on every system whenever our Ash extension helpers and types are not guarded by `cfg` attributes either.  This applies to every platform except Mac/IOS where the symbols and external raw-window-metal crate are themselves guarded by cfg's.

CC @msiglreith, seems like we need an update to https://github.com/norse-rs/raw-window-metal before this could even work :)

### DIFF
--- a/ash-window/Cargo.toml
+++ b/ash-window/Cargo.toml
@@ -16,10 +16,10 @@ rust-version = "1.59.0"
 
 [dependencies]
 ash = { path = "../ash", version = "0.37", default-features = false }
-raw-window-handle = "0.3.4"
+raw-window-handle = "0.4.2"
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-raw-window-metal = "0.1"
+raw-window-metal = "0.2"
 
 [dev-dependencies]
 winit = "0.26"
--- a/ash-window/Changelog.md
+++ b/ash-window/Changelog.md
@@ -2,7 +2,13 @@
 
 ## [Unreleased] - ReleaseDate
 
+### Changed
+
+- Bumped `raw-window-handle` to `0.4.2` (#505)
+
 ## [0.10.0] - 2022-03-23
+
+### Changed
 
 - Bumped `ash` version to [`0.37`](https://github.com/MaikKlein/ash/releases/tag/0.37.0) (#600)
 - Make `enumerate_required_extensions()` return `&[*const c_char]` instead of `Vec<&CStr>` to match `ash::vk::InstanceCreateInfo` (#590)
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -2,11 +2,12 @@
 
 use std::os::raw::c_char;
 
-use ash::{extensions::khr, prelude::*, vk, Entry, Instance};
+use ash::{
+    extensions::{ext, khr},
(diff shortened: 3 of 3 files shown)


MarijnS95:
Thanks @msiglreith, seems we can update to https://github.com/norse-rs/raw-window-metal/pull/4 (v0.2) without any other changes (though the CI doesn't build-test a mac/ios build afaik).

dvc94ch:
What is this waiting on?

MarijnS95:
@dvc94ch As long as we stick on 0.3.4, this crate can be used with both the 0.3.4 and and 0.4.1 window handles. When we upgrade that fallback to 0.3.4 is gone, so it actually makes sense to just stay on 0.3.4 - until we start missing functionality.

dvc94ch:
Not quite sure what you mean, this is the error I get with the winit provided handle:

```
error[E0308]: mismatched types
   --> piet-gpu-hal/src/vulkan.rs:176:70
    |
176 |                 for ext in ash_window::enumerate_required_extensions(*handle)? {
    |                                                                      ^^^^^^^ expected trait `raw_window_handle::HasRawWindowHandle`, found trait `HasRawWindowHandle`
    |
    = note: expected reference `&dyn raw_window_handle::HasRawWindowHandle`
               found reference `&dyn HasRawWindowHandle`
    = note: perhaps two different versions of crate `raw_window_handle` are being used?

error[E0308]: mismatched types
   --> piet-gpu-hal/src/vulkan.rs:226:76
    |
226 |                     surface: ash_window::create_surface(&entry, &instance, handle, None)?,
    |                                                                            ^^^^^^ expected trait `raw_window_handle::HasRawWindowHandle`, found trait `HasRawWindowHandle`
    |
    = note: expected reference `&dyn raw_window_handle::HasRawWindowHandle`
               found reference `&dyn HasRawWindowHandle`
    = note: perhaps two different versions of crate `raw_window_handle` are being used?
```

MarijnS95:
@dvc94ch I think https://github.com/MaikKlein/ash/pull/553 may be of interest to you.

Ralith (approved):

PR #506, opened 2021-12-05
Introduce ash-swapchain helper crate

Ralith:
This implements a reusable, lightly-opinionated helper for typical resizable windowed swapchain operation, which is otherwise error-prone.


ash-swapchain/examples/demo.rs
```diff
@@ -0,0 +1,310 @@
...
+            self.device.end_command_buffer(cmd).unwrap();
+            self.device
+                .queue_submit(
+                    self.queue,
+                    &[vk::SubmitInfo::builder()
+                        .wait_semaphores(&[acq.ready])
+                        .wait_dst_stage_mask(&[vk::PipelineStageFlags::TRANSFER])
+                        .signal_semaphores(&[self.frames[acq.frame_index].complete])
+                        .command_buffers(&[cmd])
+                        .build()],
```
MarijnS95, line 275: When needing a slice of a single element that is created by a builder, I typically use `std::slice::as_ref()` which allows the builder to `Deref` without calling `.build()`.

Ralith: That seems like extraneous ceremony to me, since this `build` call is lexically within the statement that consumes it and therefore sound regardless.

ash-swapchain/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    /// Dimensions of images in [`images`](Self::images)
+    pub fn extent(&self) -> vk::Extent2D {
+        self.extent
+    }
+
+    /// Acquire a swapchain image
+    ///
+    /// Returns [`vk::Result::ERROR_OUT_OF_DATE_KHR`] iff the configured format or present modes
```
aloucks, line 115:
```suggestion
    /// Returns [`vk::Result::ERROR_OUT_OF_DATE_KHR`] if the configured format or present modes
```

Ralith: `iff` is deliberately used in the technical sense here. Did I miss a case?

aloucks: I think you should spell out `if-and-only-if` in this case. It looks like a typo.

Ralith:
I'm now confident the resize handling here is as well-behaved as the presentation API permits. Expanded the doc comments a good bit too.

ash-swapchain/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+    ///
+    /// # Safety
+    ///
+    /// `surface_fn` and `swapchain_fn` must have been acquired from `device`, which must match the
+    /// `device` passed to [`new`](Self::new)
+    pub unsafe fn acquire(
+        &mut self,
+        device: &Device,
+        surface_fn: &khr::Surface,
+        swapchain_fn: &khr::Swapchain,
```
aloucks, line 130: I think it would better to allow the user to provide a timeout (or perhaps an optional timeout and default to `!0` if not provided).

Ralith: What's the use case? Handling every possible requirement is a non-goal here, since people doing something unusual can always fall back on bare Vulkan.

aloucks: Dropping a frame if one can't be acquired in time or performing other work while waiting. It seems like a pretty simple addition that doesn't complicate this API.

aloucks: Also, the PR is self described as 'lightly opinionated' and not having a timeout feels 'strongly opinionated' (imo).

Ralith: > Dropping a frame if one can't be acquired in time

Ah, fair enough.

> It seems like a pretty simple addition that doesn't complicate this API.

Agreed, given that it provides some value outside of extremely niche cases. Applied.

MaikKlein:
A few high level comments:

This would need a readme that quickly explains the purpose of this library (The why and how).  Similar to what ash-window has. We should probably also promote our helper crates in the main readme

On another note, should we use this in the ash examples as well?

Ralith:
Added a brief README.

> On another note, should we use this in the ash examples as well?

I'm kinda tempted to drop ash's current example entirely in favor of this demo.rs; if we don't try to teach Vulkan in general and focus on maximal simplicity, the (currently empirically unsustainable) maintenance effort required to keep the example in good shape will be much reduced.

Ralith:
Replaced the `AcquiredFramge::generation: u64` with a `AcquiredFrame::invalidate_images: bool` to make things a little less opinionated.

ash-swapchain/examples/demo.rs
```diff
@@ -0,0 +1,343 @@
...
+                        .get_physical_device_queue_family_properties(dev)
+                        .into_iter()
+                        .enumerate()
+                        .find(|(_index, info)| {
+                            info.queue_flags.contains(vk::QueueFlags::GRAPHICS)
+                        })?;
+                    let family = family as u32;
+                    let supported = surface_fn
+                        .get_physical_device_surface_support(dev, family, surface)
+                        .unwrap();
```
theHamsta, line 93: this should be `.ok()?` instead of `.unwrap()`. Maybe `new` should also return a `Result`. There are a lot of `unwrap`s in this function.

ash-swapchain/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+                            .present_mode_preference
+                            .iter()
+                            .position(|&pref| pref == mode)?,
+                    ))
+                })
+                .min_by_key(|&(_, priority)| priority)
+                .map(|(mode, _)| mode)
+                .ok_or(vk::Result::ERROR_OUT_OF_DATE_KHR)?;
+
+            let desired_image_count =
```
Neo-Zhixing, line 217: Why `min_image_count + 1`?

ash-swapchain/src/lib.rs
```diff
@@ -0,0 +1,415 @@
...
+                            invalidate_images,
+                        });
+                    }
+                    Err(vk::Result::ERROR_OUT_OF_DATE_KHR) => {}
+                    Err(e) => return Err(e),
+                }
+            };
+            self.needs_rebuild = true;
+
+            // Rebuild swapchain
```
Neo-Zhixing, line 179: Might be a good idea to split this out into a separate method too

ash-swapchain/src/lib.rs
Ralith: This is standard practice to reduce sync coupling with the WSI. See also e.g. https://github.com/charles-lunarg/vk-bootstrap/blob/master/src/VkBootstrap.cpp#L1763.

Ralith:
This should be heavily revised to take advantage of [VK_EXT_swapchain_maintenance1](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_swapchain_maintenance1.html), which resolves some key ambiguities in the base extension.

Ralith:
I think this should draw inspiration from https://github.com/ash-rs/ash/pull/1019 to load and cache the function pointer tables internally rather than requiring the user to awkwardly pass them in. Storing some redundant function pointers isn't a big deal.

> [VK_EXT_swapchain_maintenance1](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_swapchain_maintenance1.html)

AFAIK this is *still* not implemented by all major vendors on desktop, so maybe not a good target :(

PR #507, opened 2021-12-06
Fix new clippy warnings since Rust 1.57

MarijnS95:
The removal of `handle` from `DrawIndirectCount` will slightly interfere with #494, but it remains to be seen which PR lands first.


Ralith (approved):
I vote for merging the other PR first, to reduce the toil placed on a more casual contributor.

MarijnS95:
@Ralith Same thought here, this is easily enough and we've already pushed #494 through a lot of changes.

PR #512, opened 2021-12-19
Simplify triangle example's vertex input state

swooster:
Use a `PipelineVertexInputStateCreateInfo` builder to pass a slice instead of a pointer and length. Also makes the triangle example a little more consistent with the texture example.

### DIFF
--- a/examples/src/bin/triangle.rs
+++ b/examples/src/bin/triangle.rs
@@ -263,13 +263,9 @@ fn main() {
             },
         ];
 
-        let vertex_input_state_info = vk::PipelineVertexInputStateCreateInfo {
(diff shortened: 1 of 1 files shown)


MarijnS95 (approved):
Nice catch! I'm quite surprised though why the original commit, that said "to use builders and defaults", removed this exact use of the builder pattern to substitute it with that manual assignment... https://github.com/MaikKlein/ash/commit/0a16b5bd9b8bdece2cb43f139947b020c63fbd61#diff-99480e84b5476c223fd790be90ef853c9f4295d15113172f83679b2882ece4e6L259-R278

PR #514, opened 2021-12-20
Update Vulkan-Headers to 1.2.203

MarijnS95:


ash/src/vk/native.rs
```diff
@@ -99,6 +99,8 @@ pub const StdVideoH264ChromaFormatIdc_STD_VIDEO_H264_CHROMA_FORMAT_IDC_444:
     StdVideoH264ChromaFormatIdc = 3;
 pub const StdVideoH264ChromaFormatIdc_STD_VIDEO_H264_CHROMA_FORMAT_IDC_INVALID:
     StdVideoH264ChromaFormatIdc = 2147483647;
+pub const StdVideoH264ChromaFormatIdc_STD_VIDEO_H264_CHROMA_FORMAT_IDC_MAX_ENUM:
+    StdVideoH264ChromaFormatIdc = 2147483647;
```
Ralith, line 103: Should we be emitting MAX_ENUM values?

MarijnS95: Probably not, but I'm also not sure if we should go to the length of removing these from `bindgen` output. The regex for this should be just `\w+_MAX_ENUM` in `blocklist-item`, I'll give it a go.

Ralith: Ah, didn't realize this was the bindgen output! No worries if it's a pain.

MarijnS95: I thought this was trivial with `.blocklist_item(r"\w+_MAX_ENUM")` but apparently (at least back in 2019, haven't found anything recent) this requires a custom callback: https://github.com/rust-lang/rust-bindgen/issues/1526

Bit ugly, but it works.

Ralith (approved):

generator/src/lib.rs
```diff
@@ -2877,7 +2877,25 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
...
+            &self,
+            _enum_name: Option<&str>,
+            original_variant_name: &str,
+            _variant_value: bindgen::callbacks::EnumVariantValue,
+        ) -> Option<bindgen::callbacks::EnumVariantCustomBehavior> {
+            if original_variant_name.ends_with("_MAX_ENUM") {
+                Some(bindgen::callbacks::EnumVariantCustomBehavior::Hide)
+            } else {
+                None
+            }
```
MarijnS95, line 2893: I just realized, this could be turned into `.then(|| bindgen::callbacks::EnumVariantCustomBehavior::Hide)` :hand_over_mouth: 

Been told before that that's a bit too tricky for the average Rust dev though, leaving it out.

Ralith: I'm partial to cheeky combinators personally, but it's a matter of taste :grinning:

MarijnS95: Many loathe control flow hidden behind functions, functional style :D

PR #515, opened 2021-12-20
Prepare release notes for 0.34 (and ash-window 0.8) with Vulkan 1.2.203

MarijnS95:
Pending #514

CC @filnet

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,27 +4,43 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased] - ReleaseDate
+## [0.34.0] - 2021-12-22
 
 ### Added
 
+- Update Vulkan-Headers to 1.2.203 (#477, #497, #504, #509, #514)
+- Add missing documentation to bitflag extension variants (#501)
 - Added `VK_KHR_present_wait` device extension (#493)
+- Added `VK_KHR_maintenance4` device extension (#489, #498)
+- Link `_len()` functions to their array-getter using intradoc-links (#490)
+- Added `VK_KHR_dynamic_rendering` device extension (#488)
 
 ### Changed
 
 - Extension names from `fn name()` will not be checked for interior nuls anymore at runtime (#522)
 - examples: Use `c_char` for pointer to raw string (#521)
+- Group enum extension variants together per `impl T` block (#519)
+- examples: Use `slice::from_ref` to not loose lifetime on nested slices (#513)
+- Simplify triangle example's vertex input state (#512)
 - Device extension `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` now expose `fn device()` instead of `fn instance()` (#499)
 - Changed `khr::PipelineExecutableProperties::new()` and `khr::TimelineSemaphore::new()` to take `instance` and `device` as arguments (#499)
+- Fix broken vulkan-tutorial link in README.md (#492)
+- Make `enumerate_physical_device_groups` unsafe (#491)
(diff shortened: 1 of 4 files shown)


Changelog.md
```diff
@@ -4,24 +4,44 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased] - ReleaseDate
+## [0.34.0] - ReleaseDate
```
MarijnS95, line 7: I'll bump this as soon as there's review-approval, and do the release straight after. Not sure if we can make it today (2021-12-20) but hopefully tomorrow at last.

Changelog.md
```diff
@@ -4,24 +4,44 @@ All notable changes to this project will be documented in this file.
...
 ### Changed
 
+- examples: Use `slice::from_ref` to not loose lifetime on nested slices (#513)
+- Simplify triangle example's vertex input state (#512)
 - Device extension `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` now expose `fn device()` instead of `fn instance()` (#499)
 - Changed `khr::PipelineExecutableProperties::new()` and `khr::TimelineSemaphore::new()` to take `instance` and `device` as arguments (#499)
+- Fix broken vulkan-tutorial link in README.md (#492)
+- Make `enumerate_physical_device_groups` unsafe (#491)
+- Added `Packed24_8` helper-type for constructing AS Instance bitfields, used in `AccelerationStructureInstanceKHR`, `AccelerationStructureSRTMotionInstanceNV` and `AccelerationStructureMatrixMotionInstanceNV` (#476)s (#490)
+- Link Vulkan directly, statically, by default. Enable the `loaded` feature to get old `libloading` behaviour (#457)
```
Ralith, line 27:
```suggestion
- Link Vulkan directly under the default `linked` feature. Disable default features and enable the `loaded` feature for the old `libloading` behaviour (#457)
```
On most platforms, we're *dynamically* linking (not to be confused with dynamically *loading*), requiring the OS to come up with a vulkan-1.dll/libvulkan.so/whatever at runtime.

Enabling the `loaded` feature alone is insufficient, since that leaves the `linked` feature enabled.

Changelog.md
```diff
@@ -4,24 +4,44 @@ All notable changes to this project will be documented in this file.
...
+- Removed `device` argument from `ext::DebugMarkers::debug_marker_set_object_name` function, `khr::PipelineExecutableProperties` and `khr::TimelineSemaphore` functions (#494, #499)
 - Removed `From<vk::Result>` trait for `VkResult` (#495)
 - Removed `instance` argument from `ext::DebugUtils::submit_debug_utils_message` function (#499)
+- Removed misleading `all()`/`-`/`-=` function/ops from bitflags (#478)
+
+### Internal
+
+Changes inside Ash (ie. the `generator`) that should not affect our public API:
+
+- generator: Use `"Backwards-compatible"` comment to omit deprecated aliases (#502)
```
Ralith, line 42: Isn't the removal of deprecated aliases a public change?

Ralith (changes requested):
Thanks for driving tihs!

Changelog.md
MarijnS95: Fair enough, we're not linking _statically_ and that doesn't make any sense.

> Thanks for driving this!

Not sure if I should be driving in the literal sense at this point... :hot_face:

Changelog.md
MarijnS95: I recall forgetting to update that PR title now, as it was indeed decided during that PR to just drop deprecated enum variants while at it instead of coming up with suboptimal and useless names.

Moved it up to the `Removed` section, and reworded in a way that end-users know how to deal with the now-removed variants (though I do hope nobody was actively using and perhaps allowing these `#[deprecated]` constants, except of course the ones that only got their missing docs added in #501).

Ralith (approved):

Changelog.md
MarijnS95: Dropped this entire section and moved the `Self` change - which is much larger than the title suggests again :disappointed: - to `### Changed`.

Ralith:
Should we get https://github.com/MaikKlein/ash/pull/482 in as well?

Ralith (approved):

PR #519, opened 2021-12-20
extensions: Group enum extension variants together per `impl T` block

MarijnS95:
Instead of generating an impl block - together with "Generated from XXX" doc - for every single item that extends an enum type, group all these extensions together in a single `impl` block per extension per type.

This cuts down a couple thousand lines of repeated `impl T` and `#[doc]` annotations, and makes the code more readable at the same time: it is now possible to immediately see exactly all the constants that a certain extension extends a type with.


Ralith (approved):
Nice!

PR #523, opened 2021-12-22
ci: Run clippy once more without default features

MarijnS95:
With more and more features being added to `ash`, now seems to be the right time to make sure the crate is clean of clippy warnings when building without any features in addition to building with the deafult set of features.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
           profile: minimal
           toolchain: stable
           override: true
-      - run: rustup component add rustfmt
+          components: rustfmt
       - uses: actions-rs/cargo@v1
         with:
           command: fmt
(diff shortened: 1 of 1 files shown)


Ralith (commented):
Does `--workspace --no-default-features` actually work? I recall cargo being prone to silently ignoring feature-related flags in workspace-scoped operations.

MarijnS95:
Fair enough, it does not even seem to catch compiler errors :scream: 

Currently on master `cargo clippy --all --all-targets --no-default-features` succeeds, but `cargo clippy -p ash --all-targets --no-default-features` fails...

MarijnS95:
Ah, it's obvious of course: those crates such as `ash-window` or the examples (used with `--workspace`/`--all`) select the default features again.

Ralith (approved):
Needs a rebase (my bad) but LGTM otherwise

PR #526, opened 2021-12-23
Don't link Vulkan by default

### LINKED ISSUE #525: Don't require explicitly linking libvulkan by default
> Yes, the build environment will need to have a libvulkan for the target available. IMO it's a reasonable default. For one thing, the linker error you get for a missing foreign library is highly recognizable and easy to google, with the same solution across all programming languages, whereas a build error due to not having default features would be more idiosyncratic.

I'll resurrect this thread now that ash 0.34 is out and we're starting to see the first linker errors across our crates. I think this may have been a mistake. Rust users aren't used to fixing (or seeing) linker errors, and I honestly think this is a bad precedent to set from a usability point of view.

_Originally posted by @Jasper-Bekkers in https://github.com/MaikKlein/ash/issues/457#issuecomment-999941199_

Ralith:
This was found to be unreasonably disruptive to downstream CI configurations. Fixes #525.

I opted to additionally enable `loaded` by default, and reposition `Entry::load` above `Entry::linked`, so that a working configuration is available and easily found out of the box.

### DIFF
--- a/ash-window/Cargo.toml
+++ b/ash-window/Cargo.toml
@@ -22,6 +22,7 @@ raw-window-metal = "0.1"
 
 [dev-dependencies]
 winit = "0.19.4"
+ash = { path = "../ash", version = "0.34", default-features = false, features = ["linked"] }
 
 [[example]]
 name = "winit"
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -29,39 +29,11 @@ pub struct Entry {
 /// Vulkan core 1.0
 #[allow(non_camel_case_types)]
 impl Entry {
-    /// Load entry points from a Vulkan loader linked at compile time
-    ///
-    /// Note that instance/device functions are still fetched via `vkGetInstanceProcAddr` and
-    /// `vkGetDeviceProcAddr` for maximum performance.
-    ///
-    /// ```no_run
-    /// use ash::{vk, Entry};
-    /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
-    /// let entry = Entry::new();
-    /// let app_info = vk::ApplicationInfo {
-    ///     api_version: vk::make_api_version(0, 1, 0, 0),
-    ///     ..Default::default()
-    /// };
-    /// let create_info = vk::InstanceCreateInfo {
-    ///     p_application_info: &app_info,
-    ///     ..Default::default()
-    /// };
-    /// let instance = unsafe { entry.create_instance(&create_info, None)? };
-    /// # Ok(()) }
-    /// ```
-    #[cfg(feature = "linked")]
-    #[cfg_attr(docsrs, doc(cfg(feature = "linked")))]
-    pub fn new() -> Self {
-        // Sound because we're linking to Vulkan, which provides a vkGetInstanceProcAddr that has
-        // defined behavior in this use.
-        unsafe {
-            Self::from_static_fn(vk::StaticFn {
-                get_instance_proc_addr: vkGetInstanceProcAddr,
-            })
-        }
-    }
-
     /// Load default Vulkan library for the current platform
+    ///
+    /// Prefer this over [`linked`](Self::linked) when your application can gracefully handle
+    /// environments that lack Vulkan support, and when the build environment might not have Vulkan
+    /// development packages installed (e.g. the Vulkan SDK, or Ubuntu's libvulkan-dev).
     ///
     /// # Safety
     /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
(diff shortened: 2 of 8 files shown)


ash/src/entry.rs
```diff
@@ -29,40 +29,10 @@ pub struct Entry {
...
-        unsafe {
-            Self::from_static_fn(vk::StaticFn {
-                get_instance_proc_addr: vkGetInstanceProcAddr,
-            })
-        }
-    }
-
     /// Load default Vulkan library for the current platform
     ///
+    /// Prefer this over [`linked`](Self::linked) when Vulkan support is not required at runtime.
```
MarijnS95, line 34: Vulkan support is always required... _linking Vulkan at runtime_ however, is not :)

You may want to mention though that using `linked` needs the aptly named feature to be enabled.

ash/src/entry.rs
```diff
@@ -103,6 +73,43 @@ impl Entry {
...
+    /// let create_info = vk::InstanceCreateInfo {
+    ///     p_application_info: &app_info,
+    ///     ..Default::default()
+    /// };
+    /// let instance = unsafe { entry.create_instance(&create_info, None)? };
+    /// # Ok(()) }
+    /// ```
+    #[cfg(feature = "linked")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "linked")))]
+    pub fn linked() -> Self {
```
MarijnS95, line 103: That's a much needed name-change :+1:

ash/src/entry.rs
Ralith: That wording was intended; tweaked it to be clearer. The `#[cfg_attr(docsrs, doc(cfg(feature = "loaded")))]` should take care of the latter part.

MarijnS95: Much better, thanks!

ash/src/entry.rs
```diff
@@ -355,7 +365,7 @@ mod loaded {
 
     impl From<MissingEntryPoint> for LoadingError {
         fn from(err: MissingEntryPoint) -> Self {
-            LoadingError::MissingEntryPoint(err)
+            Self::MissingEntryPoint(err)
```
MarijnS95, line 368: Clippy woke up / started working all of a sudden. I don't understand why it's broken most of the time :/

Ralith: I think this wasn't getting coverage because the `loaded` feature wasn't enabled for clippy.

Ralith: https://github.com/MaikKlein/ash/pull/527

examples/Cargo.toml
```diff
@@ -7,5 +7,5 @@ edition = "2018"
 [dependencies]
 winit = "0.25.0"
 image = "0.10.4"
-ash = { path = "../ash" }
+ash = { path = "../ash", default-features = false, features = ["linked", "debug"] }
```
aloucks, line 10: It might be helpful to have some explanation as to why the examples are using `linked` instead `loaded`. I assume this is because the examples are requiring the validation layer from the SDK. If that's not the case, the examples should probably use `loaded` to minimize friction.

MarijnS95:
Does the CI test job still need to install `libvulkan-dev` now?

Ralith:
Yes, to allow tests/examples to be built with the `linked` feature enabled.

examples/Cargo.toml
Ralith: Added a note.

ash/src/entry.rs
```diff
@@ -29,40 +29,12 @@ pub struct Entry {
...
-                get_instance_proc_addr: vkGetInstanceProcAddr,
-            })
-        }
-    }
-
     /// Load default Vulkan library for the current platform
     ///
+    /// Prefer this over [`linked`](Self::linked) when your application can gracefully handle
+    /// environments that lack Vulkan support, and when the build environment might not have the
+    /// Vulkan loader available.
```
aloucks, line 36: It might be helpful to rephrase this in such a way that emphasizes the impact on downstream dependencies. 

`Entry::loaded` should be preferred unless you explicitly want to impose a dependency on the Vulkan SDK (e.g. perhaps you depend on validation layers). For example, `bevy` depends on `wgpu`. If `wgpu` decided to used `Entry::linked`, every developer who attempted to experiment with bevy would need to install the Vulkan SDK or face compile time linker errors.

Ralith: Adjusted the wording to make the build environment requirements clearer.

MaikKlein (approved):

Jasper-Bekkers (approved):
I think this reverts back to the old behavior so I can approve it. @MaikKlein I think it would make sense to push a new release with this fix as soon as it's merged. What do you think?

ash/src/entry.rs
```diff
@@ -103,6 +75,43 @@ impl Entry {
         Self::load_from(LIB_PATH)
     }
 
+    /// Load entry points from a Vulkan loader linked at compile time
+    ///
+    /// Compared to [`load`](Self::load), this is infallible, but requires that the build
+    /// environment have Vulkan development packages installed (e.g. the Vulkan SDK, or Ubuntu's
+    /// libvulkan-dev), and prevents the resulting binary from starting in environments that do not
+    /// support Vulkan.
```
MarijnS95, line 83: `s/support Vulkan/ship with Vulkan libraries`?

ash/src/entry.rs
```diff
@@ -103,6 +75,43 @@ impl Entry {
         Self::load_from(LIB_PATH)
     }
 
+    /// Load entry points from a Vulkan loader linked at compile time
+    ///
+    /// Compared to [`load`](Self::load), this is infallible, but requires that the build
+    /// environment have Vulkan development packages installed (e.g. the Vulkan SDK, or Ubuntu's
+    /// libvulkan-dev), and prevents the resulting binary from starting in environments that do not
```
MarijnS95, line 82:
```suggestion
    /// `libvulkan-dev`), and prevents the resulting binary from starting in environments that do not
```

ash/src/entry.rs
```diff
@@ -29,40 +29,12 @@ pub struct Entry {
...
-                get_instance_proc_addr: vkGetInstanceProcAddr,
-            })
-        }
-    }
-
     /// Load default Vulkan library for the current platform
     ///
+    /// Prefer this over [`linked`](Self::linked) when your application can gracefully handle
+    /// environments that lack Vulkan support, and when the build environment might not have Vulkan
+    /// development packages installed (e.g. the Vulkan SDK, or Ubuntu's libvulkan-dev).
```
MarijnS95, line 36:
```suggestion
    /// development packages installed (e.g. the Vulkan SDK, or Ubuntu's `libvulkan-dev`).
```

examples/Cargo.toml
MarijnS95: @aloucks The note is nice, but keep in mind that dynamic linking or (dynamic) runtime loading has no impact on being able to use the validation layers. It "just so happens" that both the `.lib` (needed for dynamic linking on Windows) (on Linux the runtime `.so` is used for this) and validation layers ship with the SDK.

FWIW I thought the examples would use `linked` to both showcase this feature, and because examples are generally ran on the same machine (ie. build environment == runtime environment) so it makes sense that the user intends to develop Vulkan applications, hence has the ability to both "develop" (with the SDK) and run them.

ash-window/Cargo.toml
```diff
@@ -22,6 +22,7 @@ raw-window-metal = "0.1"
 
 [dev-dependencies]
 winit = "0.19.4"
+ash = { path = "../ash", version = "0.34", default-features = false, features = ["linked"] }
```
MarijnS95, line 25: That's a nice way to satisfy `required-features = ["ash/linked"]` :)

MarijnS95 (approved):

PR #527, opened 2021-12-23
Cover all features with clippy

Ralith:
Experimentally verified that this works as intended.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,4 +96,8 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --workspace --all-targets -- -D warnings
+          args: -p ash --no-default-features -- -D warnings
+      - uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --workspace --all-targets --all-features -- -D warnings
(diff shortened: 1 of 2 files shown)


MarijnS95:
Should we also do a run next to it with `--no-default-features`, and, if feasible, _also_ keep the default features? That sounds like the three most common configurations for the time being.

Ralith:
Given that our features are independent, I don't think there's benefit to running clippy under any other configurations; we just want to be sure it sees all the code, and `--all-features` is sufficient for that. We might want to have `--no-default-features` coverage for other CI stages, but that can be pursued independently.

MarijnS95:
Features are (supposed to be) additive, so it is indeed true that all code should be seen with `--all-features`. However, since that's not the default I'd also like the other extreme of no features at all to be tested, especially given "widespread" guards for the `"debug"` feature. That way we'll catch an accidental unconstrained `derive(Debug)` being added, for example.

Of course that won't find `linked` code accidentally referencing `loaded` code or vice-versa, but the chances for that seem way slimmer. I'd be fine to just have the `--no-default-features` stage under `clippy` too.
The `check` step is a bit useless as `clippy` performs that as a base pass too (or 99% of it).

Ralith:
Added a clippy no-features case.

MarijnS95:
Would have been fine for me to keep it under the same `steps` - should save another VM from being spun unnecessarily? I'd say it could reuse built dependencies but we don't have much of those.

Ralith:
Folded them together.

MarijnS95 (approved):

PR #529, opened 2021-12-27
Release ash 0.35 and ash-window 0.9 without default compile-time linking

MarijnS95:


ash-window/README.md
```diff
@@ -33,8 +33,7 @@ The library exposes two functions:
 
 ## Versions
 ```toml
-ash = "0.33"
-raw-window-handle = "0.3"
+ash = "0.35"
 ```
```
MarijnS95, line 37: @msiglreith I'm not really sure about listing the `raw-window-handle` version here. Theoretically end-users don't need to import this (unless they want to use the trait bound in their own crate) as it should only be provided/implemented by their windowing provider crate, like `winit`.

Otherwise, `0.3.4` now also implements its `HasRawWindowHandle` trait for the `RawWindowHandle` from `0.4` meaning that a `0.3` consumer like `ash-window` can accept the `RawWindowHandle` enum from both versions.

msiglreith: Listing it there got less important given the backwards compatibility of the latest `raw-window-handle` to ensure that versions match with end-users, I agree. The motivation for this section was mostly to show compatibility so nowadays probably `raw-window-handle = "0.3, 0.4"`

Ralith (approved):

ash-window/README.md
Ralith: My inclination would be to drop the unnecessary direct dep entirely to avoid confusion about what's required.

MaikKlein (approved):

ash-window/README.md
MarijnS95: Fair enough indeed. Dropped it yet again, and making a release now.

PR #530, opened 2021-12-27
extensions: Add VK_KHR_get_surface_capabilities2

asnordstrand:


ash/src/extensions/khr/get_surface_capabilities2.rs
```diff
@@ -0,0 +1,55 @@
...
+            .result_with_success(surface_capabilities)
+    }
+
+    #[doc = "https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPhysicalDeviceSurfaceFormats2KHR.html"]
+    pub unsafe fn get_physical_device_surface_formats2(
+        &self,
+        physical_device: vk::PhysicalDevice,
+        surface_info: &vk::PhysicalDeviceSurfaceInfo2KHR,
+    ) -> VkResult<Vec<vk::SurfaceFormat2KHR>> {
+        read_into_uninitialized_vector(|count, data| {
```
MarijnS95, line 42: `vk::SurfaceFormat2KHR` has an `s_type` and `p_next` (unlike `vk::SurfaceFormatKHR`) hence needs to be initialized with `read_into_defaulted_vector` here. Perhaps this `p_next` should be settable externally just like other call-sites that have a separate `_len()` function (see for example `fn get_device_image_sparse_memory_requirements` in `maintenance4`).

MarijnS95 (commented):
Nice! Good use of `Default::default()` for `vk::SurfaceCapabilities2KHR`, but I think you need to do the same for `vk::SurfaceFormat2KHR`.

I don't think `rustfmt` is going to be happy with `fp()` and `name()` on a single line though?

ash/src/extensions/khr/get_surface_capabilities2.rs
asnordstrand: Good points that I overlooked. Have made the suggested changes, let me know if I'm handling the errors correctly.

MarijnS95 (approved):
Looks totally right to me now!

MarijnS95:
By the way, don't forget to add a note to `./Changelog.md`, see the other new extensions in there for examples.

(We should setup a PR template listing these necessary items later)

PR #534, opened 2021-12-27
Assert that Vulkan array-getters return the same length

MarijnS95:
Originally introduced in [#489] this inserts the array-length equality check everywhere else: in the supposedly invalid and inexistant event where Vulkan suddenly returns less items (`count` has been modified) than were originally queried through respective `_len()` functions (or more likely: a slice of invalid length passed by the user) and some elements at the end of the slice are left uninitialized, panic.

Wherever there is valid concern or possibility for this to happen - or to circumvent assertions and panics altogether - mutable references to mutable slices should be passed allowing the length to be promptly updated.

[#489]: https://github.com/MaikKlein/ash/pull/489#discussion_r753020089

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -328,6 +328,7 @@ impl Device {
             &mut count,
             out.as_mut_ptr(),
         );
+        assert_eq!(count as usize, out.len());
     }
 
     #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkTrimCommandPool.html>"]
--- a/ash/src/extensions/khr/get_memory_requirements2.rs
+++ b/ash/src/extensions/khr/get_memory_requirements2.rs
@@ -70,6 +70,7 @@ impl GetMemoryRequirements2 {
             &mut count,
             out.as_mut_ptr(),
         );
+        assert_eq!(count as usize, out.len());
     }
 
     pub fn name() -> &'static CStr {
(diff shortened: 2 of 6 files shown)


ash/src/device.rs
```diff
@@ -328,6 +328,7 @@ impl Device {
             &mut count,
             out.as_mut_ptr(),
         );
+        assert_eq!(count, out.len() as u32);
```
filnet, line 331: could use `as _` ?

Ralith (approved):
Built-in sanity-check asserts feels like it might be a bit opinionated compared to ash's status quo, but none of this is hotpath stuff anyway.

ash/src/device.rs
MarijnS95: What's the advantage of not being explicit about this (trivial to write) type?

That said, it's probably better to cast `count` to `usize` to prevent a possible truncation.

filnet: I am always wondering wether to use `as _` or be explicit...

MarijnS95: @filnet There are probably no real API guidelines but it's probably better to be explicit unless intentionally hiding long type casts.

I've also turned on the `trivial-casts` and `trivial-numereric-casts` default-`allow` `rustc` warning locally to notice that there are quite a few casts to the same type, mostly `as _` which was copied verbatim without the programmer really knowing or caring about the source and target type.

In either case we should aim to flip that warning on and make the code ever so slightly more readable by removing stray `as` casts.

> That said, it's probably better to cast `count` to `usize` to prevent a possible truncation.

On second thought I did this by intention to match the truncation from `let mut count = out.len() as u32;`. However, that means - if someone manages to pass a slice longer than `u32::max` at all - this assert won't fail even if it should. Erring on the side of correctness, this is now flipped.

Ralith (approved):

PR #537, opened 2021-12-28
Generate `RGBA=R|G|B|A` helper constant for `ColorComponentFlags`

### LINKED ISSUE #536: Helper function for rgba `vk::ColorComponentFlags`
with the removal of `vk::ColorComponentFlags::all()`, I tried to use `!vk::ColorComponentFlags::empty()`. This causes a vulkan validation error:

`
VUID-VkPipelineColorBlendAttachmentState-colorWriteMask-parameter(ERROR / SPEC): msgNum: -1191591886 - Validation 
Error: [ VUID-VkPipelineColorBlendAttachmentState-colorWriteMask-parameter ] Object 0: handle = 0x1e7895b3b98, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xb8f9c032 | vkCreateGraphicsPipelines: value of pCreateInfos[9].pColorBlendState->pAttachments[0].colorWriteMask contains flag bits that are not recognized members of VkColorComponentFlagBits The Vulkan spec states: colorWriteMask must be a valid combination of VkColorComponentFlagBits values (https://vulkan.lunarg.com/doc/view/1.2.198.0/windows/1.2-extensions/vkspec.html#VUID-VkPipelineColorBlendAttachmentState-colorWriteMask-parameter)
`

It'd be great to have a `vk::ColorComponentFlags::rgba()` helper function for this.

MarijnS95:
Fixes #536

When the misleading `all()` function was removed in #478 it also made all color components for `ColorComponentFlags` significantly more verbose to write, see #536.

### DIFF
--- a/ash/src/vk/prelude.rs
+++ b/ash/src/vk/prelude.rs
@@ -26,3 +26,8 @@ impl Packed24_8 {
         (self.0 >> 24) as u8
     }
 }
+
+impl super::ColorComponentFlags {
+    /// Contraction of [`Self::R`] | [`Self::G`] | [`Self::B`] | [`Self::A`]
+    pub const RGBA: Self = Self(Self::R.0 | Self::G.0 | Self::B.0 | Self::A.0);
+}
--- a/examples/src/bin/texture.rs
+++ b/examples/src/bin/texture.rs
@@ -666,10 +666,7 @@ fn main() {
             src_alpha_blend_factor: vk::BlendFactor::ZERO,
             dst_alpha_blend_factor: vk::BlendFactor::ZERO,
             alpha_blend_op: vk::BlendOp::ADD,
-            color_write_mask: vk::ColorComponentFlags::R
-                | vk::ColorComponentFlags::G
-                | vk::ColorComponentFlags::B
-                | vk::ColorComponentFlags::A,
+            color_write_mask: vk::ColorComponentFlags::RGBA,
         }];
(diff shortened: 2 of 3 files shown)


MarijnS95:
Not sure if this is the "right" way, perhaps it's cleaner to throw this as a manual `impl` in `src/vk/prelude.rs`. Keeps generator edgecases on a low, too.

Ralith:
I'd lean towards handwriting this. Happy to have it otherwise, though!

MarijnS95:
Much cleaner and simpler indeed :)

expenses:
Looks great! 💖

ash/src/vk/prelude.rs
```diff
@@ -26,3 +26,8 @@ impl Packed24_8 {
         (self.0 >> 24) as u8
     }
 }
+
+impl super::ColorComponentFlags {
+    /// Contraction of [`Self::R`] | [`Self::G`] | [`Self::B`] | [`Self::A`]
+    pub const RGBA: Self = Self(Self::R.0 | Self::G.0 | Self::B.0 | Self::A.0);
```
Ralith, line 32: The unwrap/rewrap dance isn't needed, is it?
```suggestion
    pub const RGBA: Self = Self::R | Self::G | Self::B| Self::A```

MarijnS95: `BitOr` implemented for `ColorComponentFlags` isn't `const` so I couldn't make that work. See compiler errors in first push, before force-push.

Issue #314, opened 2020-07-12
Support for no-std


nifker:
Can this crate be changed in a way that it supports no-std(as Vulkan explicitly supports embedded devices) or does it depend on std exclusive stuff?

nifker:
Instead of Vec<T> we could use the crate 'smallvec' for no_std.

Ralith:
`Vec` is available via the `alloc` crate in no_std. You could cfg out the high level wrappers if you want no-alloc support as well.

MaikKlein:
I haven't looked at no_std at all yet. `ash` also fetches all the function pointers at runtime, is that a problem for embedded devices?

nifker:
Ok it seems like dynamic function loading is not trivial on no_std.

Ralith:
Dynamic function loading is intrinsic to Vulkan, minimal entry points aside.

MaikKlein:
Which "embedded devices" do you want to run vulkan on?

nifker:
Dont have anything explicit in mind but I think it would be some device with an RTOS or with a statically linked vulkan driver.

hyarsan:
Also, `CStr` is only in libstd. Hopefully Rust gets better null-terminated strings eventually...

i509VCB:
I do have some interest in this, there are few things I can think of to make moving towards being no_std capable (probably with alloc given we need dynamic function loading):

- Make the generator generate types using `core` and `alloc` instead of `std` for imports where possible. This is a change that can be made without any api changes.
- Regarding CStr(ing) we either need to hope those are moved to liballoc/libc/libcore or make ash depend on https://crates.io/crates/cstr_core when no_std is used.

I imagine there is more than those two things involved with no_std support.

Ralith:
> probably with alloc given we need dynamic function loading

Loading functions does not require alloc.

i509VCB:
> > probably with alloc given we need dynamic function loading
> 
> Loading functions does not require alloc.

Well I guess I am mistaken there.

I could make some pull requests to do some of the type import stuff for core and alloc if desired

i509VCB:
To help with this, I have created the following clippy lint: https://github.com/rust-lang/rust-clippy/pull/9103

The lint should help us ensure we don't forget any types that could be specified in core or alloc instead.

i509VCB:
[...]

PR #538, opened 2021-12-28
use core and alloc imports where possible

i509VCB:
In order to get anywhere with #314 (no_std support), we do need some changes around the place.

This pull requests replaces any references to `std` with `core`/`alloc` where possible in ash. Since std builds off of libcore and liballoc, this has no breaking changes.

Obviously not everything can be changed, namely things that only exist in std such as `CStr(ing)` (which is a discussion for another pull request) or things that require an operating system, namely `loaded` feature.

The lightweight abstractions over the extensions currently all require CStr, so those modules are for a later date.


MarijnS95:
Fwiw I prefer force-pushes to update existing PRs instead of opening duplicates and closing old ones.

PR #542, opened 2022-01-03
example: Refactor event loop handling for continuous redraw

msiglreith:
Fixes hangs in case of non-mailbox present modes. I assume acquiring an image from the swapchain hangs when redrawing unconditionally on each incoming event. See https://github.com/rust-windowing/winit/issues/2127 and https://github.com/MaikKlein/ash/issues/541

Tested on Windows with Nvidia gpu

### DIFF
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -185,23 +185,23 @@ impl ExampleBase {
             .borrow_mut()
             .run_return(|event, _, control_flow| {
                 *control_flow = ControlFlow::Wait;
-                f();
-                if let Event::WindowEvent {
-                    event:
-                        WindowEvent::CloseRequested
-                        | WindowEvent::KeyboardInput {
-                            input:
-                                KeyboardInput {
-                                    state: ElementState::Pressed,
-                                    virtual_keycode: Some(VirtualKeyCode::Escape),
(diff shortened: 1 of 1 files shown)


Ralith:
Per winit's docs, `RedrawRequested` isn't appropriate for continuously rendering applications. Instead, we should redraw on `MainEventsCleared`.

MarijnS95:
In addition, `ControlFlow` should be set to `Poll` instead of `Wait` so that it renders continuously, instead of whenever the next series of (input) event(s) arrive. Afaik there is no event that wakes up the loop continuously or monotonically.

MarijnS95 (approved):
Thanks @msiglreith! After all these changes I'd rename this PR (= the resulting title of the squash-merge commit) to something like "process all events before rendering a frame" and also work in the switch over to "polling" instead of sometimes blocking until any (unrelated to rendering) event. I'll leave the details of that to you unless you prefer me to change it.

msiglreith:
@MarijnS95 I tried to fine a better title, feel free to adjust :D

MarijnS95:
Yeah, I think that perfectly covers it without giving away too much of the (here unnecessary) implementation details, thanks!

PR #543, opened 2022-01-03
changelog: Add entries for recent PRs #533, #537, #542

MarijnS95:
We should probably come up with a PR template using checkmarks - bit of an annoying barrier IMO, but I keep forgetting to even document my own contributions.  Much easier to do them up-front than to come up with this in hindsight.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added `VK_KHR_get_surface_capabilities2` device extension (#530)
 
+### Changed
+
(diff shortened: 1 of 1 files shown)


Ralith:
I'm game for automated checking, so long as we can easily ignore it for stuff like internal refactoring/cleanup.

MarijnS95:
> I'm game for automated checking

I don't think it's possible to mandate a change to `Changelog.md` automatically, but we could request users to tick checkmarks in [a PR template](https://github.com/rust-windowing/winit/blob/master/.github/PULL_REQUEST_TEMPLATE.md) reminding them of the steps needed.

PR #545, opened 2022-01-05
Update README.md

hoj-senna:
Change Entry::new to Entry::linked in accordance with the changelog for ash 0.35.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,8 +221,9 @@ let pool = device.create_command_pool(&pool_create_info).unwrap();
 
 ### Optional linking
 
-The default `linked` cargo feature will link your binary with the Vulkan loader directly and expose the infallible `Entry::linked`.
-If your application can handle Vulkan being missing at runtime, you can instead enable the `loaded` feature to dynamically load Vulkan with `Entry::load`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -221,7 +221,7 @@ let pool = device.create_command_pool(&pool_create_info).unwrap();
 
 ### Optional linking
 
-The default `linked` cargo feature will link your binary with the Vulkan loader directly and expose the infallible `Entry::new`.
+The default `linked` cargo feature will link your binary with the Vulkan loader directly and expose the infallible `Entry::linked`.
 If your application can handle Vulkan being missing at runtime, you can instead enable the `loaded` feature to dynamically load Vulkan with `Entry::load`.
```
Ralith, line 225: Thanks for catching this! This text is also out of date with regard to the default, which was rolled back to `loaded`, so we should switch around the order we describe the features in and accurately identify which one is the default.

hoj-senna: Thanks, you're right, the text about the default should probably be corrected, too.

Ralith (approved):
Thanks!

MarijnS95:
Whoops, looks like a rebase-merge again :scream: 

Also, I'd appreciate a bit more descriptive titles than "Update README.md" :worried:

Ralith:
:facepalm:

PR #547, opened 2022-01-08
Update dependencies.

dvc94ch:


ash-window/Cargo.toml
```diff
@@ -15,13 +15,13 @@ workspace = ".."
 
 [dependencies]
 ash = { path = "../ash", version = "0.35", default-features = false }
-raw-window-handle = "0.3"
+raw-window-handle = "0.4.2"
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-raw-window-metal = "0.1"
+raw-window-metal = "0.2.0"
```
MarijnS95, line 21: #505

ash-window/Cargo.toml
```diff
@@ -15,13 +15,13 @@ workspace = ".."
...
 
 [dev-dependencies]
-winit = "0.19.4"
+winit = "0.26.1"
```
MarijnS95, line 24: This probably needs additional changes, no?

examples/Cargo.toml
```diff
@@ -5,8 +5,8 @@ authors = ["maik klein <maikklein@googlemail.com>"]
 edition = "2018"
 
 [dependencies]
-winit = "0.25.0"
-image = "0.10.4"
+winit = "0.26.1"
+image = "0.23.14"
```
MarijnS95, line 9: Do we need these minimum patch versions? Otherwise just omit them entirely, this is just an example.

generator/Cargo.toml
```diff
@@ -5,15 +5,15 @@ authors = ["Maik Klein <maikklein@googlemail.com>"]
...
 
 [dependencies]
-bindgen = "0.58"
-heck = "0.3"
+bindgen = "0.59.2"
+heck = "0.4.0"
 itertools = "0.10"
 nom = "6.0"
-once_cell = "1.7"
+once_cell = "1.9.0"
```
MarijnS95, line 12: Again no need for patch versions.

generator/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 #![recursion_limit = "256"]
 
-use heck::{CamelCase, ShoutySnakeCase, SnakeCase};
+use heck::{ToLowerCamelCase, ToShoutySnakeCase, ToSnakeCase};
```
MarijnS95, line 3: Can we use the cleaner `As*` wrappers in formatting places?

generator/src/lib.rs
```diff
@@ -1200,7 +1200,7 @@ pub fn generate_extension_commands<'a>(
 
     let ident = format_ident!(
         "{}Fn",
-        extension_name.to_camel_case().strip_prefix("Vk").unwrap()
+        extension_name.to_lower_camel_case().strip_prefix("Vk").unwrap()
```
MarijnS95, line 1203: Lower... So this won't strip the uppercase Vk prefix if the whole thing is lowered or will it retain characters that were uppercase?

Please run the generator at least once to make sure output doesn't change or remains what we expect.

MarijnS95 (commented):
The ash-window change should be reverted as we need to give that some more time and thought. Then, this only bumps example and generator dependencies which at least have no effect on the released crates.

ash-window/Cargo.toml
dvc94ch: cargo check --workspace seemed to work

generator/Cargo.toml
dvc94ch: that's a bug in cargo upgrade --skip-compatible

dvc94ch:
Closing in favor of #505, sorry didn't see it sooner.

ash-window/Cargo.toml
MarijnS95: `--workspace` only checks the crates (binaries, libraries) in the `workspace` (and it seems to be the default when the root `Cargo.toml` only defines the workspace but doesn't contain any crate of its own).

You should however also `check` the examples with `--examples`, or more generically `--all-targets`.

generator/Cargo.toml
MarijnS95: I'd do these by hand :)

MarijnS95:
No need to insta-close, we could use the rest of the upgrades but they need some more thought. Whenever upgrading libraries I always do a quick check in the changelog to see if there's anything useful, new features, or worrying shortcomings that I should keep in mind (which my only show at runtime).

PR #549, opened 2022-01-09
Convert `vk_bitflags_wrapped!` methods to `const fn`

### LINKED ISSUE #546: vk_bitflags_wrapped! should support const better
`is_empty`, `intersects`, and `contains` should all be `const fn`. Additionally, it would be very nice if there were `intersection`, `union`, `difference` const fn methods because bitwise operators do not work in const contexts except on stock integers. This is similar to how bitflags works: https://docs.rs/bitflags/1.3.2/bitflags/example_generated/struct.Flags.html

This would allow static validation directly on the bitflag types, as I am starting to do, to be a lot more convenient since they wouldn't need to be converted to/from raw to do any operations on them in a const context.

thorjelly:
Closes #546

So that vk_bitflags may be used directly in const contexts more ergonomically, similar to bitflags: https://docs.rs/bitflags/1.3.2/bitflags/example_generated/struct.Flags.html

### DIFF
--- a/ash/src/vk/macros.rs
+++ b/ash/src/vk/macros.rs
@@ -32,18 +32,6 @@ macro_rules! vk_bitflags_wrapped {
             pub const fn contains(self, other: Self) -> bool {
                 self.0 & other.0 == other.0
             }
-            #[inline]
-            pub const fn intersection(self, other: Self) -> Self {
-                Self(self.0 & other.0)
-            }
-            #[inline]
-            pub const fn union(self, other: Self) -> Self {
-                Self(self.0 | other.0)
-            }
-            #[inline]
-            pub const fn difference(self, other: Self) -> Self {
-                Self(self.0 & !other.0)
-            }
         }
         impl ::std::ops::BitOr for $name {
             type Output = Self;
--- a/ash/src/vk/macros.rs
+++ b/ash/src/vk/macros.rs
@@ -20,17 +20,29 @@ macro_rules! vk_bitflags_wrapped {
                 self.0
             }
             #[inline]
-            pub fn is_empty(self) -> bool {
-                self == Self::empty()
+            pub const fn is_empty(self) -> bool {
+                self.0 == Self::empty().0
             }
             #[inline]
-            pub fn intersects(self, other: Self) -> bool {
-                self & other != Self::empty()
+            pub const fn intersects(self, other: Self) -> bool {
+                !Self(self.0 & other.0).is_empty()
             }
             #[doc = r" Returns whether `other` is a subset of `self`"]
             #[inline]
-            pub fn contains(self, other: Self) -> bool {
-                self & other == other
+            pub const fn contains(self, other: Self) -> bool {
+                self.0 & other.0 == other.0
+            }
+            #[inline]
+            pub const fn intersection(self, other: Self) -> Self {
+                Self(self.0 & other.0)
+            }
+            #[inline]
+            pub const fn union(self, other: Self) -> Self {
+                Self(self.0 | other.0)
+            }
+            #[inline]
+            pub const fn difference(self, other: Self) -> Self {
+                Self(self.0 & !other.0)
             }
         }
         impl ::std::ops::BitOr for $name {


ash/src/vk/macros.rs
```diff
@@ -20,17 +20,29 @@ macro_rules! vk_bitflags_wrapped {
...
+            pub const fn contains(self, other: Self) -> bool {
+                self.0 & other.0 == other.0
+            }
+            #[inline]
+            pub const fn intersection(self, other: Self) -> Self {
+                Self(self.0 & other.0)
+            }
+            #[inline]
+            pub const fn union(self, other: Self) -> Self {
+                Self(self.0 | other.0)
```
MarijnS95, line 41: I don't think this function (and the others) is any more convenient nor ergonomic than unwrapping and re-wrapping, as you'd have to chain a bunch of these function calls.

Context https://github.com/MaikKlein/ash/pull/537#discussion_r776037034

thorjelly: I think it'd be situational. To me, `vk::BufferUsageFlags::VERTEX_BUFFER.union(vk::BufferUsageFlags::TRANSFER_DST)` is preferable and clearer than `vk::BufferUsageFlags::from_raw(vk::BufferUsageFlags::VERTEX_BUFFER::as_raw() | vk::BufferUsageFlags::TRANSFER_DST::as_raw())`

MarijnS95: Perhaps we should just turn the wrapped flags into `pub` instead of `pub(crate)` and drop `from/as_raw` too? On my machine that nets 0.1s compile-time win from ~9.45s, and there's no casting/validation ongoing here anyway.

Then crate users should also be able to write something like `pub const RGBA: Self = Self(Self::R.0 | Self::G.0 | Self::B.0 | Self::A.0);` in const context instead of those `from/as_raw` contraptions.

Fortunately the addition of these `union` functions don't seem to add any compile-time on my end, so I'm not entirely opposed to this.

Note that in the past your PR should have updated the generator where this static macro also lives, when #552 lands that is not needed anymore.

thorjelly: I think making the wrapped flags pub would be good idea too.

Honestly thinking about it further you're probably right that `union`, `difference`, `intersection` aren't super useful. Making the wrapped value pub would simplify things. Honestly, I mainly included them because bitflags did it and it felt like a good idea, but using them in this context still doesn't feel terribly great except in simple cases.

MarijnS95: In the end it can all be blamed on `BitOr` and friends not supporting `const` one way or another.

I'll get @Ralith involved in the decision, but I'm quite sure he agrees on simplifying this by removing `(crate)` from `pub` and doing away with the raw getters and setters entierly :)

ash/src/vk/macros.rs
```diff
@@ -20,17 +20,29 @@ macro_rules! vk_bitflags_wrapped {
...
             }
             #[doc = r" Returns whether `other` is a subset of `self`"]
             #[inline]
-            pub fn contains(self, other: Self) -> bool {
-                self & other == other
+            pub const fn contains(self, other: Self) -> bool {
+                self.0 & other.0 == other.0
+            }
+            #[inline]
+            pub const fn intersection(self, other: Self) -> Self {
```
aloucks, line 36: The `bitflags` crate calls this [`intersects`](https://docs.rs/bitflags/latest/bitflags/example_generated/struct.Flags.html#method.intersects). I think we should use the same name.

ash/src/vk/macros.rs
Ralith: If dropping the accessors in favor of a more concise and equally obvious option saves us time, I'm all in favor! 1%-ish isn't a lot, but these things add up.

MarijnS95: Alright! Since I tested this variant locally I can PR that tomorrow. @thorjelly if you can meanwhile drop the additional functions so that we just have the constifying of existing functions in this PR that'd be awesome :) (don't forget to rebase/merge, #552 is now merged).

thorjelly: Thanks, I've dropped the additional functions and merged.

MarijnS95: Nice!

I'll open a PR later today that removes `(crate)` from `pub`, but i'll likely keep `from/as_raw` for now but with `#[deprecated]` annotation so that we can do a nonbreaking patch release, there have been no breaking changes thus far IIRC.

MarijnS95 (approved):

ash/src/vk/macros.rs
MarijnS95: #554 (done with `#[deprecated]`, will make a patch release soon with all these nice updates.)

PR #552, opened 2022-01-10
generator: Exclude static macro code (`macros.rs`) from the generator

MarijnS95:
Much like `platform_types.rs` and `vk.rs` before, there is no need to generate this file as its contents are completely static.  This allows direct editing of the file without realizing that a copy lives in the generator - and will overwrite `macros.rs` when the generator is being run.

CC #549 which updated this file without changing the generator.


Ralith:
Nice cleanup!

PR #555, opened 2022-01-11
changelog: Add entries for recent PRs #545, #534, #553, #551, #549

MarijnS95:
And drop the deprecated =/- markdown syntax from our readme: this is analogous to #/## for a h1/h2 header, instead of defining a title and (usually smaller font) subtitle or description.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Convert `vk_bitflags_wrapped!` methods to `const fn` (#549)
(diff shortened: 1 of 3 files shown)


MarijnS95:
@Ralith I think we can push out the first patch release with this merged, probably without #554 for the time being (I'll have to scoot over its changelog addition that PR).

Ralith (approved):

Ralith:
Should mention of #554 be dropped from the commit message?

MarijnS95:
Removed through the web-UI in the _squash_-merge. I'll draft a patch release tonight or tomorrow, would like to use #556.

PR #556, opened 2022-01-13
extensions/ext: Add VK_EXT_calibrated_timestamps extension

MarijnS95:
This extension can be used to correlate timestamps on the GPU timeline (`vkCmdWriteTimestamp`, `vkCmdWriteTimestamp2KHR`) to the CPU timeline (ie. `CLOCK_MONOTONIC` and friends).


ash/src/extensions/ext/calibrated_timestamps.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPhysicalDeviceCalibrateableTimeDomainsEXT.html>"]
+    pub unsafe fn get_physical_device_calibrateable_time_domains(
+        &self,
+        physical_device: vk::PhysicalDevice,
+    ) -> VkResult<Vec<vk::TimeDomainEXT>> {
+        read_into_uninitialized_vector(|count, data| {
+            self.fp
+                .get_physical_device_calibrateable_time_domains_ext(physical_device, count, data)
```
MarijnS95, line 29: Whoops, we can't actually load this through `get_device_proc_addr`. As we did with at least one other extension, just load it through `get_instance_proc_addr` and take the indirection table hit :(

MarijnS95:
If anyone knows why they called the function here `calibrateable` instead of the arguably more correct `calibratable`, do let me know!

ash/src/extensions/ext/calibrated_timestamps.rs
Ralith: Is the problem that the extension has both an instance-level and a device-level function? Sounds like we should probably handle those separately.

ash/src/extensions/ext/calibrated_timestamps.rs
```diff
@@ -0,0 +1,64 @@
...
+        })
+    }
+
+    #[doc = "<https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetCalibratedTimestampsEXT.html>"]
+    ///
```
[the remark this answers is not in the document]
Ralith: Seems sensible.

Ralith:
Wouldn't be the first editorial mistake in the API; raise an issue upstream if you're sure.

Ralith (approved):

MarijnS95:
https://en.wiktionary.org/wiki/calibrateable Says that it works as an alternative, so I guess no harm done :grimacing:

PR #557, opened 2022-01-18
Add conversions between Extent2D, Extent3D and Rect2D

swooster:
When using `ash`, I frequently find myself wishing that converting between 2D/3D extents was less verbose, since it's so common. Likewise, you need to convert from an `Extent2D` to a `Rect2D` any time you want to start a render pass that affects an entire framebuffer.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- Added conversions from `Extent2D` to `Extent3D` and `Rect2D` (#557)
+
 ## [0.35.1] - 2022-01-18
 
 ### Added
--- a/ash/src/vk/prelude.rs
+++ b/ash/src/vk/prelude.rs
@@ -1,3 +1,5 @@
+use crate::vk;
+
 /// Holds 24 bits in the least significant bits of memory,
 /// and 8 bytes in the most significant bits of that memory,
 /// occupying a single [`u32`] in total. This is commonly used in
@@ -30,4 +32,23 @@ impl Packed24_8 {
 impl super::ColorComponentFlags {
     /// Contraction of [`Self::R`] | [`Self::G`] | [`Self::B`] | [`Self::A`]
     pub const RGBA: Self = Self(Self::R.0 | Self::G.0 | Self::B.0 | Self::A.0);
+}
+
+impl From<vk::Extent2D> for vk::Extent3D {
+    fn from(value: vk::Extent2D) -> Self {
+        Self {
+            width: value.width,
+            height: value.height,
+            depth: 1,
+        }
+    }
+}
+
+impl From<vk::Extent2D> for vk::Rect2D {
+    fn from(extent: vk::Extent2D) -> Self {
+        Self {
+            offset: Default::default(),
+            extent,
+        }
+    }
 }
--- a/examples/src/bin/texture.rs
+++ b/examples/src/bin/texture.rs
@@ -264,7 +264,8 @@ fn main() {
         let image = image::load_from_memory(include_bytes!("../../assets/rust.png"))
             .unwrap()
             .to_rgba8();
-        let image_dimensions = image.dimensions();
+        let (width, height) = image.dimensions();
+        let image_extent = vk::Extent2D { width, height };
         let image_data = image.into_raw();
         let image_buffer_info = vk::BufferCreateInfo {
             size: (std::mem::size_of::<u8>() * image_data.len()) as u64,
@@ -313,11 +314,7 @@ fn main() {
         let texture_create_info = vk::ImageCreateInfo {
             image_type: vk::ImageType::TYPE_2D,
             format: vk::Format::R8G8B8A8_UNORM,
-            extent: vk::Extent3D {
-                width: image_dimensions.0,
-                height: image_dimensions.1,
-                depth: 1,
-            },
+            extent: image_extent.into(),
             mip_levels: 1,
             array_layers: 1,
             samples: vk::SampleCountFlags::TYPE_1,
@@ -388,11 +385,7 @@ fn main() {
                             .layer_count(1)
                             .build(),
                     )
-                    .image_extent(vk::Extent3D {
-                        width: image_dimensions.0,
-                        height: image_dimensions.1,
-                        depth: 1,
-                    });
+                    .image_extent(image_extent.into());
 
                 device.cmd_copy_buffer_to_image(
                     texture_command_buffer,
@@ -623,10 +616,7 @@ fn main() {
             min_depth: 0.0,
             max_depth: 1.0,
         }];
-        let scissors = [vk::Rect2D {
-            extent: base.surface_resolution,
-            ..Default::default()
-        }];
+        let scissors = [base.surface_resolution.into()];
         let viewport_state_info = vk::PipelineViewportStateCreateInfo::builder()
             .scissors(&scissors)
             .viewports(&viewports);
@@ -727,10 +717,7 @@ fn main() {
             let render_pass_begin_info = vk::RenderPassBeginInfo::builder()
                 .render_pass(renderpass)
                 .framebuffer(framebuffers[present_index as usize])
-                .render_area(vk::Rect2D {
-                    offset: vk::Offset2D { x: 0, y: 0 },
-                    extent: base.surface_resolution,
-                })
+                .render_area(base.surface_resolution.into())
                 .clear_values(&clear_values);
 
             record_submit_commandbuffer(
--- a/examples/src/bin/triangle.rs
+++ b/examples/src/bin/triangle.rs
@@ -277,10 +277,7 @@ fn main() {
             min_depth: 0.0,
             max_depth: 1.0,
         }];
-        let scissors = [vk::Rect2D {
-            offset: vk::Offset2D { x: 0, y: 0 },
-            extent: base.surface_resolution,
-        }];
+        let scissors = [base.surface_resolution.into()];
         let viewport_state_info = vk::PipelineViewportStateCreateInfo::builder()
             .scissors(&scissors)
             .viewports(&viewports);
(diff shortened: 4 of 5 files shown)


swooster:
This is just a first-pass PR since there are some outstanding questions:

* Should conversion from `Extent3D` to `Extent2D` require the `depth` to be 1? (should I be using `TryFrom` instead of `From`?)
* What's the proper way to add this? (I just made a best guess about how to tweak the generator)
* When I tried to regenerate everything, this caused additional changes to `native.rs` (e.g. changing `StdVideoH264PocType` from `c_uint` to `c_int`) that I've avoided committing because I don't believe my changes should have affected that. Any idea what's up with that?

Anyway, let me know if this seems sane, or how I can improve things.

MarijnS95:
> * Should conversion from `Extent3D` to `Extent2D` require the `depth` to be 1? (should I be using `TryFrom` instead of `From`?)

I'd do it the way vector libraries such as `glam` do: a `fn extend(depth: u32)` so that it is up to the user to decide, instead of a seemingly unsuspecting `from`/`into`.

> * What's the proper way to add this? (I just made a best guess about how to tweak the generator)

We're moving away from having static code in the generator. Feel free to add this in `ash/src/vk/prelude.rs`.

Over time we will work towards a cleaner split between handwritten files, and dump everything autogenerated in an `auto/` subfolder. For now there's not much that's hand-written, so the prelude is fine.

> * When I tried to regenerate everything, this caused additional changes to `native.rs` (e.g. changing `StdVideoH264PocType` from `c_uint` to `c_int`) that I've avoided committing because I don't believe my changes should have affected that. Any idea what's up with that?

That is odd, are submodules or your local `Cargo.lock` (ie. `bindgen`) out of date? The CI uses the most recent version just as I use locally, and that change doesn't show. I don't recall those types changing there historically, though.

Ralith:
For 2D -> 3D, I think a `From` impl that bakes in `depth = 1` makes sense both logically, because that's the natural embedding of a 2D slice in a 3D space, and practically, because it is *super* common to e.g. construct an `Extent3D` for a 2D image in `ImageCreateInfo`, which requires `depth = 1` exactly. FRU syntax can be used if something else is required; we could also offer `extend` but that seems like overkill to me.

Is there a specific use case for the 3D -> 2D case?

Ralith (approved):

swooster:
> I'd do it the way vector libraries such as `glam` do: a `fn extend(depth: u32)` so that it is up to the user to decide, instead of a seemingly unsuspecting `from`/`into`.

Unless you feel strongly about this, my reasoning is the same as in @Ralith's comment; my own use-cases (and the `ash` examples) have only ever had a need for a depth of 1.

> EDIT: For the truncation, not really sure what makes most sense, we try to not be too opinionated, and as a result not provide too many such helpers...

I'll skip the 3D-to-2D conversion for now since that's less clear cut..? While I've encountered situations where I've wanted that, maybe it'll turn out to be easy enough to work around using 2D-to-3D conversions.

> We're moving away from having static code in the generator. Feel free to add this in `ash/src/vk/prelude.rs`.

Ok, I've redone this to use the prelude instead of modifying the generator.

> That is odd, are submodules or your local `Cargo.lock` (ie. `bindgen`) out of date? The CI uses the most recent version just as I use locally, and that change doesn't show. I don't recall those types changing there historically, though.

I don't _think_ `bindgen `was out of date, as I set it up for the first time yesterday (I installed LLVM 13.0.0), but maybe I've installed it wrong? (I was just flailing around trying to get to a point where I could run the generator again)

> As an aside, you may showcase your new helpers in the examples :)

Ok, examples updated.

swooster:
Huh... I just now noticed there are two preludes: `ash::prelude` and `ash::vk::prelude`. Which one should I have modified?

MarijnS95:
> Unless you feel strongly about this, my reasoning is the same as in @Ralith's comment; my own use-cases (and the `ash` examples) have only ever had a need for a depth of 1.

Gave his comment a thumbs-up because of that, no strong feelings either way.

> I don't _think_ `bindgen` was out of date, as I set it up for the first time yesterday (I installed LLVM 13.0.0), but maybe I've installed it wrong? (I was just flailing around trying to get to a point where I could run the generator again)

It might be the `bindgen` crate itself that the `generator` depends on, instead of the installable binary. Though even its library version should leverage the system `clang` and `LLVM`? Using LLVM 13 here too, no diff when running the generator. What host OS and architecture are you running on?

> > As an aside, you may showcase your new helpers in the examples :)
> 
> Ok, examples updated.

Shaves off a lot of code :wink: 

> Huh... I just now noticed there are two preludes: `ash::prelude` and `ash::vk::prelude`. Which one should I have modified?

`src/ash/vk/prelude.rs`, as per my comment above. These helpers solely relate to the `vk` definitions and go there. `src/ash/prelude.rs` holds more general helpers that are used by the wrappers outside the `vk` module.

ash/src/vk/prelude.rs
```diff
@@ -1,3 +1,5 @@
+use crate::vk;
```
MarijnS95, line 1: `ColorComponentFlags` uses `super::`, perhaps we should update that if everyone agrees going through the crate root is cleaner?

MarijnS95 (commented):
D'oh I forgot to ask in my previous comment: please add an entry to `ash/changelog.md` :)

ash/src/vk/prelude.rs
MarijnS95: Addendum: The documentation for `Packed24_8` goes through `super::` as well, I think I did that on purpose.

I bet we can remove those indirections (ie. ``[`vk::AccelerationStructureInstanceKHR`][super::AccelerationStructureInstanceKHR]`` -> ``[`vk::AccelerationStructureInstanceKHR`]``) if `vk` is imported in scope.

examples/src/bin/texture.rs
```diff
@@ -264,7 +264,8 @@ fn main() {
         let image = image::load_from_memory(include_bytes!("../../assets/rust.png"))
             .unwrap()
             .to_rgba8();
-        let image_dimensions = image.dimensions();
+        let (width, height) = image.dimensions();
+        let image_dimensions = vk::Extent2D { width, height };
```
filnet, line 268: could add and use:
```
impl From<(u32, u32)> for vk::Extent2D {
    fn from(value: (u32, u32)) -> Self {
        Self {
            width: value.0,
            height: value.1,
        }
    }
}
````

Comment (filnet) [inline examples/src/bin/texture.rs:268]: rename local variable to `image_extent2D` ?

File: ash/src/vk/prelude.rs (hunk 1 above)
Comment (swooster) [reply ash/src/vk/prelude.rs:1]: Ah, I saw a lot of `use crate::vk::submodule::*;` in `ash::vk::constants`/`ash::vk::definitions` etc. I'm also happy to go with `super::Stuff` for consistency with `ColorComponentFlags`.

Comment (MarijnS95) [reply ash/src/vk/prelude.rs:1]: That's how the generator emits those since the old days, I haven't cleaned them up but it's probably good practice to do so. My only reason for preferring this is making the module oblivious of how and where it is located within the `crate`, making it a bit easier to relocate even if that's something we'd never do. I'm sure @Ralith has a better judgement here, let's wait for that.

**Comment by swooster** [comment]
> It might be the `bindgen` crate itself that the `generator` depends on, instead of the installable binary. Though even its library version should leverage the system `clang` and `LLVM`? Using LLVM 13 here too, no diff when running the generator. What host OS and architecture are you running on?

I'm running Windows 10 on x86. The difference persists even after running `cargo update`.

**Comment by filnet** [comment]
> I'm running Windows 10 on x86. The difference persists even after running `cargo update`.

Did you run `cargo update` in ash or in your project ? Got bitten by that in the past.

File: ash/src/vk/prelude.rs (hunk 1 above)
Comment (MarijnS95) [reply ash/src/vk/prelude.rs:1]: Fwiw this change is being test-driven in #559, documentation is broken anyway :(

File: examples/src/bin/texture.rs (hunk 2 above)
Comment (swooster) [reply examples/src/bin/texture.rs:268]: I like the idea of conversions to/from tuples because of how it'd make interop simpler, though both @MarijnS95 and @Ralith have told me that having a small minimally-opinionated API is a goal, so I'm a little reticent to solve more than my biggest pain points until additional people express enthusiasm for tuple conversions.

**Comment by MarijnS95** [comment]
> I'm running Windows 10 on x86. The difference persists even after running `cargo update`.

Yeah, that explains things. We had more Windows-vs-Linux `bindgen` issues with the Video extension in the past: https://github.com/MaikKlein/ash/pull/417#issuecomment-850998722.

**Comment by swooster** [comment]
> D'oh I forgot to ask in my previous comment: please add an entry to `ash/changelog.md` :)

I _just_ now noticed this comment... added.

File: examples/src/bin/texture.rs (hunk 2 above)
Comment (swooster) [reply examples/src/bin/texture.rs]: I've renamed it to `image_extent`... `image_extent2D` violates my understanding of naming conventions due to the uppercase `D`. I'd be ok with `image_extent_2d` though my preference is to omit the `_2d`.

File: Changelog.md [hunk 3]
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added conversions from `Extent2D` to `Extent3D` and `Rect2D` (#557)
```
MarijnS95, line 13: You'll have to introducce a new `### Added` heading under `## [Unreleased] - ReleaseDate
` though, this isn't part of `## [0.35.1] - 2022-01-18` that released yesterday :grimacing:

swooster: D'oh! 🐑😬 I didn't notice that 0.35.1 had been released and wrongly assumed those changes were still in the unreleased section.

ash/src/vk/prelude.rs
Ralith: I favor going through `vk` rather than `super`; it's more concise and more consistent with how everyone's used to dealing with ash types. I'm not necessarily opposed to globbing in modules from the same crate either, but it's probably not justified here.

Ralith (approved):

examples/src/bin/texture.rs
Ralith: If we were to pursue this (and I don't find it super compelling) then we should do it as a separate PR so as not to bog down this one. For most purposes I favor nominal over structural typing, for the usual reasons.

MarijnS95 (approved):

PR #559, opened 2022-01-19
ci: Build-test documentation with warnings disallowed, and fix broken links

MarijnS95:
I was pretty certain our documentation was at least build-tested since that's generally one of the first things I check/PR on the crates I touch... Supposedly not here. It allowed some broken links to slip in, most recently in #530 and #537 where I _wrongly assumed_ the CI was having my back. ~Let's make sure it fails now before pushing the necessary fixes.~ CI is properly working and finding+disallowing issues: https://github.com/MaikKlein/ash/runs/4865846622?check_suite_focus=true

### DIFF
--- a/ash/src/vk/prelude.rs
+++ b/ash/src/vk/prelude.rs
@@ -4,9 +4,9 @@ use crate::vk;
 /// and 8 bytes in the most significant bits of that memory,
 /// occupying a single [`u32`] in total. This is commonly used in
 /// [acceleration structure instances] such as
-/// [`vk::AccelerationStructureInstanceKHR`][super::AccelerationStructureInstanceKHR],
-/// [`vk::AccelerationStructureSRTMotionInstanceNV`][super::AccelerationStructureSRTMotionInstanceNV] and
-/// [`vk::AccelerationStructureMatrixMotionInstanceNV`][super::AccelerationStructureMatrixMotionInstanceNV].
+/// [`vk::AccelerationStructureInstanceKHR`],
+/// [`vk::AccelerationStructureSRTMotionInstanceNV`] and
+/// [`vk::AccelerationStructureMatrixMotionInstanceNV`].
 ///
 /// [acceleration structure instances]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureInstanceKHR.html#_description
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
@@ -29,7 +29,11 @@ impl Packed24_8 {
     }
 }
 
-impl super::ColorComponentFlags {
+// Intradoc `Self::` links refuse to resolve if `ColorComponentFlags`
+// isn't directly in scope: https://github.com/rust-lang/rust/issues/93205
+use vk::ColorComponentFlags;
+
+impl ColorComponentFlags {
     /// Contraction of [`Self::R`] | [`Self::G`] | [`Self::B`] | [`Self::A`]
     pub const RGBA: Self = Self(Self::R.0 | Self::G.0 | Self::B.0 | Self::A.0);
 }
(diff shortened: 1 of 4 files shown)


ash/src/vk/prelude.rs
```diff
@@ -27,7 +29,7 @@ impl Packed24_8 {
     }
 }
 
-impl super::ColorComponentFlags {
-    /// Contraction of [`Self::R`] | [`Self::G`] | [`Self::B`] | [`Self::A`]
+impl vk::ColorComponentFlags {
+    /// Contraction of [`vk::ColorComponentFlags::R`] | [`vk::ColorComponentFlags::G`] | [`vk::ColorComponentFlags::B`] | [`vk::ColorComponentFlags::A`]
```
MarijnS95, line 33: This was throwing a very strange:

```rust
warning: unresolved link to `ColorComponentFlags::R`
  --> ash/src/vk/prelude.rs:31:26
   |
31 |     /// Contraction of [`Self::R`] | [`Self::G`] | [`Self::B`] | [`Self::A`]
   |                          ^^^^^^^ no item named `ColorComponentFlags` in scope
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default
```

Despite residing in an `impl` block, `Self` should just expand to that. Maybe a `rustdoc` bug? Perhaps isolated to `const`.

ash/src/vk/prelude.rs
```diff
@@ -1,10 +1,12 @@
+use super::super::vk;
```
MarijnS95, line 1: This is still up for debate, see https://github.com/MaikKlein/ash/pull/557#discussion_r787534974.

ash/src/vk/prelude.rs
Ralith: Weird! Reported upstream? I think this form of the doc is fine too, fortunately.

ash/src/vk/prelude.rs
Ralith: I favor going through the root in ash's case, since it's fairly concise, makes impl code read like usage code, and makes the docs easier.

Ralith (approved):

ash/src/vk/prelude.rs
MarijnS95: https://github.com/rust-lang/rust/issues/93205

I planned to report it upstream but got discouraged after finding out that play.rust-lang.org _still_ can't build docs.

As it seems `Self` resolves to the type name only, not the full path. If I make sure that `ColorComponentFlags` is in scope (after all the error says ``no item named `ColorComponentFlags` in scope``) it'll work again.

What do you think? I've already gone back-and-forth multiple times on whether to write `Self` in public docs (makes for elegant links in markdown) or write out the full name. Do think I prefer `Self` though, that takes out the "was I looking at the `ColorComponentFlags` already?" mental question.

MarijnS95: Went with `Self::`, made sure that `ColorComponentFlags` is available, and linked to the Rust issue.

ash/src/vk/prelude.rs
MarijnS95: Fair enough, my only objection would be to make this `mod` (or rather it's parent `mod vk`) self-supporting, ie. oblivious to where it's located in the tree. No strong feelings though and I prefer your suggestion if this is considered more of a "copy paste this is how you can use these types" example.

ash/src/vk/prelude.rs
Ralith: I actually tend to favor dropping the prefix entirely (as if `Self::*` was in scope, though rust doesn't actually support globbing associated items (yet?)), but I don't have strong feelings; this is fine as is.

Ralith (approved):

ash/src/vk/prelude.rs
MarijnS95: We could us a "named (intradoc) link" (no idea if that's the technical terminology):

```markdown
Contraction of [`R`][Self::R] | [`G`][Self::G] | [`B`][Self::B] | [`A`][Self::A]
```

image

Ralith: Yeah, exactly. Bit more concise to read, if a bit less convenient to write.

PR #561, opened 2022-01-24
Use link text in docs to omit `Self::` prefix in rendered docs

MarijnS95:
Suggestion from [#559].

[#559]: https://github.com/MaikKlein/ash/pull/559#discussion_r790178735

~Note that `Self::` in `_len()` function docs hasn't been replaced (yet?) as the longer function names make this format rather verbose.  However, we should probably prefer docs readability over code readability?~

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -264,10 +264,10 @@ impl Entry {
         &self.entry_fn_1_1
     }
 
-    #[deprecated = "This function is unavailable and therefore panics on Vulkan 1.0, please use `try_enumerate_instance_version` instead"]
+    #[deprecated = "This function is unavailable and therefore panics on Vulkan 1.0, please use `try_enumerate_instance_version()` instead"]
     /// <https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkEnumerateInstanceVersion.html>
     ///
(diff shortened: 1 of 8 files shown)


Ralith (approved):

MarijnS95:
@Ralith Consider it regexed :)

Ralith (approved):
Nice!

ash/src/entry.rs
```diff
@@ -264,10 +264,10 @@ impl Entry {
         &self.entry_fn_1_1
     }
 
-    #[deprecated = "This function is unavailable and therefore panics on Vulkan 1.0, please use `try_enumerate_instance_version` instead"]
+    #[deprecated = "This function is unavailable and therefore panics on Vulkan 1.0, please use `try_enumerate_instance_version()` instead"]
```
MarijnS95, line 267: I thought (intradoc) links were also possible in deprecation blocks (similar to other markdown), but nope :(

Ralith: Aw, that seems like a really useful place for them too. Feature request upstream?

MarijnS95: Heh feature requests have to go to the "internals" forum, I'll see what I can do.

PR #563, opened 2022-01-25
Update Vulkan-Headers to 1.3.206

MarijnS95:
Vulkan 1.3 is out, with many extensions stabilized.  This'll have to remain draft until the next spec version (hopefully next week) where (again, hopefully) https://github.com/KhronosGroup/Vulkan-Docs/pull/1746 is incorporated - otherwise we'll have to work around in our generator :(

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Update Vulkan-Headers to 1.3.205 (#563)
+- Update Vulkan-Headers to 1.3.206 (#563)
 
 ## [0.35.2] - 2022-02-19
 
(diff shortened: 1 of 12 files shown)


MarijnS95:
Can't update to 1.3.205 either, `vk.xml` got broken there as well: https://github.com/KhronosGroup/Vulkan-Headers/pull/258#discussion_r799449490

Rua:
Seems to have been fixed in https://github.com/KhronosGroup/Vulkan-Docs/issues/1764.

Ralith (approved):
If it wasn't clear, I think this is a breaking change, because e.g. `Extends*` for structs that were made core have been renamed.

MarijnS95:
Should have probably left out the `unsafe` part, since being able to use this trait bound in a safe wrapper is already well enough of a possibility to risk an API break.

I don't want to go figure out how to add a bunch of `type` aliases for these either, at least :)

Ralith:
Yeah, I don't think that's worth the trouble; this only comes up on Vulkan minor version bumps anyway.

PR #564, opened 2022-01-25
ash: Remove unnecessary trivial_casts and trivial_numeric_casts

MarijnS95:
While making the code only marginally harder to read such casts can also introduce subtle bugs when used incorrectly, and are best omitted whenever unnecessary: Rust already coerces borrows into raw pointers when the types on both ends are clear, and even then there remain many casts that are identical to the source type.

In addition these errors show up when using a local crate reference to `ash` in a workspace that uses "the `.cargo/config.toml` setup" from [EmbarkStudios/rust-ecosystem#68] to configure linter warnings project-wide instead of for all crates in that workspace individually.  In our case aforementioned linter warnings are enabled on top of Embark's configuration, leading to a lot of these warnings in our build process.

[EmbarkStudios/rust-ecosystem#68]: https://github.com/EmbarkStudios/rust-ecosystem/pull/68


ash/src/extensions/khr/acceleration_structure.rs
```diff
@@ -156,7 +156,7 @@ impl AccelerationStructure {
         info: &vk::CopyAccelerationStructureInfoKHR,
     ) -> VkResult<()> {
         self.fp
-            .copy_acceleration_structure_khr(self.handle, deferred_operation, info as *const _)
+            .copy_acceleration_structure_khr(self.handle, deferred_operation, info)
```
MarijnS95, line 159: Lots of coercion in this file - most extensions did this properly already.

ash/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![deny(clippy::use_self)]
+#![warn(/* trivial_casts, */ trivial_numeric_casts)]
```
MarijnS95, line 2: I enabled `trivial_casts` here to find many violations "by hand". Unfortunately we have many false-positives where `rustc` thinks a coercion can take place because it firsts casts a borrow to a raw pointer of the same type, then casts that pointer to a different type. But that coercion of the first cast cannot happen without having the type explicit :(

```rust
warning: trivial cast: `&'a mut T` as `*mut T`
   --> ash/src/extensions/experimental/amd.rs:697:28
    |
697 |             let next_ptr = next as *mut T as *mut BaseOutStructure;
    |                            ^^^^^^^^^^^^^^
    |
    = help: cast can be replaced by coercion; this might require a temporary variable
```

generator/src/lib.rs
```diff
@@ -1499,7 +1500,7 @@ pub fn derive_debug(_struct: &vkxml::Struct, union_types: &HashSet<&str>) -> Opt
         let debug_value = if is_static_array(field) && field.basetype == "char" {
             quote! {
                 &unsafe {
-                    ::std::ffi::CStr::from_ptr(self.#param_ident.as_ptr() as *const c_char)
+                    ::std::ffi::CStr::from_ptr(self.#param_ident.as_ptr())
```
MarijnS95, line 1503: I made sure this one is safe: `CStr::from_ptr` already takes a `c_char` and because `field.basetype == "char"` here the underlying type of `$param_ident` is already `[c_char; ...]`. Hence, no issues with `i8` vs `u8` when compiling on armv7/aarch64 :)

ash/src/lib.rs
Ralith: It might be helpful to make more explicit use of the `cast` method on pointers. This should be able to replace most uses of `as` entirely. For example, `let next_ptr = <*mut T>::cast::<BaseOutStructure>(next)`.

ash/src/extensions/experimental/amd.rs
```diff
@@ -240,11 +240,16 @@ impl<'a> ::std::ops::Deref for PhysicalDeviceGpaPropertiesAmdBuilder<'a> {
...
     where
         T: ExtendsPhysicalDeviceGpaPropertiesAmd,
     {
-        self.inner.p_next = next as *mut T as *mut c_void;
+        unsafe {
+            let next_ptr = <*const T>::cast(next);
+            let last_next = ptr_chain_iter(next).last().unwrap();
+            (*last_next).p_next = self.inner.p_next as _;
+            self.inner.p_next = next_ptr;
+        }
```
MarijnS95, line 252: Not sure why this wasn't matching the other `push_next` :thinking: - the name change is an API break though but we're in the `experimental` module (that isn't guarded by a feature though).

generator/src/lib.rs
```diff
@@ -1755,7 +1773,7 @@ pub fn derive_setters(
                     //                 next chain
                     let last_next = ptr_chain_iter(next).last().unwrap();
                     (*last_next).p_next = self.inner.p_next as _;
```
MarijnS95, line 1775: Not sure about this `as _` - it seems we have some `inner.p_next` that are `*mut c_void`, others are `*const c_void`. The target type of `BaseOutStructure::p_next` (LHS here) is always `*mut BaseOutStructure`.

ash/src/extensions/experimental/amd.rs
Ralith: Seems fine to me.

generator/src/lib.rs
Ralith: Shouldn't it be `BaseInStructure` for the `*const` cases?

MarijnS95: That wouldn't be entirely faithful since we can still be assigning to more `p_next` down the line, realistically this should be one long chain of `*mut`?

Ralith: I'm not sure I follow. We only write to `self.p_next` and `next.p_next`, we don't traverse the pointers ourselves.

MarijnS95: Never mind.

Point is that, if we add a `next = A->B` to an existing chain `C->D->E` (with `C->A->B->D->E` as a result), if `A` were to point to B through `p_next: *const` we would "not be allowed" to assign `D` as next to `B`, since `B` "is const here".

> we don't traverse the pointers ourselves.

In that sense yes, `ptr_chain_iter` definitely does to get hold of `last_next`.

In other words, this whole Vulkan declaration seems a bit flawed and I haven't thus far found (nor really searched for) any reasons for some structs to have this const. That pretty much "forces" you to always build the chain from end to start.

(Before you ask, I'm aware that the above doesn't directly map to writing something _inside_ `self.inner.p_next`, where we do the const cast above. It's trivial to come up with an example that does though, with multiple calls of `push_next`).

Again though, none of this really matters, it's just an oddity and shouldn't hold us back from merging this cleanup. We can look at the `as _` another time if there appears to be some valid reasoning that we should follow through in our API, though I doubt it (most likely it'll only limit the order to prepending).

MarijnS95:
@Ralith is this something we can get in before any of the followup releases?

Ralith (approved):

PR #565, opened 2022-01-25
Iterator simplification

swooster:
This has a bunch of minor iterator-related spot refactors, primarily to reduce nesting of iterators.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -2099,22 +2099,15 @@ pub fn generate_feature<'a>(
         .filter_map(|cmd_ref| commands.get(&cmd_ref.name))
         .fold(
             (Vec::new(), Vec::new(), Vec::new(), Vec::new()),
-            |mut acc, &cmd_ref| {
-                match cmd_ref.function_type() {
-                    FunctionType::Static => {
-                        acc.0.push(cmd_ref);
-                    }
-                    FunctionType::Entry => {
-                        acc.1.push(cmd_ref);
-                    }
-                    FunctionType::Device => {
-                        acc.2.push(cmd_ref);
-                    }
-                    FunctionType::Instance => {
-                        acc.3.push(cmd_ref);
-                    }
-                }
-                acc
+            |mut accs, &cmd_ref| {
+                let acc = match cmd_ref.function_type() {
+                    FunctionType::Static => &mut accs.0,
+                    FunctionType::Entry => &mut accs.1,
+                    FunctionType::Device => &mut accs.2,
+                    FunctionType::Instance => &mut accs.3,
+                };
+                acc.push(cmd_ref);
+                accs
             },
         );
     let version = feature.version_string();
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -18,6 +18,21 @@ use std::fmt::Display;
 use std::path::Path;
 use syn::Ident;
 
+macro_rules! get_variant {
+    ($variant:path) => {
+        |enum_| match enum_ {
+            $variant(inner) => Some(inner),
+            _ => None,
+        }
+    };
+    ($variant:path { $($member:ident),+ }) => {
+        |enum_| match enum_ {
+            $variant { $($member),+, .. } => Some(( $($member),+ )),
+            _ => None,
+        }
+    };
+}
+
 const BACKWARDS_COMPATIBLE_ALIAS_COMMENT: &str = "Backwards-compatible alias containing a typo";
 
 pub trait ExtensionExt {}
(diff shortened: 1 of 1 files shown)


generator/src/lib.rs
```diff
@@ -17,6 +17,21 @@ use std::fmt::Display;
 use std::path::Path;
 use syn::Ident;
 
+macro_rules! match_variant {
```
swooster, line 20: I wasn't really sure what to call this... other names I thought of were `partial_match` or `extract_variant`.

generator/src/lib.rs
```diff
@@ -1362,22 +1360,21 @@ pub fn generate_enum<'a>(
...
         })
-    });
+        .map(|constant| {
+            let variant_name = constant.name.as_str();
+            let notation = constant.comment.as_deref().unwrap_or("");
+            let variant_ident = variant_ident(enum_.name.as_ref().unwrap(), variant_name);
+            quote! {
+                Self::#variant_ident => Some(#notation)
+            }
+        });
```
Ralith, line 1371: I'm not sure this is an improvement; we're replacing a single `filter_map` with a chained `filter_map` -> `map`. It only saves one line of code, and that can be attributed to different formatting of the `EnumsChild` match--in fact, the same syntax tweak would result in the original form being shorter still.

generator/src/lib.rs
```diff
@@ -2085,22 +2082,14 @@ pub fn generate_feature<'a>(
...
+        .filter_map(|feature| match feature {
+            vkxml::FeatureElement::Require(ref spec) => Some(spec),
+            _ => None,
+        })
+        .flat_map(|spec| spec.elements.iter())
+        .filter_map(|feature_spec| match feature_spec {
+            vkxml::FeatureReference::CommandReference(ref cmd_ref) => Some(cmd_ref),
+            _ => None,
         })
         .filter_map(|cmd_ref| commands.get(&cmd_ref.name))
```
Ralith, line 2079: Let's not have two `filter_map`s chained in a row.

generator/src/lib.rs
```diff
@@ -17,6 +17,21 @@ use std::fmt::Display;
 use std::path::Path;
 use syn::Ident;
 
+macro_rules! maybe_extract {
```
Ralith, line 20: Would `get_variant` be a clearer name for this?

Ralith (changes requested):

generator/src/lib.rs
swooster: I wasn't sure what name was clearest and had tried a bunch of different ones... `get_variant` sounds good to me; I'll switch to that.

generator/src/lib.rs
swooster: Part of this is that it sort of separated out a repeated pattern of `match x { X::Foo(y) => y, _ => None }`, which ended up making it clearer to me that a `get_variant` macro would be useful. Would it be preferable to avoid this intermediate commit?

generator/src/lib.rs
swooster: The same comment as below applies to this... the idea of `get_variant` hadn't occurred to me when I started, so I was initially focusing on trying to unnest things and break them into small clear repeated steps. I'm thinking of redoing the commits so that I define and use `get_variant` first, then unnest in a later commit (I was thinking that first shrinking the code with `get_variant` might also making it easier to see that the unnesting doesn't change behavior).

swooster: It hit me that maybe I should describe why splitting up the code this way appealed to me: It's easier for me to interpret XML-related iters if all intermediate steps are constrained to forms `.filter_map(get_variant!(Foo::BarVariant))` or `.flat_map(&foo.children)` and only the last step has custom logic. Doing that allows me to read the iterator definition as a path: "traverse into `BarVariant` elements, then traverse into their children, then... etc... until we've reached the elements we want to process, then we apply unique application logic to those elements"

swooster: I'll go ahead leave this change out of the latest PR.

swooster:
@Ralith: I've pushed a new version... I've renamed `maybe_extract!()` to `get_variant!()` and I've squashed the unnest/flatten/maybe_extract commits together because they tend to go together and the intermediate steps aren't always a clear win. If you preferred those separated, let me know and I'll undo that.

I've undone the change refactoring `generate_result()` per your review, but have left in the change to `generate_feature()` because that allowed collapsing the code down with `get_variant!()`. If you still feel strongly about `generate_feature()` even with `get_variant!()` applied, please let me know.

Also, I changed `.map(|union| union.name.as_str())` to `.map(|union_| union_.name.as_str())` to avoid a binding with a reserved name.

Ralith (approved):
Good stuff, thanks!

PR #568, opened 2022-01-27
Add helper wrappers for Vulkan core 1.3 `Instance` and `Device` functions

MarijnS95:
Depends on #563

Now the loaders are called so that `fp_1_3()` is available for `Entry` (not relevant because it's empty), `Instance` and `Device`.  Also implement Rust'y wrapper functions for all new commands that are added in Vulkan 1.3.

### TODO

- [x] Implement the remaining device functions;
- [x] Switch all documentation links to `1.3-extensions`: #569;
- [x] ~Perhaps~ Copy these functions back to the extensions that are missing right now: #570, #571, #572;
- [x] Wait for https://github.com/KhronosGroup/Vulkan-Docs/pull/1746;
- [ ] Separate PR: Improve builders to leverage our `Handle` trait.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -14,6 +14,7 @@ pub struct Device {
     pub(crate) device_fn_1_0: vk::DeviceFnV1_0,
     pub(crate) device_fn_1_1: vk::DeviceFnV1_1,
     pub(crate) device_fn_1_2: vk::DeviceFnV1_2,
+    pub(crate) device_fn_1_3: vk::DeviceFnV1_3,
 }
 
 impl Device {
@@ -28,11 +29,20 @@ impl Device {
             device_fn_1_0: vk::DeviceFnV1_0::load(load_fn),
             device_fn_1_1: vk::DeviceFnV1_1::load(load_fn),
             device_fn_1_2: vk::DeviceFnV1_2::load(load_fn),
+            device_fn_1_3: vk::DeviceFnV1_3::load(load_fn),
         }
     }
 
     pub fn handle(&self) -> vk::Device {
         self.handle
+    }
+}
+
+/// Vulkan core 1.3
+#[allow(non_camel_case_types)]
+impl Device {
+    pub fn fp_v1_3(&self) -> &vk::DeviceFnV1_3 {
+        &self.device_fn_1_3
     }
 }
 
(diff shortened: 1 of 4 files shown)


ash/src/device.rs
```diff
@@ -36,6 +38,243 @@ impl Device {
...
 }
 
+/// Vulkan core 1.3
+#[allow(non_camel_case_types)]
+impl Device {
+    pub fn fp_v1_3(&self) -> &vk::DeviceFnV1_3 {
+        &self.device_fn_1_3
+    }
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkCreatePrivateDataSlot.html>
```
filnet, line 48: should the doc point to version 1.3 ?

MarijnS95: I guess I'll correct that in the generator in #563 and find-replace it in manual code. The links are still valid though, they just redirect. Don't ask me why these nice isolated pages are behind a `-extensions` URL anyway...

ash/src/device.rs
```diff
@@ -44,6 +44,456 @@ impl Device {
...
+    pub unsafe fn cmd_bind_vertex_buffers2(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        first_binding: u32,
+        buffers: &[vk::Buffer],
+        offsets: &[vk::DeviceSize],
+        sizes: Option<&[vk::DeviceSize]>,
+        strides: Option<&[vk::DeviceSize]>,
+    ) {
+        assert_eq!(offsets.len(), buffers.len());
```
Ralith, line 320: Should these be `debug_assert_eq`?

Ralith (approved):

ash/src/device.rs
MarijnS95: We've always done this, though in the extensions (where these have been copy-pasted from verbatim). Not sure, we could shove the potential UB behind the functions already being `unsafe`, or relax the bounds so that everything is equal in size or longer than the `.len()` that is passed into the Vulkan function.

The other way around: what would be compelling reasons to _not_ have an assert here outside of debug?

Ralith: Just the principle of being a zero-cost unsafe binding. No strong opinion in the absence of it being an actual problem for someone, though.

PR #574, opened 2022-02-04
entry: Allow querying `enumerate_instance_extension_properties()` by layer name

pollend:
I think it would be useful to query by extension name. 

```
When pLayerName parameter is NULL, only extensions provided by the Vulkan implementation or by implicitly enabled layers are returned. When pLayerName is the name of a layer, the instance extensions provided by that layer are returned.
```

the extensions not explicitly provided will be listed even if they have not been used. you would query the available extensions and then look at those extensions from a predefined list.


MarijnS95:
Gah, I have to figure what this `chore` means some day... :(

Perhaps we should just update the `enumerate_instance_extension_properties` API to take an `Option<&CStr>`? That's a breaking change though.

Also, it looks like something went wrong in the commit title and PR title?

pollend:
> Gah, I have to figure what this `chore` means some day... :(
> 
> Perhaps we should just update the `enumerate_instance_extension_properties` API to take an `Option<&CStr>`? That's a breaking change though.
> 
> Also, it looks like something went wrong in the commit title and PR title?

I've been marking my work using this: https://www.conventionalcommits.org/en/v1.0.0/

I can change the commit when I get back home.

umm yea, I think it make sense to use another method so I don't break the orignal method. too bad you can't just have arguments with defaults.

Ralith:
> Perhaps we should just update the enumerate_instance_extension_properties API to take an Option<&CStr>?

+1 to this. It's breaking but not super disruptively so, and I don't think anyone expects ash to be semver-stable for long periods at the moment anyway.

ash/src/entry.rs
```diff
@@ -238,11 +238,12 @@ impl Entry {
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkEnumerateInstanceExtensionProperties.html>
     pub fn enumerate_instance_extension_properties(
         &self,
+        p_layer_name: *const c_char
```
Ralith, line 241:
```suggestion
        p_layer_name: Option<&CStr>
```

Ralith (changes requested):

MarijnS95:
> > Gah, I have to figure what this `chore` means some day... :(
> > Perhaps we should just update the `enumerate_instance_extension_properties` API to take an `Option<&CStr>`? That's a breaking change though.
> > Also, it looks like something went wrong in the commit title and PR title?
> 
> I've been marking my work using this: https://www.conventionalcommits.org/en/v1.0.0/

The main point being that `chore` is usually a (mundane) change outside of (public) code: updating version/lock files, changing things in gitattributes/gitmodules, and the like; meta stuff if you wish. This specific prefix doesn't seem documented at that website, they refer to some "Angular" docs where it is not referenced either.

This change is not that. Perhaps it's a fix because we're adding a method that's missing? Perhaps it's a feature because of giving the new "feature" possibility of using it this way?

> umm yea, I think it make sense to use another method so I don't break the orignal method. too bad you can't just have arguments with defaults.

Like @Ralith said - and you've already changed nicely - a breaking change is preferable here. I'll have to think about merge order as we don't have anything breaking right now and I'm not fond of making tons of backport branches, but we can probably come up with a reason to make another breaking release soon anyway.

ash/src/entry.rs
```diff
@@ -238,11 +238,18 @@ impl Entry {
...
     ) -> VkResult<Vec<vk::ExtensionProperties>> {
         unsafe {
             read_into_uninitialized_vector(|count, data| {
-                self.entry_fn_1_0
-                    .enumerate_instance_extension_properties(ptr::null(), count, data)
+                self.entry_fn_1_0.enumerate_instance_extension_properties(
+                    match p_layer_name {
+                        None => ptr::null(),
+                        Some(str) => str.as_ptr(),
+                    },
```
MarijnS95, line 249: We have `as_raw_ptr` for this:

https://github.com/MaikKlein/ash/blob/bb02b203391d77e30944f4fc90d7a446474de4c7/ash/src/lib.rs#L55-L67
```suggestion
                    layer_name.as_raw_ptr(),
```

ash/src/entry.rs
```diff
@@ -238,11 +238,18 @@ impl Entry {
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkEnumerateInstanceExtensionProperties.html>
     pub fn enumerate_instance_extension_properties(
         &self,
+        p_layer_name: Option<&CStr>,
```
MarijnS95, line 241:
```suggestion
        layer_name: Option<&CStr>,
```

We don't commonly use the `p_` prefix for "safe" rusty items.

MarijnS95 (changes requested):

ash/src/entry.rs
MarijnS95: Welp, thinking about this again this `as_ptr()` is different from a regular borrow of course!

Unfortunately there exists no `.map_or_default()`, but I wonder if the rest of the code uses `.map_or(Default::default(), |s| s.as_ptr())` elsewhere too?

MarijnS95:
@pollend I updated the title of this PR (what will show up in the patch) as I have no idea what `entry_name` nor "extension name" have to do with these changes. Let me know if this is correct.

Please also add this to the changelog, then we can merge your patch in time for the breaking 0.36 release somewhere in the - hopefully - next 24 hours.

MarijnS95 (approved):
Thanks, we're all set now!

PR #579, opened 2022-02-16
extensions/khr: Add VK_KHR_external_memory_win32

MarijnS95:


ash/src/extensions/khr/external_memory_win32.rs
```diff
@@ -0,0 +1,63 @@
...
+        });
+        Self { handle, fp }
+    }
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetMemoryWin32HandleKHR.html>
+    pub unsafe fn get_memory_win32_handle(
+        &self,
+        create_info: &vk::MemoryGetWin32HandleInfoKHR,
+    ) -> VkResult<vk::HANDLE> {
+        let mut handle = ptr::null_mut();
```
MarijnS95, line 29: Fwiw this should probably be `INVALID_HANDLE_VALUE` or `-1`?

MarijnS95: Anyone opposed to me adding a `pub const INVALID_HANDLE_VALUE: HANDLE = -1isize as *mut c_void;` to `platform_types.rs`?

Ralith: In what circumstance is the initial value here observable?

MarijnS95: The Vulkan function could observe it (even though it shouldn't). Theoretically it could also not update it while returning `vk::Result::SUCCESS`.

Probably chasing ghosts here.

Ralith: I think that would constitute a violation of the spec.

MarijnS95: Fair enough. Let's get it merged then.

Ralith (approved):

PR #585, opened 2022-02-19
Release ash 0.36.0 and ash-window 0.9.1

MarijnS95:


ash-window/Cargo.toml
```diff
@@ -14,15 +14,15 @@ exclude = [".github/*"]
 workspace = ".."
 
 [dependencies]
-ash = { path = "../ash", version = "0.35", default-features = false }
+ash = { path = "../ash", version = ">=0.34, <=0.36", default-features = false }
```
MarijnS95, line 17: As discussed earlier I'd like to test-drive a range selection for `ash`, such that we don't have to perform as many breaking releases for `ash-window` when nothing else changes. I checked backwards compatibility and a change inflicted by `ash` last happened in 0.34. As such, this crate compiles fine with `ash=0.34` all the way up to `0.36`.

This was tested by removing `ash-window` out of the workspace temporarily (get some strange `rmeta` duplicate errors otherwise) and dropping the `path="../ash"` here to use the crates release.
Then I alternated the version of the example below (which is how users of `ash`/`ash-window` select their version of `ash`), which implicitly dictates the version that is used for `ash-window` as well (as long as it is within the allowed range, of course).

Now, when we release a new `ash` version that is compatible with `ash-window` without further changes, we can just perform a non-breaking patch release that extends the range with a newer version.

Sounds sensible/useful?

ash-window/Cargo.toml
```diff
@@ -14,15 +14,15 @@ exclude = [".github/*"]
...
 raw-window-handle = "0.3.4"
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 raw-window-metal = "0.1"
 
 [dev-dependencies]
 winit = "0.19.4"
-ash = { path = "../ash", version = "0.35", default-features = false, features = ["linked"] }
+ash = { path = "../ash", version = ">=0.35, <=0.36", default-features = false, features = ["linked"] }
```
MarijnS95, line 25: Note though that we renamed `Entry::new()` to `Entry::linked()` in `0.35`, as such this example isn't compatible with `0.34`. Perhaps we can also just keep a fixed `0.36` in here though...

ash-window/Cargo.toml
MaikKlein: :thinking: 
What is the usecase? I assume people can do `ash-window = "0.10"` and then they don't have to bump the version because we release patch version. I am not sure how useful this will be in practice.

> As discussed earlier I'd like to test-drive a range selection for ash

Can you share a link to the discussion?

MarijnS95: @MaikKlein You self-answered the question, it is indeed as I wrote above:

> such that we don't have to perform as many breaking releases for `ash-window` when nothing else changes

> Can you share a link to the discussion?

In hindsight I don't think we've truly discussed this, more of an off-hand mention that I wanted to try this some day. It might not even have been on this repo, perhaps on gitter or a different project entirely.

MarijnS95: That one-sided discussion might have been https://github.com/rust-windowing/android-ndk-rs/issues/211#issuecomment-1014707079.

ash-window/Cargo.toml
Ralith: I think a fixed version for the examples makes sense from a maintenance perspective.

Ralith (approved):

ash-window/Cargo.toml
MarijnS95: I presume that - and the approval below - is a yay for trying this on `ash-window`'s `ash` dependency though?

MarijnS95:
Fwiw I think we can already keep `ash-window` at `0.9.1`. It's currently `version="0.35"` which is extendable with `0.36`.

MaikKlein (approved):

ash-window/Cargo.toml
Ralith: Yeah, I'm all in favor of reduced toil. We'll have to be careful not to introduce new idioms, especially without CI coverage, but I think the risk is low considering the lib's narrow scope.

MarijnS95: Cool, let's see how this works out over time. If mistakes happen we can always re-evaluate and most likely come up with a simple CI test that ensures compatibility.

PR #590, opened 2022-02-26
ash-window: Make `enumerate_required_extensions` return `&[*const c_char]`

swooster:
This PR makes `enumerate_required_extensions` return the same type that is expected by `vk::InstanceCreateInfoBuilder::enabled_extension_names` to slightly reduce the boilerplate needed to write small Vulkan apps.

### DIFF
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -1,8 +1,9 @@
 #![warn(trivial_casts, trivial_numeric_casts)]
 
+use std::os::raw::c_char;
+
 use ash::{extensions::khr, prelude::*, vk, Entry, Instance};
 use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
-use std::ffi::CStr;
 
 #[cfg(any(target_os = "macos", target_os = "ios"))]
 use ash::extensions::ext; // portability extensions
@@ -122,10 +123,16 @@ pub unsafe fn create_surface(
 /// The returned extensions will include all extension dependencies.
 pub fn enumerate_required_extensions(
     window_handle: &dyn HasRawWindowHandle,
-) -> VkResult<Vec<&'static CStr>> {
+) -> VkResult<&'static [*const c_char]> {
     let extensions = match window_handle.raw_window_handle() {
         #[cfg(target_os = "windows")]
-        RawWindowHandle::Windows(_) => vec![khr::Surface::name(), khr::Win32Surface::name()],
+        RawWindowHandle::Windows(_) => {
+            const WINDOWS_EXTS: [*const c_char; 2] = [
+                khr::Surface::name().as_ptr(),
+                khr::Win32Surface::name().as_ptr(),
+            ];
+            &WINDOWS_EXTS
+        }
 
         #[cfg(any(
             target_os = "linux",
@@ -134,7 +141,13 @@ pub fn enumerate_required_extensions(
             target_os = "netbsd",
             target_os = "openbsd"
         ))]
-        RawWindowHandle::Wayland(_) => vec![khr::Surface::name(), khr::WaylandSurface::name()],
+        RawWindowHandle::Wayland(_) => {
+            const WAYLAND_EXTS: [*const c_char; 2] = [
+                khr::Surface::name().as_ptr(),
+                khr::WaylandSurface::name().as_ptr(),
+            ];
+            &WAYLAND_EXTS
+        }
 
         #[cfg(any(
             target_os = "linux",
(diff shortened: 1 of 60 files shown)


swooster:
Currently, the mismatch between what is returned by [`enumerate_required_extensions()`](https://docs.rs/ash-window/latest/ash_window/fn.enumerate_required_extensions.html) and what is taken by [`vk::InstanceCreateInfoBuilder::enabled_extension_names()`](https://docs.rs/ash/latest/ash/vk/struct.InstanceCreateInfoBuilder.html#method.enabled_extension_names) requires all Vulkan apps to write a bit of extra boilerplate to remap the list of extensions:
```rust
let extensions: Vec<_> = extensions
    .iter()
    .map(|ext| ext.as_ptr())
    .collect();
```
This PR will make that unnecessary.

swooster:
Note: I don't think I've touched any of the areas with clippy warnings.

ash-window/src/lib.rs
```diff
@@ -122,10 +123,16 @@ pub unsafe fn create_surface(
...
+) -> VkResult<&'static [*const c_char]> {
     let extensions = match window_handle.raw_window_handle() {
         #[cfg(target_os = "windows")]
-        RawWindowHandle::Windows(_) => vec![khr::Surface::name(), khr::Win32Surface::name()],
+        RawWindowHandle::Windows(_) => {
+            const WINDOWS_EXTS: [*const c_char; 2] = [
+                khr::Surface::name().as_ptr(),
+                khr::Win32Surface::name().as_ptr(),
+            ];
+            &WINDOWS_EXTS
```
Ralith, line 134: I don't think it's actually necessary to have a separate `const` declaration, so we could avoid that boilerplate here and below. `'static` promotion applies regardless.

Ralith (commented):
Thanks, I've been wanting this for ages!

ash-window/src/lib.rs
swooster: Is there any chance you could elaborate on what you have in mind? I've been surprised in the past that Rust was smart enough to promote `&()` into a `&'static ()` but I had trouble getting that to work here. I tried this but Rust interprets it as a temporary value:

```rust
RawWindowHandle::Windows(_) => &[
    khr::Surface::name().as_ptr(),
    khr::Win32Surface::name().as_ptr(),
],
```

Is there another approach I should take?

Ralith (approved):

Ralith:
It should be noted that this will require ash-window's ash dependency to be collapsed down to the latest version once released.

swooster:
(rebasing onto latest commit)

Ralith:
Just holding off on merging until @MarijnS95 has a look since it's breaking.

MarijnS95 (approved):
I've been hesitant to do this in the past because of not having lifetime annotations on `*const c_char`, but these are anyway all `'static` and there's no owned `Vec<>` being returned anymore (separate, still valid change) which users could accidentally append their own pointers with different lifetime to (this is a problem I vaguely recall having in our own application surrounding extension names, but not sure how since these would have anyway been `static` string literals :thinking: ).

As such, this change seems totally fine and sensible to me.

Can we make `fn` `const` as a non-breaking patch release?

swooster:
> Can we make `fn` `const` as a non-breaking patch release?

This PR makes the names const in a separate commit from the breaking change specifically to allow that.

MarijnS95:
@swooster We're still squash-merging here, so that'd have little effect unless the commit is transplanted to a separate PR (or #594 is reopened, retitled, and merged).

There are currently anyway very few changes (both on `master` and the open-PR-pipeline) that are non-breaking and would be interesting for a patch release. On the contrary, we have about 2-4 breaking changes queued up, so we might as well shoot for `ash 0.37` breaking release straight away.

MarijnS95:
Rebased to regenerate against `1.3.208` :)

MarijnS95:
@swooster Are you aware that your email address used in the commits isn't associated with your GitHub account?

swooster:
> @swooster Are you aware that your email address used in the commits isn't associated with your GitHub account?

If the only difference between the two is that my github account is `(my username)+github@gmail.com` and my commit is `(my username)@gmail.com` then that's ok. If they're substantially different, then that'd be a bad sign.

I've gone ahead and added `(my username)@gmail.com` to my github account.

MarijnS95:
@swooster it seems to be your initials + surname, separated with dots, at gmail dot com. See the `Co-authored-by:` field that GitHub generated for the merge commit :wink:

PR #591, opened 2022-02-26
Appease clippy

swooster:


swooster:
Note: The first commit is essentially the same as #565 as warnings are what initially spurred me to attempt the iterator simplification PR. If this PR goes in, I'll rebase #565 and remove its first commit.

Ralith (approved):

PR #594, opened 2022-03-13
Make names const

Neo-Zhixing:
`std::ffi::CStr::from_bytes_with_nul_unchecked` was const stabilized in [Rust 1.59](https://github.com/rust-lang/rust/blob/21b0325c68421b00c6c91055ac330bd5ffe1ea6b/RELEASES.md?plain=1#L74), meaning that the `name()` functions in extensions can now be `const` as well.

Alternatively we can do away with `name()` functions and use associated consts instead. However this would be a breaking change.

```rs
const NAME: &'static CStr = xxxxxx;
```

### DIFF
--- a/ash/src/extensions/ext/buffer_device_address.rs
+++ b/ash/src/extensions/ext/buffer_device_address.rs
@@ -26,7 +26,7 @@ impl BufferDeviceAddress {
         self.fp.get_buffer_device_address_ext(self.handle, info)
     }
 
-    pub fn name() -> &'static CStr {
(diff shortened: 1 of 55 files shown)


Ralith:
Is this redundant to https://github.com/MaikKlein/ash/pull/590?

Neo-Zhixing:
Oops, didn't see that #590 also did this. Closing now.

MarijnS95:
I'd be happy to land both changes as separate PRs, tbh.

PR #596, opened 2022-03-14
Fix Rust 1.61 unsafe in unsafe warning in Entry::from_static_fn

Neo-Zhixing:
```
warning: unnecessary `unsafe` block
   --> ash\src\entry.rs:146:47
    |
145 |     pub unsafe fn from_static_fn(static_fn: vk::StaticFn) -> Self {
    |     ------------------------------------------------------------- because it's nested under this `unsafe` fn
146 |         let load_fn = |name: &std::ffi::CStr| unsafe {
    |                                               ^^^^^^ unnecessary `unsafe` block
    |
    = note: `#[warn(unused_unsafe)]` on by default
```

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -143,7 +143,7 @@ impl Entry {
     /// `static_fn` must contain valid function pointers that comply with the semantics specified by
     /// Vulkan 1.0, which must remain valid for at least the lifetime of the returned [`Entry`].
     pub unsafe fn from_static_fn(static_fn: vk::StaticFn) -> Self {
-        let load_fn = |name: &std::ffi::CStr| unsafe {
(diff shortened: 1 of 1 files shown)


Ralith (approved):

MarijnS95 (approved):

MarijnS95:
Seems fine to me to suppress the warning for now, but IIRC @Ralith brought up recently to enable `unsafe_block_in_unsafe_fn` (`-Wunsafe-op-in-unsafe-fn`) to explicitly require `unsafe {}` blocks even inside `unsafe fn` since both serve a fundamentally different purpose (marking a function `unsafe` to call, and using `unsafe` functionality within this function).

Couldn't find the relevant discussion here in 30 seconds of searching though.

PR #598, opened 2022-03-17
Add const impls for common bitflag ops

Neo-Zhixing:
This PR added a few const implementations for common bitwise operations for the `vk_bitflags_wrapped` macro.

They're currently implemented as operator loads, making them non-const.


MarijnS95:
https://github.com/MaikKlein/ash/pull/549#discussion_r780719676

PR #599, opened 2022-03-22
Omit wrapper functions on Fp structs

Ralith:
These wrappers contributed thousands of lines of code but offered insignificant ergonomic benefit as the same functions are also wrapped at a higher level and, if necessary, wrapper functions can be called directly. This will also make the doc search much less confusing.

No obvious impact to compile time on my machine, but dropping several thousand lines of code can't hurt.


MarijnS95:
I've been longing for this change for quite some time, glad to see it finally happen!

Disappointing it doesn't contribute anything meaningful to compile times, though :/

MarijnS95:
Don't forget to add an entry to the changelog, then I think we have enough material (in terms of open PRs) for the next breaking `ash` release :)

Ralith:
Changelog updated. We really do just need to work out how to profile the compiler rather than relying on guesswork.

MarijnS95 (approved):
Haven't checked all of the ~7k line changes in detail, most of it looks good though.

PR #602, opened 2022-03-24
Replace builders with lifetimes/setters directly on Vulkan structs

Ralith:
Based on prior discussion in https://github.com/MaikKlein/ash/issues/441 and https://github.com/MaikKlein/ash/pull/483.

Remarkably, this doesn't seem to significantly impact build times despite deleting on the order of 20k(!) lines of code. Arguably still an improvement for the sake of simplicity and safety, though. Could use confirmation on the build impact by someone with a better test environment.

TODO: Try making all the setters by `&mut` rather than by value, per discussion in #483.


MarijnS95:
> Remarkably, this doesn't seem to significantly impact build times

This may have been the `#[inline]` change pushed later, but this PR in its current form nets me ~1.3 seconds:

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.113 s ±  0.062 s    [User: 7.927 s, System: 0.405 s]
  Range (min … max):    7.053 s …  7.244 s    10 runs
```
```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      5.812 s ±  0.036 s    [User: 6.644 s, System: 0.356 s]
  Range (min … max):    5.725 s …  5.862 s    10 runs
```

I recall discussing earlier that the builder structs themselves likely weren't the problem, but it was down to lifetime propagation on the setter methods. Can't find that discussion anymore, but I'd love to get this in and replace trivial setters with just member updates or struct initialization, as previously discussed and proposed in #441.

generator/src/lib.rs
```diff
@@ -2323,8 +2327,14 @@ pub fn generate_aliases_of_types(
...
-            let tokens = quote! {
-                pub type #name_ident = #alias_ident;
+            let tokens = if has_lifetimes.contains(&alias_ident) {
+                quote! {
+                    pub type #name_ident<'a> = #alias_ident<'a>;
+                }
+            } else {
+                quote! {
+                    pub type #name_ident = #alias_ident;
+                }
```
MarijnS95, line 2337: Nit, I prefer to use `quote!` with parentheses instead of curly brackets - and put them on a single line - if the lines are short.

```suggestion
                quote!(pub type #name_ident<'a> = #alias_ident<'a>;)
            } else {
                quote!(pub type #name_ident = #alias_ident;)
```

Might look weird with the semicolons though?

generator/src/lib.rs
```diff
@@ -1916,14 +1900,24 @@ pub fn generate_struct(
...
+            quote!(#ty #lifetime)
         };
         quote! {pub #param_ident: #param_ty_tokens}
     });
 
-    let debug_tokens = derive_debug(_struct, union_types);
-    let default_tokens = derive_default(_struct);
-    let setter_tokens = derive_setters(_struct, root_structs);
+    let has_lifetime = has_lifetimes.contains(&name);
+    let (generics, marker) = match has_lifetime {
```
MarijnS95, line 1913: Bit odd to call it `generics` in one place, `lifetime` in others (there are no generic types involved in this instance too?).

generator/src/lib.rs
```diff
@@ -1763,56 +1784,18 @@ pub fn derive_setters(
...
             #(#setters)*
 
             #next_function
-
-            /// Calling build will **discard** all the lifetime information. Only call this if
-            /// necessary! Builders implement `Deref` targeting their corresponding Vulkan struct,
-            /// so references to builders can be passed directly to Vulkan functions.
-            pub fn build(self) -> #name {
-                self.inner
-            }
```
MarijnS95, line 1815: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes:

generator/src/lib.rs
```diff
@@ -1763,56 +1784,18 @@ pub fn derive_setters(
         .map(|extends| format_ident!("Extends{}", name_to_tokens(extends)))
         .map(|extends| {
             quote! {
-                unsafe impl #extends for #name_builder<'_> {}
-                unsafe impl #extends for #name {}
+                unsafe impl #lifetime #extends for #name #lifetime {}
```
MarijnS95, line 1787: Couldn't this stay as-is or am I not knowing something about lifetimes in trait impls?


```suggestion
                unsafe impl #extends for #name<'_> {}
```

Ralith:
> This may have been the #[inline] change pushed later, but this PR in its current form nets me ~1.3 seconds:

Nice! Can you verify whether it was the inlining or not? If so, we definitely can/should pursue it further. I don't expect that to have mattered, since I believe codegen still occurs for inline stuff (so it can only add work), but there's some conjecture there.

> EDIT: Keeping in mind the twitter link at https://github.com/MaikKlein/ash/issues/441#issuecomment-956073795, would be nice to also drop setters for pointers (Rust borrows with a proper lifetime now) and "trivial slices" with a length (unique!) + pointer pair.

Per previous discussion I don't believe it's possible to avoid setters for any slice case due to ABI concerns. I'm also a bit leery of requiring a mixture of different approaches to initialize a struct; let's leave that for future work.

MarijnS95:
```
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      6.251 s ±  0.032 s    [User: 7.082 s, System: 0.387 s]
  Range (min … max):    6.176 s …  6.295 s    10 runs
```

Seems to be both, with your initial approach netting about `~800ms` and the `#[inline]` addition another `~400ms`.

> Per previous discussion I don't believe it's possible to avoid setters for any slice case due to ABI concerns. I'm also a bit leery of requiring a mixture of different approaches to initialize a struct; let's leave that for future work.

Whoops, I meant to write the `VulkanSlice1` wrapper type. Still needs a setter and/or `Into`/`new()` construction pattern, but could also expose a safe iterator. Perhaps that type is a bit neater to initialize and use than having a `pointer_to_first_element: &'a vk::Something` (or doing away with lifetime safety and keeping it `*const`). Let's indeed leave both the discussion and implementation for a different place though ^^

MarijnS95 (approved):

PR #603, opened 2022-03-27
Add VK_EXT_image_drm_format_modifier

i509VCB:


Changelog.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Added
+
+- Added `VK_EXT_image_drm_format_modifier` (#603)
```
MarijnS95, line 11:
```suggestion
- Added `VK_EXT_image_drm_format_modifier` device extension (#603)
```

ash/src/extensions/ext/drm_format_modifier.rs
```diff
@@ -0,0 +1,83 @@
...
+use std::ffi::CStr;
+use std::mem;
+
+#[derive(Clone)]
+pub struct DrmFormatModifier {
+    handle: vk::Device,
+    fp: vk::ExtImageDrmFormatModifierFn,
+}
+
+impl DrmFormatModifier {
```
MarijnS95, line 13:
```suggestion
/// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_EXT_image_drm_format_modifier.html>

impl DrmFormatModifier {
```

ash/src/extensions/ext/drm_format_modifier.rs
```diff
@@ -0,0 +1,83 @@
+use crate::prelude::*;
```
MarijnS95, line 1: Shouldn't this file be named `image_drm_format_modifier`? As a consequence, `ImageDrmFormatModifier` on the struct.

ash/src/extensions/ext/drm_format_modifier.rs
```diff
@@ -0,0 +1,83 @@
...
+        Self { handle, fp }
+    }
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetImageDrmFormatModifierPropertiesEXT.html>
+    pub unsafe fn get_image_drm_format_modifier_properties(
+        &self,
+        image: &vk::Image,
+        properties: &mut vk::ImageDrmFormatModifierPropertiesEXT,
+    ) -> VkResult<()> {
+        (self.fp.get_image_drm_format_modifier_properties_ext)(self.handle, *image, properties)
```
MarijnS95, line 28: Handles are copy types, and as such always passed by value:

```suggestion
        image: vk::Image,
        properties: &mut vk::ImageDrmFormatModifierPropertiesEXT,
    ) -> VkResult<()> {
        (self.fp.get_image_drm_format_modifier_properties_ext)(self.handle, image, properties)
```

ash/src/extensions/ext/drm_format_modifier.rs
```diff
@@ -0,0 +1,83 @@
...
+        format: vk::Format,
+        format_properties: vk::FormatProperties,
+    ) -> Vec<vk::DrmFormatModifierPropertiesEXT> {
+        let mut list = vk::DrmFormatModifierPropertiesListEXT::default();
+
+        // Get count of DRM format modifiers for the format.
+        {
+            let mut format_properties_2 = vk::FormatProperties2::builder()
+                .format_properties(format_properties)
+                .push_next(&mut list);
```
MarijnS95, line 44: Just like other typical list getter functions, how will you let the caller append more `pNext` structs to `vk::DrmFormatModifierPropertiesListEXT`?

ash/src/extensions/ext/drm_format_modifier.rs
```diff
@@ -0,0 +1,83 @@
...
+
+            instance.get_physical_device_format_properties2(
+                pdevice,
+                format,
+                &mut format_properties_2,
+            );
+        }
+
+        let mut data = Vec::with_capacity(list.drm_format_modifier_count as usize);
+        list.p_drm_format_modifier_properties = data.as_mut_ptr();
```
MarijnS95, line 54: Note that we have helper functions for this pattern. It may be a little less trivial to use thanks to the extra struct holding a count and pointer, but it should suffice.

ash/src/extensions/ext/drm_format_modifier.rs
```diff
@@ -0,0 +1,83 @@
...
+    ) -> Vec<vk::DrmFormatModifierPropertiesEXT> {
+        let mut list = vk::DrmFormatModifierPropertiesListEXT::default();
+
+        // Get count of DRM format modifiers for the format.
+        {
+            let mut format_properties_2 = vk::FormatProperties2::builder()
+                .format_properties(format_properties)
+                .push_next(&mut list);
+
+            instance.get_physical_device_format_properties2(
```
MarijnS95, line 46: This extension requires Vulkan 1.0 yet this function was only stabilized in 1.1. How are you going to communicate this requirement to the caller, or otherwise fall back to the extension (presuming the caller enabled it in the first place...) if the device supports only Vulkan 1.0?

FWIW this function needs a doc-comment anyway.

MarijnS95 (changes requested):
When resubmitting this, can you also retitle the commit to `extensions/ext: Add VK_EXT_image_drm_format_modifier`?

ash/src/extensions/ext/drm_format_modifier.rs
i509VCB: Seems kinda odd to put it above the `impl` statement. I would assume the struct would be a better spot?

MarijnS95: I think I managed to break github :)

ash/src/extensions/ext/drm_format_modifier.rs
```diff
@@ -0,0 +1,83 @@
+use crate::prelude::*;
+use crate::vk;
+use crate::{Device, Instance};
+use std::ffi::CStr;
+use std::mem;
+
+#[derive(Clone)]
+pub struct DrmFormatModifier {
```
MarijnS95, line 8:
```suggestion
/// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_EXT_image_drm_format_modifier.html>
#[derive(Clone)]
pub struct DrmFormatModifier {
```

MarijnS95 (changes requested):

ash/src/extensions/ext/drm_format_modifier.rs
i509VCB: I can't pass `&mut FormatProperties2` into the function since the `push_next` function does not exist on those structs.

I could emulate it via copying the content of FormatProperties2Builder::push_next, but that does not sound very maintainable.

Taking a builder feels odd but would be least prone to mistakes.

ash/src/extensions/ext/drm_format_modifier.rs
i509VCB: The one solution I could think of for Vulkan 1.0 is essentially duplicating this function inside an `impl GetPhysicalDeviceProperties2 { ... }`

ash/src/extensions/ext/drm_format_modifier.rs
i509VCB: This might actually be quite a challenge to use the helper functions:

1. Since `DrmFormatModifierPropertiesListEXT` is mutably borrowed inside `FormatProperties2`, we end up double borrowing when trying to set pointer to the vec inside `DrmFormatModifierPropertiesListEXT` during the second roundtrip after the function has given us the number of drm format modifier properties structs the implementation has to give us.

2. Even if we make the following change to `read_into_uninitialized_vector`, the double borrow issue is still there.:
```diff
pub(crate) unsafe fn read_into_uninitialized_vector<N: Copy + Default + TryInto<usize>, T>(
-    f: impl Fn(&mut N, *mut T) -> vk::Result,
+    mut f: impl FnMut(&mut N, *mut T) -> vk::Result,
) -> VkResult<Vec<T>>
```

ash/src/extensions/ext/drm_format_modifier.rs
i509VCB: It appears in experimentation that even if I take ownership of a `FormatProperties2Builder`, I end up with some of the borrowing issues I have described in the comment about using the helper functions.

i509VCB: The recent changes to remove builders have made this possible now.

i509VCB:
boop

i509VCB:
@MarijnS95 soft reminder that I have resolved the review comments.

MarijnS95:
Regardless of previous review points, ash so far has been completely unopinionated in terms of not providing any helper functions, and `get_drm_format_properties_list` appears to be the first to break that "contract".

In addition the function signature and implementation make it so that I don't see any real advantage over having it, in favour of letting users implement it directly themselves.

@Ralith What's your though on this? My feeling is that we can probably merge this PR straight away without it, and I'm pretty opposed to having it.

i509VCB:
The primary reason for the helper is very much for convince. Since the return value is effectively an array of variable size, getting a Vec makes sense.

The readme does agree with you two regarding the opinionated function.
> Note: Functions don't return Vec<T> if this would limit the functionality. See p_next.

Since `get_drm_format_properties_list` requires extension via the p_next field and may in fact be influenced by other p_next members, it is quite hard to provide that utility in a way that isn't opinionated. 

There is some precedence for helpers like those in PhysicalDeviceDrm for getting the physical device property but that does not use the same commands as this case and does not need two implementations either.

i509VCB:
A further question, if the user is expected to implement that part of the api since it is opinionated, should the helper functions in ash for things such as uninitialized vectors be publicly available?

MarijnS95:
> The readme does agree with you two regarding the opinionated function.
> 
> > Note: Functions don't return Vec if this would limit the functionality. See p_next.
> 
> Since `get_drm_format_properties_list` requires extension via the p_next field and may in fact be influenced by other p_next members, it is quite hard to provide that utility in a way that isn't opinionated.

This note specifically applies to the _return type_; `vk::DrmFormatModifierPropertiesEXT` does not have a `p_next` field and is as such fine to be returned in a `Vec` rather than through a mutable slice that is to be filled by the called like the `vk::FormatProperties2` bit that you subsequently have to modify. That is by the way the part which makes this weird, because it's mainly mutable to let them influence the `p_next` chain in which you insert another object, and perhaps retrieve the fields in here at the same time. 

Opinionated here can be interpreted in two ways:
1. The choice to not provide any helpers that do not 1:1 map to a single Vulkan function and/or restrict functionality of that function;
2. Helpers that have a specific or peculiar way to be used, the way you described.

> There is some precedence for helpers like those in PhysicalDeviceDrm for getting the physical device property but that does not use the same commands as this case and does not need two implementations either.

This is a very simple helper that was apparently introduced in 0ed0a06634e86e77cbfd30b9e13d60ff03682071 and I've adopted it in the KHR Ray Tracing and later `PhysicalDeviceDrm` extension without thinking much of it. In hindsight I'd rather get rid of it since it's probably more efficient to just chain all necessary properties into a single `vkGetPhysicalDeviceProperties2()` call at once.

> A further question, if the user is expected to implement that part of the api since it is opinionated, should the helper functions in ash for things such as uninitialized vectors be publicly available?

Not sure, it's mostly been an internal helper for now. At the same time it doesn't _really_ fit these functions. It's great that it automagically calls the function twice, once with a nullptr to acquire the length, but this particular use of  `get_physical_device_format_properties2()` doesn't return errors that require a retry (what I originally added it for) nor have the pointer/count in a place that makes it convenient to use.

i509VCB:
I have removed `get_drm_format_properties_list` then. It's not a whole lot to implement getting the list yourself. The use cases for that function being in ash is arguably quite niche (Wayland compositors, implementing WSI, KMS instead of using VkDisplayKHR).

MarijnS95 (approved):

PR #607, opened 2022-04-01
Remove experimental AMD extensions.

msiglreith:
The extensions haven't being promoted to actual extensions within a long time frame and don't provide any value in its current state. I added them some long time ago but not using them anymore since I switched my IHV. As I guess noone else has use for them I would propose to just drop them.


Ralith:
Seems reasonable to me, these look pretty niche and they cost maintenance effort.

Changelog.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Update Vulkan-Headers to 1.3.210 (#605)
 
+### Removed
+
+- Removed experimental AMD extensions.
```
MarijnS95, line 15:
```suggestion
- Removed experimental AMD extensions (#607)
```

MarijnS95 (approved):
I can definitely sympathize with that. This extension has been a "pain" at times to update due to having code that mimicks what is autogenerated.

If anyone wants this extension back there should be some effort to write out the types and commands in "`vk.xml` format" and make that mergeable in the generator.

MarijnS95:
Thanks!

PR #609, opened 2022-04-06
Fix instructions for running the examples on a Mac.

jfgoog:
Tested with version 1.3.204.1.

The following displays the triangle example on my Mac:

```
VULKAN_SDK=$HOME/VulkanSDK/1.3.204.1/macOS \
DYLD_FALLBACK_LIBRARY_PATH=$VULKAN_SDK/lib \
VK_LAYER_PATH=$VULKAN_SDK/share/vulkan/explicit_layer.d \
VK_ICD_FILENAMES=$VULKAN_SDK/share/vulkan/icd.d/MoltenVK_icd.json \
cargo run --bin triangle
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,11 +231,12 @@ Make sure that you have a Vulkan ready driver and install the [LunarG Vulkan SDK
 #### Linux
 Make sure that you have a Vulkan ready driver and install the [LunarG Vulkan SDK](https://lunarg.com/vulkan-sdk/). You also have to add the library and layers to your path. Have a look at my [post](http://askubuntu.com/a/803110/77183) if you are unsure how to do that.
 #### macOS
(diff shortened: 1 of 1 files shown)


MarijnS95 (commented):
Thanks a lot for doing this! Basically, if I understand the changes correctly, all you're doing here is:

- Give the user a sensible value for `VULKAN_SDK=$HOME/VulkanSDK/<version>/macOS` (was `SDK_PATH` previously);
- ICD and layers moved from `etc` to `share`.

Furthermore, do you think it's advantageous to stuff these all in a single ` ``` ` codeblock (probably with `sh` type, backslashes at the end for the use proposed in the PR description, or `export` to put them in the current environment)? That ought to be easier to copy-paste in markdown (it seems to be fine when rendered though).

MarijnS95:
> and it is necessary to set VULKAN_SDK, which didn't seem to be necessary before.

Ah okay, thanks for clarifying!

MarijnS95 (approved):

PR #611, opened 2022-04-11
Add platform-specific extension list for creating a surface to ash_window 

iMplode-nZ:
This pull request adds a constant `TARGET_EXTENSIONS` to ash_window, which allows people to create the `Instance` before the window by using it instead of `enumerate_required_extensions`. (see #595)

### DIFF
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -119,6 +119,7 @@ pub unsafe fn create_surface(
 }
 
 #[cfg(target_os = "windows")]
+/// Extensions necessary for creating a surface on the target platform.
 pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
     khr::Surface::name().as_ptr(),
     khr::Win32Surface::name().as_ptr(),
@@ -130,6 +131,7 @@ pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
     target_os = "netbsd",
     target_os = "openbsd"
 ))]
+/// Extensions necessary for creating a surface on the target platform.
 pub const TARGET_EXTENSIONS: [*const c_char; 4] = [
     khr::Surface::name().as_ptr(),
     khr::WaylandSurface::name().as_ptr(),
@@ -137,16 +139,19 @@ pub const TARGET_EXTENSIONS: [*const c_char; 4] = [
     khr::XcbSurface::name().as_ptr(),
 ];
 #[cfg(any(target_os = "android"))]
+/// Extensions necessary for creating a surface on the target platform.
 pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
     khr::Surface::name().as_ptr(),
     khr::AndroidSurface::name().as_ptr(),
 ];
 #[cfg(any(target_os = "macos"))]
+/// Extensions necessary for creating a surface on the target platform.
 pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
     khr::Surface::name().as_ptr(),
     ext::MetalSurface::name().as_ptr(),
 ];
 #[cfg(any(target_os = "ios"))]
+/// Extensions necessary for creating a surface on the target platform.
 pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
     khr::Surface::name().as_ptr(),
     ext::MetalSurface::name().as_ptr(),
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -118,6 +118,40 @@ pub unsafe fn create_surface(
     }
 }
 
+#[cfg(target_os = "windows")]
+pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
+    khr::Surface::name().as_ptr(),
+    khr::Win32Surface::name().as_ptr(),
+];
+#[cfg(any(
+    target_os = "linux",
+    target_os = "dragonfly",
+    target_os = "freebsd",
+    target_os = "netbsd",
+    target_os = "openbsd"
+))]
+pub const TARGET_EXTENSIONS: [*const c_char; 4] = [
+    khr::Surface::name().as_ptr(),
+    khr::WaylandSurface::name().as_ptr(),
+    khr::XlibSurface::name().as_ptr(),
+    khr::XcbSurface::name().as_ptr(),
+];
+#[cfg(any(target_os = "android"))]
+pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
+    khr::Surface::name().as_ptr(),
+    khr::AndroidSurface::name().as_ptr(),
+];
+#[cfg(any(target_os = "macos"))]
+pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
+    khr::Surface::name().as_ptr(),
+    ext::MetalSurface::name().as_ptr(),
+];
+#[cfg(any(target_os = "ios"))]
+pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
+    khr::Surface::name().as_ptr(),
+    ext::MetalSurface::name().as_ptr(),
+];
+
 /// Query the required instance extensions for creating a surface from a window handle.
 ///
 /// The returned extensions will include all extension dependencies.
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -2,7 +2,7 @@
 
 use std::os::raw::c_char;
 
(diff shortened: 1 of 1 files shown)


ash-window/src/lib.rs
```diff
@@ -118,6 +118,45 @@ pub unsafe fn create_surface(
...
+    khr::Surface::name().as_ptr(),
+    ext::MetalSurface::name().as_ptr(),
+];
+#[cfg(any(target_os = "ios"))]
+/// Extensions necessary for creating a surface on the target platform.
+pub const TARGET_EXTENSIONS: [*const c_char; 2] = [
+    khr::Surface::name().as_ptr(),
+    ext::MetalSurface::name().as_ptr(),
+];
+
```
Ralith, line 159: Let's make these slices rather than arrays so it's harder to inadvertently statically depend on the length.

iMplode-nZ: Alright.

MarijnS95 (changes requested):
Would you mind including the current platform name in the doc comment, so that users know it can vary across compilation targets? Would be bad if they start relying on the type containing ie. an array of size 2 and trivially fail compiling on Linux(es). At the same time I don't think we should be using `cfg`s to hide other options depending on the compile target, we're not doing that for platform-specific extensions either. As such, I'd rather see something like the following:

1. Make all of these arrays available, under the appropriate platform name. `TARGET_EXTENSIONS_WINDOWS`, `TARGET_EXTENSIONS_UNIX/LINUX`, `TARGET_EXTENSIONS_ANDROID`, etc. Cover it in a doc comment too;
2. Create a separate `TARGET_EXTENSIONS_CURRENT_PLATFORM` that sources from the right platform through `cfg`, documenting that it will obviously vary across targets on compile-time;
3. Reuse these slices in `fn enumerate_required_extensions()` because copying them is wasteful and confusing;
4. Nit: It seems common to write doc comments _above_ attributes (`#[cfg]`), not below.

iMplode-nZ:
@MarijnS95 Points 1, 2, and 4 seem nice, but actually enumerate_required_extensions doesn't give the same return value as TARGET_EXTENSIONS for linux (because of the 3 different windowing managers).

Ralith:
I'm not sure I agree with point 1. What's the use case for individual constants as part of the public API? If there is no specific compelling case, we shouldn't make the API wider than necessary; the current draft is solving a concrete use case as it stands. "TARGET" also already conveys "the compile target," so that naming is odd.

ash-window/src/lib.rs
```diff
@@ -118,6 +118,50 @@ pub unsafe fn create_surface(
     }
 }
 
+/// Extensions necessary for creating a surface on the target platform.
```
MarijnS95, line 121:
```suggestion
/// Extensions necessary for creating a surface on the current target platform.
```

Again, I feel like making it very clear that the value of this `const` is going to change depending on the compile target, and it doesn't feel like just mentioning "target platform" gives that enough weight.

iMplode-nZ:
@MarijnS95 Alright, I fixed all of your points. It should be good now.

Ralith:
> Its main use is reuse in fn enumerate_required_extensions(), instead of "depending" on the fact that both the match statement and the enablement of these consts share the same cfg.

This sounds good to me on its own, but they don't have to be part of the public API for that.

ash-window/src/lib.rs
```diff
@@ -170,13 +174,7 @@ pub fn enumerate_required_extensions(
...
     let extensions = match window_handle.raw_window_handle() {
         #[cfg(target_os = "windows")]
-        RawWindowHandle::Windows(_) => {
-            const WINDOWS_EXTS: [*const c_char; 2] = [
-                khr::Surface::name().as_ptr(),
-                khr::Win32Surface::name().as_ptr(),
-            ];
-            &WINDOWS_EXTS
-        }
+        RawWindowHandle::Windows(_) => &WINDOWS_SURFACE_EXTENSIONS,
```
Ralith, line 177: Is the `&` here and below needed?

iMplode-nZ: I don't think so, I'll fix that.

iMplode-nZ:
> This sounds good to me on its own, but they don't have to be part of the public API for that.

True, but there are some potential cases them be in the public api might be helpful, I don't really care though.

Ralith:
I think we should prefer not to complicate our public API except when it is specifically known to solve a real problem someone's having.

iMplode-nZ:
Alright.

ash-window/src/lib.rs
```diff
@@ -118,6 +118,59 @@ pub unsafe fn create_surface(
     }
 }
 
+/// Extensions necessary for creating a surface on windows.
+#[allow(dead_code)]
```
MarijnS95, line 122: If not making them public it seems cleaner to just repeat the `cfg()`?

ash-window/src/lib.rs
```diff
@@ -118,6 +118,59 @@ pub unsafe fn create_surface(
...
+    khr::Surface::name().as_ptr(),
+    ext::MetalSurface::name().as_ptr(),
+];
+/// Extensions necessary for creating a surface on ios.
+#[allow(dead_code)]
+const IOS_SURFACE_EXTENSIONS: &'static [*const c_char] = MACOS_SURFACE_EXTENSIONS;
+
+/// Extensions necessary for creating a surface on the current target platform.
+/// (Note that on Unix, this is not equal to the return value of [`enumerate_required_extensions`])
+pub const TARGET_EXTENSIONS: &'static [*const c_char] = {
```
MarijnS95, line 154: Since `TARGET` here refers to the current `target_os`, are we perhaps missing `_SURFACE_` in between? So:

```suggestion
pub const TARGET_SURFACE_EXTENSIONS: &'static [*const c_char] = {
```

ash-window/src/lib.rs
```diff
@@ -118,6 +118,59 @@ pub unsafe fn create_surface(
...
 }
 
+/// Extensions necessary for creating a surface on windows.
+#[allow(dead_code)]
+const WINDOWS_SURFACE_EXTENSIONS: &'static [*const c_char] = &[
+    khr::Surface::name().as_ptr(),
+    khr::Win32Surface::name().as_ptr(),
+];
+/// Extensions necessary for creating a surface on unix.
+/// Note that this is not equal to the return value of [`enumerate_required_extensions`] on Unix due to the multiple window types.
```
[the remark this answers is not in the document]
MarijnS95: How about elaborating this to return _all possible_ extensions on Unix, and explaining how `enumerate_required_extensions()` differs in that situation?

```suggestion
/// _All_ possible extensions for creating a surface on Unix.
///
/// Defer to [`enumerate_required_extensions()`] for a subset of this list applicable to the display server protocol specific to the given [`HasRawWindowHandle`].
```

Also add this to `TARGET_SURFACE_EXTENSIONS` since `UNIX_SURFACE_EXTENSIONS` is not public :disappointed: 

(If you intend these to be separate paragraphs, add an empty `///` in between. Otherwise the two sentences get concatenated into one paragraph.)

ash-window/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 
 use std::os::raw::c_char;
 
-use ash::{extensions::khr, prelude::*, vk, Entry, Instance};
+use ash::{extensions::ext, extensions::khr, prelude::*, vk, Entry, Instance};
```
MarijnS95, line 5: Then remove:

```rust
#[cfg(any(target_os = "macos", target_os = "ios"))]
use ash::extensions::ext; // portability extensions
```

You can't import the same thing in scope twice.

(But I rather see `cfg()` on the per-platform `const`s then)

MarijnS95 (changes requested):
Looks like the doc comments still need some work, and this probably fails to compile on mac/ios now.

ash-window/src/lib.rs
MarijnS95: Don't we have clippy calling [`needless_borrow`](https://rust-lang.github.io/rust-clippy/master/#needless_borrow) on this?

If not, that's bad.

MarijnS95:
@iMplode-nZ do you still intend to push this forward after a year? Otherwise I'll redo this with all of my requested changes applied.

iMplode-nZ:
Sorry, I forgot about it; I'd appreciate if you did it according to your designs though.

MarijnS95:
@iMplode-nZ no worries, it happens. I'll see what I can do, and ping @Ralith on some of my review comments to make sure my view/opinion makes sense before making any massive changes, but would like yo keep your PR and authorship as base.

PR #614, opened 2022-04-15
ash: Add `const STRUCTURE_TYPE` to all Vulkan structures for matching with `match_struct!` macro

MarijnS95:
Depends on #613

In Vulkan layers extracing a structure based on its `s_type` is a common operation, but comparing against an enum value and subsequently casting to the right type is verbose and error-prone.

By generating a `const STRUCTURE_TYPE` with the given value for every Vulkan structure it becomes possible to implement a macro that abstracts this logic away in a safer way.

### DIFF
--- a/ash/src/lib.rs
+++ b/ash/src/lib.rs
@@ -72,6 +72,105 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
     }
 }
 
+/// Given a mutable raw pointer to a type with an `s_type` member such as [`vk::BaseOutStructure`],
+/// match on a set of Vulkan structures. The struct will be rebound to the
+/// given variable of the type of the given Vulkan structure.
+///
+/// Note that all match bodies have to be enclosed by curly braces due to macro parsing limitations.
+/// It is unfortunately not possible to write `x @ ash::vk::SomeStruct => one_line_expression(),`.
+///
+/// ```
+/// let mut info = ash::vk::DeviceCreateInfo::default();
+/// let info: *mut ash::vk::BaseOutStructure = <*mut _>::cast(&mut info);
+/// unsafe {
+///     ash::match_out_struct!(match info {
+///         info @ ash::vk::DeviceQueueCreateInfo => {
+///             dbg!(&info); // Unreachable
+///         }
+///         info @ ash::vk::DeviceCreateInfo => {
+///             dbg!(&info);
+///         }
+///     })
+/// }
+/// ```
+///
+/// In addition this macro propagates implicit return values just like normal `match` blocks, as long
+/// as a default value or expression in the "any" match arm (`_ => { some_value() }).
+///
+/// ```
+/// # let mut info = ash::vk::DeviceCreateInfo::default();
+/// # let info: *mut ash::vk::BaseOutStructure = <*mut _>::cast(&mut info);
+/// let device_create_flags: Option<ash::vk::DeviceCreateFlags> = unsafe {
+///     ash::match_out_struct!(match info {
+///         info @ ash::vk::DeviceQueueCreateInfo => {
+///             dbg!(&info); // Unreachable
+///             Some(ash::vk::DeviceCreateFlags::empty())
+///         }
+///         info @ ash::vk::DeviceCreateInfo => {
+///             dbg!(&info);
+///             Some(info.flags)
+///         }
+///         _ => {
+///             None
+///         }
+///     })
+/// };
+/// ```
+#[macro_export]
+macro_rules! match_out_struct {
+    (match $p:ident { $($bind:ident @ $ty:path => $body:block $(,)?)+ $(_ => $any:block $(,)?)? }) => {
+        match std::ptr::addr_of!((*$p).s_type).read() {
+            $(<$ty as $crate::vk::TaggedStructure>::STRUCTURE_TYPE => {
+                let $bind = $p
+                    .cast::<$ty>()
+                    .as_mut()
+                    .unwrap();
+                $body
+            }),+
+            _ => { $($any)? }
+        }
+    };
+}
+
+/// Given an immutable raw pointer to a type with an `s_type` member such as [`vk::BaseInStructure`],
+/// match on a set of Vulkan structures. The struct will be rebound to the
+/// given variable of the type of the given Vulkan structure.
+///
+/// Note that all match bodies have to be enclosed by curly braces due to macro parsing limitations.
+/// It is unfortunately not possible to write `x @ ash::vk::SomeStruct => one_line_expression(),`.
+///
+/// ```
+/// let info = ash::vk::DeviceCreateInfo::default();
+/// let info: *const ash::vk::BaseInStructure = <*const _>::cast(&info);
+/// unsafe {
+///     ash::match_in_struct!(match info {
+///         info @ ash::vk::DeviceQueueCreateInfo => {
+///             dbg!(&info); // Unreachable
+///         }
+///         info @ ash::vk::DeviceCreateInfo => {
+///             dbg!(&info);
+///         }
+///     })
+/// }
+/// ```
+///
+/// See the [`match_out_struct!`] documentation for an example with implicit return values.
+#[macro_export]
+macro_rules! match_in_struct {
+    (match $p:ident { $($bind:ident @ $ty:path => $body:block $(,)?)+ $(_ => $any:block $(,)?)? }) => {
+        match std::ptr::addr_of!((*$p).s_type).read() {
+            $(<$ty as $crate::vk::TaggedStructure>::STRUCTURE_TYPE => {
+                let $bind = $p
+                    .cast::<$ty>()
+                    .as_ref()
+                    .unwrap();
+                $body
+            }),+
+            _ => { $($any)? }
+        }
+    };
+}
+
 #[cfg(test)]
 mod tests {
     use super::vk;
--- a/ash/src/lib.rs
+++ b/ash/src/lib.rs
@@ -73,8 +73,8 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
 }
 
 /// Given a mutable raw pointer to a type with an `s_type` member such as [`vk::BaseOutStructure`],
-/// match on a set of Vulkan structures. The struct will be rebound to the
-/// given variable of the type of the given Vulkan structure.
+/// match on a set of Vulkan structures. The struct will be rebound to the given variable of the
+/// type of the given Vulkan structure.
 ///
 /// Note that all match bodies have to be enclosed by curly braces due to macro parsing limitations.
 /// It is unfortunately not possible to write `x @ ash::vk::SomeStruct => one_line_expression(),`.
@@ -94,8 +94,10 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
 /// }
 /// ```
 ///
-/// In addition this macro propagates implicit return values just like normal `match` blocks, as long
-/// as a default value or expression in the "any" match arm (`_ => { some_value() }).
+/// In addition this macro propagates implicit return values just like normal `match` blocks, as
+/// long as a default value or expression is provided in the "any" match arm
+/// (`_ => { some_value() }`). For the time being said arm must be wrapped in curly braces; an
+/// expression like `_ => None` is not yet supported.
 ///
 /// ```
 /// # let mut info = ash::vk::DeviceCreateInfo::default();
@@ -133,8 +135,8 @@ macro_rules! match_out_struct {
 }
 
 /// Given an immutable raw pointer to a type with an `s_type` member such as [`vk::BaseInStructure`],
-/// match on a set of Vulkan structures. The struct will be rebound to the
-/// given variable of the type of the given Vulkan structure.
+/// match on a set of Vulkan structures. The struct will be rebound to the given variable of the
+/// type of the given Vulkan structure.
 ///
 /// Note that all match bodies have to be enclosed by curly braces due to macro parsing limitations.
 /// It is unfortunately not possible to write `x @ ash::vk::SomeStruct => one_line_expression(),`.
--- a/ash/src/vk/prelude.rs
+++ b/ash/src/vk/prelude.rs
@@ -54,9 +54,8 @@ impl From<vk::Extent2D> for vk::Rect2D {
 }
 
 /// Structures implementing this trait are layout-compatible with [`vk::BaseInStructure`] and
-/// [`vk::BaseOutStructure`]. That is, it has a `p_next` pointer field to chain other
-/// [`TaggedStructure`]s and an `s_type` field indicating its type, which must always match the
-/// value of [`TaggedStructure::STRUCTURE_TYPE`].
+/// [`vk::BaseOutStructure`]. Such structures have an `s_type` field indicating its type, which
+/// must always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
 pub unsafe trait TaggedStructure {
     const STRUCTURE_TYPE: vk::StructureType;
 }
--- a/ash/src/vk/prelude.rs
+++ b/ash/src/vk/prelude.rs
@@ -52,3 +52,11 @@ impl From<vk::Extent2D> for vk::Rect2D {
         }
     }
 }
+
+/// Structures implementing this trait are layout-compatible with [`vk::BaseInStructure`] and
+/// [`vk::BaseOutStructure`]. That is, it has a `p_next` pointer field to chain other
+/// [`TaggedStructure`]s and an `s_type` field indicating its type, which must always match the
+/// value of [`TaggedStructure::STRUCTURE_TYPE`].
+pub unsafe trait TaggedStructure {
+    const STRUCTURE_TYPE: vk::StructureType;
+}
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -497,7 +497,6 @@ pub trait FieldExt {
     /// array types (e.g. `[f32; 3]`) will be converted to pointer types (e.g. `&[f32; 3]`),
     /// which is needed for `C` function parameters. Set to `false` for struct definitions.
     fn type_tokens(&self, is_ffi_param: bool) -> TokenStream;
-    fn is_clone(&self) -> bool;
 
     /// Whether this is C's `void` type (not to be mistaken with a void _pointer_!)
     fn is_void(&self) -> bool;
@@ -632,10 +631,6 @@ fn discard_outmost_delimiter(stream: TokenStream) -> TokenStream {
 }
 
 impl FieldExt for vkxml::Field {
-    fn is_clone(&self) -> bool {
-        true
-    }
-
     fn param_ident(&self) -> Ident {
         let name = self.name.as_deref().unwrap_or("field");
         let name_corrected = match name {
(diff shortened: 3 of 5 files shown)


ash/src/lib.rs
```diff
@@ -66,6 +66,40 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+///         }
+///         info @ ash::vk::DeviceCreateInfo => {
+///             dbg!(&info);
+///         }
+///     })
+/// }
+/// ```
+#[macro_export]
+macro_rules! match_struct {
+    (match $p:ident { $($bind:ident @ $ty:path => $body:block $(,)?)+ }) => {
```
MarijnS95, line 89: `$body:block` should probably be changed to `tt` (or is there a better type) to allow bodies without `{ .. }` block.

ash/src/lib.rs
```diff
@@ -66,6 +66,40 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+///             dbg!(&info); // Unreachable
+///         }
+///         info @ ash::vk::DeviceCreateInfo => {
+///             dbg!(&info);
+///         }
+///     })
+/// }
+/// ```
+#[macro_export]
+macro_rules! match_struct {
```
MarijnS95, line 88: Naming suggestions welcome! Perhaps it should be moved into `ash/src/prelude.rs` too, but just like `mod vk` reexporting its own prelude we should surely do the same here?

(Not done currently because there's only `pub(crate)` in there)

ash/src/lib.rs
```diff
@@ -66,6 +66,40 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+    (match $p:ident { $($bind:ident @ $ty:path => $body:block $(,)?)+ }) => {
+        match (*$p).s_type {
+            $(<$ty as $crate::vk::HasStructureType>::STRUCTURE_TYPE => {
+                let $bind = $p
+                    .cast::<$ty>()
+                    .as_mut()
```
MarijnS95, line 94: Only mutable sturctures work with this, currently :/

ash/src/lib.rs
```diff
@@ -66,6 +66,40 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+                    .unwrap();
+                $body
+            }),+
+            _ => {}
```
MarijnS95, line 98: This catch-all is currently preventing the use of this `match_struct` to return values. Perhaps we can employ some sort of optional `$(_ => $...)?` argument to allow (or if necessary, force) the caller to set something here? I'd expect them to mostly use an empty block `{}` or `()` void, or `None` when otherwise using the macro to look for a specific type.

ash/src/vk/prelude.rs
```diff
@@ -56,3 +56,7 @@ impl From<vk::Extent2D> for vk::Rect2D {
         }
     }
 }
+
+pub trait HasStructureType {
+    const STRUCTURE_TYPE: vk::StructureType;
+}
```
MarijnS95, line 62: This is made into a trait with the assumption of adding some default `fn`s here that help with iterating `p_next`. I've not really filled in what to do here, and had some thoughts about also having certain getter functions in the form of `fn get_xxx<T: HasStructureType>(...) -> Option<T>` but nothing fleshed/figured out yet.

generator/src/lib.rs
```diff
@@ -2474,8 +2548,19 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
...
+            let def2 = definitions2
+                .iter()
+                .filter_map(get_variant!(vk_parse::TypesChild::Type))
+                .find(|def2| match def {
+                    vkxml::DefinitionsElement::Struct(s) => {
+                        def2.name.as_deref() == Some(s.name.as_str())
+                    }
+                    // TODO: Compare the name for more vkxml types
+                    _ => false,
+                });
```
MarijnS95, line 2560: This bit is pretty messy but we need the `vk_parse` type as `vkxml` doesn't propagate the `values` field, which tells for every `s_type` what value it *should* be when filled in on a particular type.

ash/src/lib.rs
```diff
@@ -66,6 +66,40 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+/// ```
+#[macro_export]
+macro_rules! match_struct {
+    (match $p:ident { $($bind:ident @ $ty:path => $body:block $(,)?)+ }) => {
+        match (*$p).s_type {
+            $(<$ty as $crate::vk::HasStructureType>::STRUCTURE_TYPE => {
+                let $bind = $p
+                    .cast::<$ty>()
+                    .as_mut()
+                    .unwrap();
```
MarijnS95, line 95: Perhaps a `NonNull` argument could prevent the unwrap here, and the `(*$p).s_type` becomes just `p.as_ref().s_type` above.

MarijnS95:
I'll test and post some build-timings later, see if +2300 trivial trait implementation lines have any effect. As I needed this while writing some Vulkan layers in Rust, the goal is to also submit a minimal layer crate for `ash` at some point, initially providing just some types before adding better utilities. We'll need to publicize `ptr_chain_iter` in `const` and `mut` variants too.

ash/src/lib.rs
Ralith: Maybe separate `match_out_struct` and `match_in_struct` macros?

ash/src/lib.rs
```diff
@@ -66,6 +66,40 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+///         info @ ash::vk::DeviceCreateInfo => {
+///             dbg!(&info);
+///         }
+///     })
+/// }
+/// ```
+#[macro_export]
+macro_rules! match_struct {
+    (match $p:ident { $($bind:ident @ $ty:path => $body:block $(,)?)+ }) => {
+        match (*$p).s_type {
```
Ralith, line 90: This derefs the entire `$p`, which might be unsound if any part of it is uninitialized, which could reasonably happen for out structs. Instead, we could use `ptr::addr_of` to compute the address of the field without dereffing, and then deref that field alone.

ash/src/lib.rs
Ralith: `Option` seems sensible; alternatively, the macro could require the caller to explicitly provide a default case. Or we could do nothing; after all, people can always assign via side effects if they want.

Ralith:
> I'll test and post some build-timings later, see if +2300 trivial trait implementation lines have any effect.

Easy enough to cfg-gate if needed, since it's a new trait.

ash/src/lib.rs
MarijnS95: I like those names!

Just unfortunate that it's going to be ~23 lines of duplication for the entire macro and doc-comment.

ash/src/lib.rs
MarijnS95: I've been able to get this "working" for both blocks and expressions with either `stmt` or `expr`, but the downside is that it doesn't enforce a trailing comma and successfully parses something like:

```rust
            info @ ash::vk::ApplicationInfo => panic!()
            info @ ash::vk::InstanceCreateInfo => panic!()
```

(note that `tt` doesn't understand macro calls like `panic!()` when not enclosed in anything that resembles a block or expression.)

On the other hand, with the addition of `$(_ => $any:xxx $(,)?)?` for an optional default value the compiler correctly demands the use of an explicit comma:

```rust
error: `$body:stmt` may be followed by `_`, which is not allowed for `stmt` fragments
   --> ash/src/lib.rs:115:72
    |
115 |     (match $p:ident { $($bind:ident @ $ty:path => $body:stmt $(,)?)+ $(_ => $any:stmt $(,)?)? }) => {
    |                                                                        ^ not allowed after `stmt` fragments
    |
    = note: allowed there are: `=>`, `,` or `;`
```

Ideally we should have something that either parses an optional comma for `$body:block $(,)?` _or_ an expression which is forcibly terminated by a comma: `$body:expr,`. While we could have multiple `macro_rules!` arms for either method, I'm not sure how to write that in such a way that it can be alternated for any of the clauses a user puts in the macro.

ash/src/lib.rs
MarijnS95: There's now an optional `_` wildcard arm for this case, where users can write their own body/expression just like the other match arm. Including a tested example in the docs.

ash/src/lib.rs
MarijnS95: @Ralith Will it temporarily copy all fields during dereferencing? I've previously seen a similar `&*` reborrowing technique being used and expect the `&` to be redundant here, but that probably makes no difference for the soundness issue here.

I however also remember previously having trouble with this method for converting from `*const` to `&` with `&*`, which was solved by simply transmuting. Leading to think that the compiler would perhaps run the destructor on the temporary created by `*` but that's not the case (and fwiw irrelevant on our Vulkan types since it's not allowed to implement both `Drop` and `Copy` simultaneously).

I should read up more on raw pointers :sweat_smile:

MarijnS95: I landed on `std::ptr::addr_of!((*$p).s_type).read_unaligned()` here (expression still needs to be valid), though all Vulkan structs should probably be aligned.

Ralith: > Will it temporarily copy all fields during dereferencing?

That's a fair mental model, though in formal terms I think it's more correct to say that it is the deref operation itself (which produces a place, not necessarily copying anything) that is UB.

> I however also remember previously having trouble with this method for converting from *const to & with &*, which was solved by simply transmuting

I'm not totally sure what you're alluding to here, but constructing a reference to uninitialized data is also UB.

ash/src/lib.rs
Ralith: Yeah; as far as compromises to deal with declarative macros being kind of crappy go, it could be a lot worse.

ash/src/lib.rs
Ralith: Nice, that's consistent with how `if` can have a non-unit value iff you have an `else` branch then.

ash/src/lib.rs
MarijnS95: > That's a fair mental model, though in formal terms I think it's more correct to say that it is the deref operation itself (which produces a place, not necessarily copying anything) that is UB.

Wasn't considering copying to be a mnemonic to this, just curious if it is the case at all. Fair enough, retrieving a borrow to possibly uninitialized memory - even if unused otherwise - is UB. 

> > I however also remember previously having trouble with this method for converting from _const to & with &_, which was solved by simply transmuting
> 
> I'm not totally sure what you're alluding to here, but constructing a reference to uninitialized data is also UB.

I'll probably have to repro this case again, but it was happening on initialized data.

Ralith: And yeah, `read_unaligned` is overkill, but that use of `addr_of` looks fine.

ash/src/lib.rs
MarijnS95: It's the same for any normal `match` block as well, considering all Vulkan enums should be `non_exhaustive` most definitely including `StructureType`. We don't annotate ours as such currently :sweat_smile:

ash/src/lib.rs
MarijnS95: This `NonNull::as_ref()` suggestion won't work considering potential UB discussed in another conversation thread.

MarijnS95:
`master @ 61ab543d`:

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.366 s ±  0.101 s    [User: 8.313 s, System: 0.670 s]
  Range (min … max):    7.248 s …  7.586 s    10 runs
```

`match-structure-type @ eb95efb`:

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.411 s ±  0.065 s    [User: 8.358 s, System: 0.638 s]
  Range (min … max):    7.305 s …  7.505 s    10 runs
```

We're talking approx 50ms here, nothing worth `cfg()`'ing nor worrying for.

MarijnS95:
@Ralith How do you feel about a `hyperfine` step for the core crate in CI?

Ralith:
I've heard CI tends to give erratic timings for CPU-bound stuff due to contention, but it might be interesting to try all the same.

ash/src/lib.rs
Ralith: That's because our enums aren't actually enums, they're all just newtyped integers with associated constants. `#[non_exhaustive]` enums wouldn't even work here, since it's still UB to transmute an unused bitpattern into those.

MarijnS95: Nvm, I got this mixed up with some other crate that tried to do this. These integers are nonexhaustive anyways (unless you actually match on every possible value/range, I guess?)

ash/src/vk/prelude.rs
```diff
@@ -56,3 +56,7 @@ impl From<vk::Extent2D> for vk::Rect2D {
         }
     }
 }
+
+pub trait HasStructureType {
```
Ralith, line 60: bikeshed: `TypedStructure`? `TaggedStructure`? `Tagged`?

MarijnS95: I think I only mentioned it on the macro but I was very much looking for naming suggestions here too. `TaggedStructure` seems the most fitting to me as it's already "typed" in the Rust sense. Applied all your suggestions, thanks :+1: 

Then, do you want to take this trait one step further?

- (default) implementation `fn in()` and `fn out()` to cast to the respective `Base*Structure`;
- Reuse that in `push_next()`, either by adding a trait bound on the function or require `TaggedStructure` for every `Extends*` trait;
- Helpers to get/find structures of a specific `s_type` in the chain, like #628?

Ralith: > (default) implementation fn in() and fn out() to cast to the respective Base*Structure;

I'm on board in theory, but needs more design. Should these return pointers or references? Mutable or shared, or double the methods for both?

> Helpers to get/find structures of a specific s_type in the chain

I don't feel like I understand the use case for that (or the macro this PR introduces, for that matter) well enough to judge usefulness. Are we trying to support people implementing drivers/layers?

MarijnS95: Good question, since we have `push_next()` hiding this detail it is perhaps less relevant than I originally thought, to waste compiletime over.

Same for the getter functions, those would be implemented using `match_*_struct` _anyway_ and the user has to deal with the possibility of a certain typed struct appearing in the chain more than once. Putting it on the user to implement this with a `for` loop at least requires them to think about the possibility.

Ralith: :+1:, easier to add stuff in the future if use cases arise.

ash/src/lib.rs
Ralith: This seems too niche to be part of the prelude, IMO. Most downstreams of ash are not implementing layers (which seems to be the sole motivation here?).

ash/src/lib.rs
```diff
@@ -72,6 +72,107 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+///         }
+///         _ => {
+///             None
+///         }
+///     })
+/// };
+/// ```
+#[macro_export]
+macro_rules! match_out_struct {
+    (match $p:ident { $($bind:ident @ $ty:path => $body:block $(,)?)+ $(_ => $any:block $(,)?)? }) => {
```
Ralith, line 123: Are the trailing commas accepted here appropriate, given that blocks are required?

MarijnS95: It's valid Rust syntax, I even came across a crate recently that enabled this in `rustfmt.toml`: https://rust-lang.github.io/rustfmt#match_block_trailing_comma

Ralith: Huh, okay then.

ash/src/lib.rs
MarijnS95: Let's move this to `util`, good call.

ash/src/vk/prelude.rs
```diff
@@ -56,3 +56,11 @@ impl From<vk::Extent2D> for vk::Rect2D {
         }
     }
 }
+
+/// Structures implementing this trait are layout-compatible with [`vk::BaseInStructure`] and
+/// [`vk::BaseOutStructure`]. That is, it has a `p_next` pointer field to chain other
+/// [`TaggedStructure`]s and an `s_type` field indicating its type, which must always match the
+/// value of [`TaggedStructure::STRUCTURE_TYPE`].
```
Ralith, line 63: This isn't a complete definition of layout-compatibility, and might confuse an inexperienced reader. If we want to provide a definition, we should also more clearly note the field ordering, and that the struct must be `repr(C)`.

Ralith (approved):

ash/src/lib.rs
MarijnS95: (It was part of the library root fwiw, but that can be considered a `prelude` or at least "too easily" accessible for how niche it is).

Yes indeed, I have currently only found this to be useful to write Vulkan layers in Rust with `ash`. When the user chains structs together themselves to e.g. read properties or features, they already have access to and should use the `let` bindings for the individual structs instead of prying out their structs from the chain again.

ash/src/vk/prelude.rs
MarijnS95: Only the trait bound (`Extends*` requiring this) may be interesting to consider up front, but no use yet so no need to add it I guess.

ash/src/lib.rs
Ralith: I thought it wasn't actually possible to export declarative macros from anywhere but the crate root?

ash/src/vk/prelude.rs
Ralith: Yeah, clever trait hierarchies that we don't actually need for anything sounds like a trap.

ash/src/vk/prelude.rs
MarijnS95: The main point here was stating that it must have the same layout as `Base*Structure`, then proceeding to explain that `s_type` is what's used here to be equal to `TaggedStructure`. Perhaps it is better to remove ``That is, it has a `p_next` pointer field to chain other [`TaggedStructure`]s`` and only mention `s_type` explicitly?

ash/src/lib.rs
MarijnS95: Yeah, I guess.

ash/src/vk/prelude.rs
MarijnS95: I don't consider it particularly clever or special, just something that exists. Users needing it outside of `ash` can simply expand their trait bounds if they want/need.

ash/src/lib.rs
MarijnS95: Right, you were replying to my question of moving to `prelude.rs`, I was convinced until writing the second comment that it actually already resided inside `prelude.rs`.

> I thought it wasn't actually possible to export declarative macros from anywhere but the crate root?

I don't actually know the rules for macros before nor after 2018. It seems that wherever they reside, the macros are only available through the crate root. As such we can move them wherever (we have an `ash/src/vk/macros.rs`, perhaps add `ash/src/macros.rs` for this non-API bit).

ash/src/vk/prelude.rs
Ralith: I understand the use of "that is, it has" here to specifically present a definition. Replacing it with "Such structures have" seems to more clearly describe properties of them without necessarily defining them in full, at which point I'm indifferent to including any discussion of p_next.

ash/src/lib.rs
Ralith: I think it's okay to support a reasonable subset of conventional syntax like this. Perhaps a more complete solution exists, but this is good enough and making declarative macros perfect can be very painful.

MarijnS95: Fortunately I had already updated the documentation to state that it _must_ be wrapped in curly braces and something like `_ => None` is currently not supported. Perhaps that triggers someone more fluent in declarative macros to PR a proper implementation :+1:

generator/src/lib.rs
MarijnS95: Resolved this by getting hold of the variant name for `s_type` just how our `Default` implementation already did. Much simpler.

MarijnS95:
@jdtatz only noticing https://github.com/jdtatz/ash/commits/match-structure-type-more-generic just now; any reason this wasn't suggested as a change to cut down on codegen for `push_next()`? What's the (hopefully positive) compile time improvement?

Or is this all blocking on the [recently submitted type support in `vk-parse`](https://github.com/krolli/vk-parse/pull/26) so that we can subsequently drop `vkxml` use in `ash`'s generator and [finally use the `validstructs` property](https://github.com/krolli/vk-parse/issues/25#issuecomment-1206235423), to keep `push_next()` for `pNext` versus `validstructs` fields apart so that we can - in the end - get https://github.com/ash-rs/ash/pull/622 in with a proper binding?

PR #619, opened 2022-05-10
Update Vulkan-Headers to 1.3.219

MarijnS95:
Note that doesn't generate cleanly as `vk.xml` since `1.3.213` contains an invalid `structextends="VkBaseOutStructure"`; I removed that locally and it's in the process of being fixed upstream.


MarijnS95:
Wasn't my intent to merge this, sorry for not making that clear. Just PRing it here for visibility (those who'd like to use this and have a supported driver can set up a git reference) and to get started on the API ports.

I'll see how the Khronos issue evolves since there was some thought behind doing this in `vk.xml`, but it's not at all future-proof. Depending on the pacing I may or may not work around this in the generator, would love to get this in and drop a `0.37.1` release.

Ralith:
Hmm, wonder how that got through review upstream; not like it's not breaking for C users either. Guess we'd just best hope they had a good reason.

MarijnS95:
We managed to fix the issue with a `validstructs` property: https://github.com/KhronosGroup/Vulkan-Headers/commit/2c823b7f27590ec0a489f7fbe14b154e13fa5cfb#diff-a852a94832433f254f7e5ff6cf889035be373606623d793dfe4a75e535600cbeR12096

While we don't support this properly yet with traits, this compiles and should pass CI now so I'd like to get this and, and I'll make sure to add support to `vk-parse` and `ash` before the next release.

Ralith (approved):

PR #622, opened 2022-05-10
extensions/ext: Add VK_EXT_pipeline_properties device extension

MarijnS95:
Includes generator support to emit `trait`s for `command` `param`s with `validstructs` attribute, and (`unsafely`) implements said `trait` for every struct listed in that attribute.


MarijnS95:
Finally putting https://github.com/krolli/vk-parse/pull/24 to use :)

ash/src/vk/extensions.rs
```diff
@@ -17672,6 +17672,20 @@ impl ExtPipelinePropertiesFn {
     pub const SPEC_VERSION: u32 = 1u32;
 }
 #[allow(non_camel_case_types)]
+#[doc = "Implemented for all types that can be passed as argument to `p_pipeline_properties` in [`PFN_vkGetPipelinePropertiesEXT`]"]
+pub unsafe trait PFN_vkGetPipelinePropertiesEXT_param_p_pipeline_properties_validstructs {
```
MarijnS95, line 17676: Better trait naming suggestions welcome!

ash/src/vk/extensions.rs
```diff
@@ -17672,6 +17672,20 @@ impl ExtPipelinePropertiesFn {
...
+    unsafe fn as_base_in_struct(&self) -> *const BaseInStructure {
+        <*const _>::cast(self)
+    }
+    unsafe fn as_base_out_struct(&mut self) -> *mut BaseOutStructure {
+        <*mut _>::cast(self)
+    }
```
MarijnS95, line 17682: Perhaps we should just cast within the trait, but this seems to be more clearly conveying the use-case. Note that the original `extern "system" fn PFN_vkGetPipelinePropertiesEXT` doesn't use the trait but retains `vk.xml`'s `*mut vk::BaseOutStructure` type.

ash/src/vk/extensions.rs
```diff
@@ -17672,6 +17672,20 @@ impl ExtPipelinePropertiesFn {
     pub const SPEC_VERSION: u32 = 1u32;
 }
 #[allow(non_camel_case_types)]
+#[doc = "Implemented for all types that can be passed as argument to `p_pipeline_properties` in [`PFN_vkGetPipelinePropertiesEXT`]"]
```
MarijnS95, line 17675: Perhaps the docs should link to the high-level wrapper, but I'm not sure if we should assume the existence of `crate::extensions::...::PipelineProperties::get_pipeline_properties`] within the generator.

ash/src/vk/extensions.rs
Ralith: Camel-casing the argument name alone, maybe? Some risk of collisions, but not the worst breakage if we do hit it.

ash/src/vk/extensions.rs
```diff
@@ -17672,6 +17672,20 @@ impl ExtPipelinePropertiesFn {
...
+pub unsafe trait PFN_vkGetPipelinePropertiesEXT_param_p_pipeline_properties_validstructs {
+    unsafe fn as_base_in_struct(&self) -> *const BaseInStructure {
+        <*const _>::cast(self)
+    }
+    unsafe fn as_base_out_struct(&mut self) -> *mut BaseOutStructure {
+        <*mut _>::cast(self)
+    }
+}
+unsafe impl PFN_vkGetPipelinePropertiesEXT_param_p_pipeline_properties_validstructs
+    for PipelinePropertiesIdentifierEXT<'_>
```
Ralith, line 17685: When do we expect more impls to arise? If there's only the one, this seems perhaps overkill.

ash/src/vk/extensions.rs
Ralith: I'd favor casting in the wrapper rather than introducing trait methods, just for the sake of simplicity.

ash/src/vk/extensions.rs
MarijnS95: Not sure, this is still the only param with `validstructs`, perhaps @oddhack is able to comment?

We could stick to `vk::BaseOutStructure` and a doc-comment in the `PipelineProperties` wrapper (effectively dropping the second commit from this PR) but it might fall out of date. Similar if we hand-roll this trait: rather generate as much as possible from the underlying `vk.xml` spec.

oddhack: I can't say this will never happen again but had I noticed it early enough I would have argued vigorously against this type punning (vs. passing a real VkBaseOutStructure or other no-op structure, and the actual structure in the pNext chain, say).

MarijnS95: Is this a pattern you don't want to support going forward? If so you might still introduce such a no-op structure and leverage existing `structextends` machinery for it, and somehow deprecate this attribute notion leaving it only for this `p_pipeline_properties` param (extended to also accept a no-op structure with the current `VkPipelinePropertiesIdentifierEXT` in `pNext`)?

ash/src/vk/extensions.rs
Ralith: I'd strip `p_` as unidiomatic in Rust as well. I think it's actually very descriptive in context, but the collision risk is unfortunate. Maybe `GetPipelinePropertiesPipelineProperties`? Reads odd, but makes sense with a moment's thought.

Ralith: How about `{}Param{}`? A delimiter makes sense to me, but "valid" feels redundant; implementing a trait is already intrinsically asserting a predicate.

MarijnS95: Good call; I wanted to convey that these are the `validstructs` for the parameter but that's already implicit when the parameter accepts implementers of this `trait`.

ash/src/vk/extensions.rs
oddhack: We'd have to create another extension to allow this, so I doubt there would be appetite for it in the WG. Avoiding the pattern in extensions introducing new APIs is probably doable.

MarijnS95: All the machinery is there to support it except that drivers are now forced to read the pNext chain?

Fine to keep using this construct to be honest, we just added support for it.

Ralith (approved):

PR #466, opened 2021-08-14
Fixed validation layer errors on macos

39ali:
The examples(triangle and textures) are causing validation errors on macOS (tested on Mojave 10.14.6) , i'll be listing them down below with the solutions :

**1.**

`ERROR:
VALIDATION [VUID-VkDeviceCreateInfo-pProperties-04451 (976972960)] : Validation Error: [ VUID-VkDeviceCreateInfo-pProperties-04451 ] Object 0: handle = 0x7fe53281c950, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x3a3b6ca0 | vkCreateDevice: VK_KHR_portability_subset must be enabled because physical device VkPhysicalDevice 0x7fe53281c950[] supports it The Vulkan spec states: If the [VK_KHR_portability_subset] extension is included in pProperties of vkEnumerateDeviceExtensionProperties, ppEnabledExtensions must include "VK_KHR_portability_subset". (https://vulkan.lunarg.com/doc/view/1.2.182.0/mac/1.2-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pProperties-04451)
`
solution : 
pushing `vk::KhrPortabilitySubsetFn` to `instance.enabled_extension_names()`  (examples/src/lib.rs ,line 293)

**2.**

.`ERROR:
VALIDATION [VUID-vkCreateDevice-ppEnabledExtensionNames-01387 (307460652)] : Validation Error: [ VUID-vkCreateDevice-ppEnabledExtensionNames-01387 ] Object 0: VK_NULL_HANDLE, type = VK_OBJECT_TYPE_INSTANCE; | MessageID = 0x12537a2c | Missing extension required by the device extension VK_KHR_portability_subset: VK_KHR_get_physical_device_properties2. The Vulkan spec states: All required extensions for each extension in the VkDeviceCreateInfo::ppEnabledExtensionNames list must also be present in that list (https://vulkan.lunarg.com/doc/view/1.2.182.0/mac/1.2-extensions/vkspec.html#VUID-vkCreateDevice-ppEnabledExtensionNames-01387)
`
solution : 
pushing `vk::KhrGetPhysicalDeviceProperties2Fn::name().as_ptr()` to `device.enabled_extension_names()` (examples/src/lib.rs ,line 224)

**3.**

`ERROR:
[...]

PR #623, opened 2022-05-18
examples: Fix validation errors on macOS and iOS

x3ro:
When trying to run the examples on macOS, I was seeing the following validation errors:

<details>
<summary>Error: vkCreateDevice: VK_KHR_portability_subset must be enabled [...]</summary>
ERROR: VALIDATION [VUID-VkDeviceCreateInfo-pProperties-04451 (976972960)] : Validation Error: [ VUID-VkDeviceCreateInfo-pProperties-04451 ] Object 0: handle = 0x6000001f8020, type = VK_OBJECT_TYPE_PHYSICAL_DEVICE; | MessageID = 0x3a3b6ca0 | vkCreateDevice: VK_KHR_portability_subset must be enabled because physical device VkPhysicalDevice 0x6000001f8020[] supports it The Vulkan spec states: If the VK_KHR_portability_subset extension is included in pProperties of vkEnumerateDeviceExtensionProperties, ppEnabledExtensionNames must include "VK_KHR_portability_subset" (https://vulkan.lunarg.com/doc/view/1.3.211.0/mac/1.3-extensions/vkspec.html#VUID-VkDeviceCreateInfo-pProperties-04451)
</details>

<details>
<summary>Error: Attempting to create a VkDevice from a VkPhysicalDevice which is from a portability driver without the VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR [...]</summary>
ERROR: GENERAL [Loader Message (0)] : vkCreateDevice: Attempting to create a VkDevice from a VkPhysicalDevice which is from a portability driver without the VK_INSTANCE_CREATE_ENUMERATE_PORTABILITY_BIT_KHR bit in the VkInstanceCreateInfo flags being set and the VK_KHR_portability_enumeration extension enabled. In future versions of the loader this VkPhysicalDevice will not be enumerated.
</details>

I was able to fix these errors with the attached changes, and later realized they are somewhat similar to #466. As far as I can tell (from the error messages) enabling these extensions is required by the spec for running on macOS, but I'm quite new to Vulkan, so I might be wrong.

In any case, feedback is appreciated :)

### DIFF
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -15,6 +15,11 @@ use std::ffi::CStr;
 use std::ops::Drop;
 use std::os::raw::c_char;
 
+#[cfg(any(target_os = "macos", target_os = "ios"))]
+use ash::vk::{
+    KhrGetPhysicalDeviceProperties2Fn, KhrPortabilityEnumerationFn, KhrPortabilitySubsetFn,
+};
+
 use winit::{
     event::{ElementState, Event, KeyboardInput, VirtualKeyCode, WindowEvent},
     event_loop::{ControlFlow, EventLoop},
@@ -229,6 +234,13 @@ impl ExampleBase {
                 .to_vec();
             extension_names.push(DebugUtils::name().as_ptr());
 
+            #[cfg(any(target_os = "macos", target_os = "ios"))]
+            {
+                extension_names.push(KhrPortabilityEnumerationFn::name().as_ptr());
+                // Enabling this extension is a requirement when using `VK_KHR_portability_subset`
+                extension_names.push(KhrGetPhysicalDeviceProperties2Fn::name().as_ptr());
+            }
+
             let appinfo = vk::ApplicationInfo::default()
                 .application_name(app_name)
                 .application_version(0)
@@ -236,10 +248,17 @@ impl ExampleBase {
                 .engine_version(0)
                 .api_version(vk::make_api_version(0, 1, 0, 0));
 
+            let create_flags = if cfg!(any(target_os = "macos", target_os = "ios")) {
+                vk::InstanceCreateFlags::ENUMERATE_PORTABILITY_KHR
+            } else {
+                vk::InstanceCreateFlags::default()
+            };
+
             let create_info = vk::InstanceCreateInfo::default()
                 .application_info(&appinfo)
                 .enabled_layer_names(&layers_names_raw)
-                .enabled_extension_names(&extension_names);
+                .enabled_extension_names(&extension_names)
+                .flags(create_flags);
 
             let instance: Instance = entry
                 .create_instance(&create_info, None)
@@ -293,7 +312,11 @@ impl ExampleBase {
                 })
                 .expect("Couldn't find suitable device.");
             let queue_family_index = queue_family_index as u32;
-            let device_extension_names_raw = [Swapchain::name().as_ptr()];
+            let device_extension_names_raw = [
+                Swapchain::name().as_ptr(),
+                #[cfg(any(target_os = "macos", target_os = "ios"))]
+                KhrPortabilitySubsetFn::name().as_ptr(),
+            ];
             let features = vk::PhysicalDeviceFeatures {
                 shader_clip_distance: 1,
                 ..Default::default()


examples/src/lib.rs
```diff
@@ -229,17 +230,30 @@ impl ExampleBase {
                 .to_vec();
             extension_names.push(DebugUtils::name().as_ptr());
 
+            #[cfg(any(target_os = "macos", target_os = "ios"))]
+            {
+                extension_names.push(KhrGetPhysicalDeviceProperties2Fn::name().as_ptr());
```
MarijnS95, line 235: This is a device extension and is required for `VK_KHR_portability_subset` below, not the `VK_KHR_portability_enumeration` **instance** extension here. Can you move it down and also add a comment that, while we're not using this extension directly, states it is a requirement for `KhrPortabilitySubsetFn`?

examples/src/lib.rs
```diff
@@ -229,17 +230,30 @@ impl ExampleBase {
...
                 .application_version(0)
                 .engine_name(app_name)
                 .engine_version(0)
                 .api_version(vk::make_api_version(0, 1, 0, 0));
 
+            #[cfg(any(target_os = "macos", target_os = "ios"))]
+            let create_flags = vk::InstanceCreateFlags::ENUMERATE_PORTABILITY_KHR;
+
+            #[cfg(not(any(target_os = "macos", target_os = "ios")))]
+            let create_flags = vk::InstanceCreateFlags::default();
```
MarijnS95, line 250: I usually prefer _not_ having the `#[cfg()]` clause duplicated (can be done with `cfg_if`), and compile-test both "blocks" if they're not compiletime-dependent on the given cfg:

```suggestion
            let create_flags = if cfg(any(target_os = "macos", target_os = "ios")) {
                vk::InstanceCreateFlags::ENUMERATE_PORTABILITY_KHR
            else
                vk::InstanceCreateFlags::default()
            };
```

MarijnS95 (commented):
i'm not familiar with macOS but yes, I think you have to turn this on if an implementation is not compliant, "acknowledging" that you know where it lacks and that the application can deal with it?

Perhaps @MaikKlein - who uses Mac - can provide more insight.

MarijnS95:
Note that #466 also (in a bad way) turns on features inside https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkPhysicalDevicePortabilitySubsetFeaturesKHR.html. Since you don't get any warnings for this, I presume the examples don't use any of the unsupported features (and as such are safe to run with this).

examples/src/lib.rs
x3ro: Mmh, maybe I'm misunderstanding something here regarding the difference between instance and device extension, could you clarify for me? If I try to use it as a device extension, then the application crashes with

> ERROR:
> GENERAL [Loader Message (0)] : loader_validate_device_extensions: Device extension VK_KHR_get_physical_device_properties2 not supported by selected physical device or enabled layers.

[This doc](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_KHR_get_physical_device_properties2.html) also lists as an instance extension.

MarijnS95: It indeed is a **physical** device extension which implies it is an instance extension and not a **device** extension, I don't know what I was doing while writing this comment...

examples/src/lib.rs
x3ro: Cool, I didn't know about that, thanks!

x3ro:
I think I addressed your concerns, except for the part that I don't fully understand. Let me know what you think :)

MarijnS95 (approved):

MarijnS95:
I think we're okay here, but do double-check that you don't need anything extra from https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkPhysicalDevicePortabilitySubsetFeaturesKHR.html.

x3ro:
When you say "not need anything extra", you mean for the examples, right? I can definitely say that I can run both examples without errors now, but wouldn't know how to make sure that I don't need anything else beyond that 😅

MarijnS95:
I think we're fine if there are no errors. Nothing in that struct seems applicable to this rather simple rasterization example.

MarijnS95 (approved):

Ralith:
Let's cfg-gate the new `use`s too to keep CI happy.

MarijnS95:
Fortunately we have CI catching this for us :)

x3ro:
Yay CI 🎉 This should do it, I suppose 😅

PR #625, opened 2022-05-21
Remove "drop" mention from `create_*` docs when the result doesn't implement `Drop`

### LINKED ISSUE #624: ash::Instance::create_device needs clarification on dropping
The create_device docs says:
```
    /// # Safety
    /// In order for the created [`Device`] to be valid for the duration of its
    /// usage, the [`Instance`] this was called on must be dropped later than the
    /// resulting [`Device`].
```

However, AFAIK `Instance` doesn't implement `Drop` and what this means is probably `destroy_instance` must be called after `Device::destroy_device` is called.

Is this correct? If so, the doc comment should be reworded to not use the Rust keyword "drop".

MarijnS95:
Closes #624

Ash doesn't implement `Drop` intentionally, to not be too opinionated about holding (heap) references to their parent objects (`Device`->`Instance`->`Entry`) and ensuring they are destroyed in the right order.  As such, reword the `create` documentation for `Instance` and `Device` to mention their respective `destroy_*` function instead of referring to them as being "droppable".

Note that `Entry` is droppable as it does not have a Vulkan `destroy` function _and_ the dynamically loaded library (behind the "loaded" feature) is kept alive only for the lifetime of `Entry`.

---

The documentation is quite hard to read now, I'm thinking about rewording it to simply mention that a device **must** (in terms of ordering, and in terms of something you _should_ do to prevent object leaking) be `destroy`ed _before_ the parent `Entry`/`Instance` is destroyed.

### DIFF
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -16,15 +16,22 @@ use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
 ///
 /// # Safety
 ///
-/// In order for the created [`vk::SurfaceKHR`] to be valid for the duration of its
-/// usage, the [`Instance`] this was called on must be dropped later than the
-/// resulting [`vk::SurfaceKHR`].
+/// There is a [parent/child relation] between [`Instance`] and [`Entry`], and the resulting
+/// [`vk::SurfaceKHR`].  The application must not [destroy][Instance::destroy_instance()] these
+/// parent objects before first [destroying][khr::Surface::destroy_surface()] the returned
+/// [`vk::SurfaceKHR`] child object.  [`vk::SurfaceKHR`] does _not_ implement [drop][drop()]
+/// semantics and can only be destroyed via [`destroy_surface()`][khr::Surface::destroy_surface()].
+///
+/// See the [`Entry::create_instance()`] documentation for more destruction ordering rules on
+/// [`Instance`].
 ///
 /// The window represented by `window_handle` must be associated with the display connection
 /// in `display_handle`.
 ///
 /// `window_handle` and `display_handle` must be associated with a valid window and display
 /// connection, which must not be destroyed for the lifetime of the returned [`vk::SurfaceKHR`].
+///
+/// [parent/child relation]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#fundamentals-objectmodel-lifetime
 pub unsafe fn create_surface(
     entry: &Entry,
     instance: &Instance,
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -34,8 +34,14 @@ impl Entry {
     /// development packages installed (e.g. the Vulkan SDK, or Ubuntu's `libvulkan-dev`).
     ///
     /// # Safety
+    ///
     /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
     /// for [`Library::new()`] and [`Library::get()`] apply here.
+    ///
+    /// No Vulkan functions loaded directly or indirectly from this [`Entry`]
+    /// may be called after it is [dropped][drop()].
+    ///
+    /// # Example
     ///
     /// ```no_run
     /// use ash::{vk, Entry};
@@ -83,6 +89,11 @@ impl Entry {
     /// Note that instance/device functions are still fetched via `vkGetInstanceProcAddr` and
     /// `vkGetDeviceProcAddr` for maximum performance.
     ///
+    /// Any Vulkan function acquired directly or indirectly from this [`Entry`] may be called after it
+    /// is [dropped][drop()].
+    ///
+    /// # Example
+    ///
     /// ```no_run
     /// use ash::{vk, Entry};
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -113,8 +124,12 @@ impl Entry {
     /// Load Vulkan library at `path`
     ///
     /// # Safety
+    ///
     /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
     /// for [`Library::new()`] and [`Library::get()`] apply here.
+    ///
+    /// No Vulkan functions loaded directly or indirectly from this [`Entry`]
+    /// may be called after it is [dropped][drop()].
     #[cfg(feature = "loaded")]
     #[cfg_attr(docsrs, doc(cfg(feature = "loaded")))]
     pub unsafe fn load_from(path: impl AsRef<OsStr>) -> Result<Self, LoadingError> {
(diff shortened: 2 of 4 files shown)


ishitatsuyuki (approved):

Ralith (approved):
Clear improvement, but I like your idea for further improvement too.

MarijnS95:
Otoh the proposed sentence is hard to bolt on to `to be valid for the duration of its usage`, I'll think of something better when my mind is clear :)

MarijnS95:
I have given this one more pass, let me know what you think including nits! It's pretty non-trivial to explain this correctly.

For one I was thinking about moving ``(_not_ [dropped][`drop()`])`` into a separate (loose standing paragraph) sentence stating something along the lines of ``Note that this **must** be destroyed using [`destroy_xxx()`] as this type does not implement [`drop()`] semantics``; with our without justification. Perhaps reword it to clearly state that (implicitly) dropping it is simply not enough.

ash-window/src/lib.rs
```diff
@@ -14,9 +14,15 @@ use ash::extensions::ext; // portability extensions
...
 ///
-/// In order for the created [`vk::SurfaceKHR`] to be valid for the duration of its
-/// usage, the [`Instance`] this was called on must be dropped later than the
-/// resulting [`vk::SurfaceKHR`].
+/// The resulting [`vk::SurfaceKHR`] **must** be destroyed through
+/// [`khr::Surface::destroy_surface()`] (_not_ [dropped][drop()]) before the [`Instance`] and
+/// [`Entry`] this was called on is [destroyed][Instance::destroy_instance()], in the broad sense
+/// of the word. That is, valid Vulkan usage requires the [`vk::SurfaceKHR`] to be destroyed, and
+/// it otherwise becomes invalid to use after the [`Instance`] was
+/// [destroyed][Instance::destroy_instance()] and/or the [`Entry`] was [dropped][drop()].
```
Ralith, line 22: I'm not sure what "in the broad sense of the word" is referring to here. Editing detritus?

> valid Vulkan usage requires the [`vk::SurfaceKHR`] to be destroyed

Not strictly true, I think? You can kill the process without destroying anything and that's fine from a correctness standpoint.

ash/src/entry.rs
```diff
@@ -37,9 +37,15 @@ impl Entry {
     /// development packages installed (e.g. the Vulkan SDK, or Ubuntu's `libvulkan-dev`).
     ///
     /// # Safety
+    ///
     /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
     /// for [`Library::new()`] and [`Library::get()`] apply here.
     ///
+    /// For one, no Vulkan functions (indirectly) loaded from this [`Entry`] can be called after it
+    /// is [dropped][drop()].
```
Ralith, line 45:
```suggestion
    /// No Vulkan functions loaded directly or indirectly from this [`Entry`] may be called after it
    /// is [dropped][drop()].
```

ash/src/entry.rs
```diff
@@ -116,8 +124,12 @@ impl Entry {
     /// Load Vulkan library at `path`
     ///
     /// # Safety
+    ///
     /// `dlopen`ing native libraries is inherently unsafe. The safety guidelines
     /// for [`Library::new()`] and [`Library::get()`] apply here.
+    ///
+    /// For one, no Vulkan functions (indirectly) loaded from this [`Entry`] can be called after it
+    /// is [dropped][drop()].
```
Ralith, line 132: As above.

ash/src/entry.rs
```diff
@@ -214,9 +230,11 @@ impl Entry {
...
     ///
     /// # Safety
-    /// In order for the created [`Instance`] to be valid for the duration of its
-    /// usage, the [`Entry`](Self) this was called on must be dropped later than the
-    /// resulting [`Instance`].
+    ///
+    /// The resulting [`Instance`] **must** be destroyed through [`Instance::destroy_instance()`]
+    /// (_not_ [dropped][drop()]) before the [`Entry`] this was called on is [dropped][drop()], in
+    /// the broad sense of the word. That is, valid Vulkan usage requires the [`Instance`] to be
+    /// destroyed, and it otherwise becomes invalid to use after this [`Entry`] is dropped.
```
Ralith, line 237: - What is the broad sense of "dropped"?
- Instances do not have to be destroyed; rather, it is illegal (in general) to destroy a parent object without first destroying its children.
- It's not specifically invalid to use a non-destroyed instance after the `Entry` was dropped; rather, you have already invoked undefined behavior at that point and all bets are off.

Ralith (changes requested):
I find the new wording a bit verbose and confusing TBH. Maybe we can simplify things by citing and/or cribbing from the spec? The [Object Lifetime](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#fundamentals-objectmodel-lifetime) section has relevant wording. In particular:

> when there is a parent/child relationship between objects [...] the application must not destroy a parent object before its children

ash/src/entry.rs
Ralith: Further, dropping a `linked` `Entry` is harmless.

ash-window/src/lib.rs
MarijnS95: > I'm not sure what "in the broad sense of the word" is referring to here. Editing detritus?

To `**must**`, but that link fell away after more words got inserted. Seems to be called editing detritus :)

MarijnS95: > Not strictly true, I think? You can kill the process without destroying anything and that's fine from a correctness standpoint.

That's one possible edge-case to consider. Perhaps we should reword it using the parent/child relation quoted above, as long as `when there is a parent/child relationship` is converted into **this is a parent/child relationship**, clearly stating the inter-dependency on all the types involved in this PR.

Ralith: SGTM.

ash/src/entry.rs
MarijnS95: > * What is the broad sense of "dropped"?

As explained above, referring to `**must**` instead of `dropped`. Will find a different way of writing that up.

> * Instances do not have to be destroyed; rather, it is illegal (in general) to destroy a parent object without first destroying its children.

I'll see how this can be worded in here: clearly stating that there is a parent/child relation here, and who those parent(s) are.

> * It's not specifically invalid to use a non-destroyed instance after the `Entry` was dropped; rather, you have already invoked undefined behavior at that point and all bets are off.

It is useful to mention this though. Perhaps drop the "valid Vulkan usage requires ... to be destroyed" bit? Especially considering the option below:

> Further, dropping a `linked` `Entry` is harmless.

I had considered this case but not thought how to best describe it here. Dropping it in that instance simply removes access to the function pointers inside it, but doesn't violate other pointers in any way.

Ralith: > I had considered this case but not thought how to best describe it here.

Perhaps something like "The instance, and any objects derived from it, may only be manipulated for the period described in constructor used to obtain this `Entry`"?

ash-window/src/lib.rs
MarijnS95: Almost a year (just shy of a week) in the making, I've now reworded this to say that _if_ you were about to destroy one of the parent objects (detailing which, of course), the child object should be destroyed first.

ash/src/entry.rs
MarijnS95: Reworded this to state that the returned `Instance` object (and anything derived from it) should not be used anymore after `Entry` was dropped, unless created via `Entry::linked()` or `Entry::from_parts_1_1()`. Not entirely sure about the last one, since it has more complex library/lifetime requirements provided by the user...

Ralith (approved):

ash/src/entry.rs
Ralith: > Not entirely sure about the last one, since it has more complex library/lifetime requirements provided by the user...

Seems fine. No matter what else might be going on, the `Entry`'s lifetime isn't specifically relevant.

MarijnS95: It'd be too complex to write down, if we do it should be described on `from_parts_1_1()` instead which currently carries no documentation whatsoever. There's not `Arc<libloading::Library>` involved for the function-pointer lifetime at least.

PR #626, opened 2022-05-26
Add Fuchsia support

abarth:
Use the proper name for `libvulkan.so` on Fuchsia.

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -63,11 +63,11 @@ impl Entry {
 
         #[cfg(all(
             unix,
-            not(any(target_os = "macos", target_os = "ios", target_os = "android"))
+            not(any(target_os = "macos", target_os = "ios", target_os = "android", target = "fuchsia"))
         ))]
         const LIB_PATH: &str = "libvulkan.so.1";
 
-        #[cfg(target_os = "android")]
+        #[cfg(any(target_os = "android", target_os = "fuchsia"))]
         const LIB_PATH: &str = "libvulkan.so";
 
         #[cfg(any(target_os = "macos", target_os = "ios"))]
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -63,7 +63,7 @@ impl Entry {
 
         #[cfg(all(
             unix,
-            not(any(target_os = "macos", target_os = "ios", target_os = "android", target = "fuchsia"))
+            not(any(target_os = "macos", target_os = "ios", target_os = "android", target_os = "fuchsia"))
         ))]
         const LIB_PATH: &str = "libvulkan.so.1";
 


ash/src/entry.rs
```diff
@@ -63,11 +63,11 @@ impl Entry {
 
         #[cfg(all(
             unix,
-            not(any(target_os = "macos", target_os = "ios", target_os = "android"))
+            not(any(target_os = "macos", target_os = "ios", target_os = "android", target = "fuchsia"))
```
MarijnS95, line 66:
```suggestion
            not(any(target_os = "macos", target_os = "ios", target_os = "android", target_os = "fuchsia"))
```

?

MarijnS95 (changes requested):

MarijnS95:
https://doc.rust-lang.org/reference/conditional-compilation.html#target_os I'm surprised the official docs don't mention the existence of `fuchsia` :/

Ralith (approved):
Fuchsia doesn't do the unixy `.so.$MAJOR` thing? Interesting; is this documented anywhere?

Ralith:
Note that if you build ash with `features = ["linked"], default-features = false`, existing versions of ash should Just Work, and be less fragile at runtime besides.

abarth:
> Fuchsia doesn't do the unixy .so.$MAJOR thing?

Yes, we do that for most shared libraries.  I'm not sure why `libvulkan.so` doesn't have a `.$MAJOR` suffix.

> Interesting; is this documented anywhere?

https://fuchsia.dev/fuchsia-src/development/graphics/magma/concepts/vulkan?hl=en#buildtime_dependencies

> Any way we can sensibly cover this in CI?

Not yet.  We're working on that as part of our quest for Tier 1 support in Rust.

> Note that if you build ash with features = ["linked"], default-features = false, existing versions of ash should Just Work, and be less fragile at runtime besides.

Ok, I'll give that a try.

Ralith:
> Yes, we do that for most shared libraries. I'm not sure why libvulkan.so doesn't have a .$MAJOR suffix.

Hmm, are you using the standard Khronos Vulkan loader? This may be a bug in the upstream build system, if so. I'm not entirely clear on the significance of respecting that pattern vs. not, but I figure it's probably worth being consistent with the system at large.

MarijnS95:
@abarth any update on not using/having a `.MAJOR` suffix on Fuchsia?

MarijnS95 (approved):

MarijnS95:
Looking at the linked Fuchsia docs it's still `libvulkan.so` without `.MAJOR` suffix; I'll merge this in for the `0.38` release.

MarijnS95:
For `cargo:rustc-link-lib=vulkan` under the `linked` feature, this should keep working OOTB :)

PR #627, opened 2022-05-27
Document crate `feature`s in library docs

### LINKED ISSUE #583: Add documentation for features.
In the documentation, I can see 5 features:

1. default
2. loaded
3. debug
4. libloading
5. linked

Features 1, 2, 3, 4 are enabled by default.

Could you please describe what all these features do?

lilly-lizard:
Closes #583

Just copied and pasted the Cargo.toml feature comments into lib.rs based on the first recommendation on the [cargo feature docs](https://doc.rust-lang.org/cargo/reference/features.html#feature-documentation-and-discovery) (which links [regex](https://github.com/rust-lang/regex/blob/1.4.2/src/lib.rs#L488-L583) as reference). Also added a note in Cargo.toml so that anyone who updates features in the future knows to also update the docs.

**Reasoning:**
Makes feature descriptions more accessible (see discussion on issue #583). Currently you can see the list of ash features on [docs.rs](https://docs.rs/crate/ash/latest/features) but need to get Cargo.toml to see what they do. All three features seem to have potential utility for users imo.

### DIFF
--- a/ash/Cargo.toml
+++ b/ash/Cargo.toml
@@ -14,14 +14,13 @@ rust-version = "1.59.0"
 [dependencies]
 libloading = { version = "0.7", optional = true }
 
-# note: features are also documented at the top of lib.rs
 [features]
 default = ["loaded", "debug"]
 # Link the Vulkan loader at compile time.
 linked = []
 # Support searching for the Vulkan loader manually at runtime.
 loaded = ["libloading"]
-# Whether Vulkan structs should implement Debug
+# Whether Vulkan structs should implement Debug.
 debug = []
 
 [package.metadata.release]
--- a/ash/Cargo.toml
+++ b/ash/Cargo.toml
@@ -14,6 +14,7 @@ rust-version = "1.59.0"
 [dependencies]
 libloading = { version = "0.7", optional = true }
 
+# note: features are also documented at the top of lib.rs
 [features]
 default = ["loaded", "debug"]
 # Link the Vulkan loader at compile time.
--- a/ash/src/lib.rs
+++ b/ash/src/lib.rs
@@ -36,7 +36,7 @@
 //!
 //! ## Crate features
 //!
-//! * **debug** (default): Whether Vulkan structs should implement Debug.
+//! * **debug** (default): Whether Vulkan structs should implement `Debug`.
 //! * **loaded** (default): Support searching for the Vulkan loader manually at runtime.
 //! * **linked**: Link the Vulkan loader at compile time.
 
--- a/ash/src/lib.rs
+++ b/ash/src/lib.rs
@@ -34,11 +34,11 @@
 //! using [`Entry::load()`], which uses `libloading`. If you want to perform entry point loading
 //! yourself, call [`Entry::from_static_fn()`].
 //!
-//! ## Create features
+//! ## Crate features
 //!
-//! * __debug__ (default): Whether Vulkan structs should implement Debug.
-//! * __loaded__ (default): Support searching for the Vulkan loader manually at runtime.
-//! * __linked__: Link the Vulkan loader at compile time.
+//! * **debug** (default): Whether Vulkan structs should implement Debug.
+//! * **loaded** (default): Support searching for the Vulkan loader manually at runtime.
+//! * **linked**: Link the Vulkan loader at compile time.
 
 pub use crate::device::Device;
 pub use crate::entry::Entry;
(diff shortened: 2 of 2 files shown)


ash/Cargo.toml
```diff
@@ -14,6 +14,7 @@ rust-version = "1.59.0"
 [dependencies]
 libloading = { version = "0.7", optional = true }
 
+# note: features are also documented at the top of lib.rs
```
MarijnS95, line 17: I'm not too fussed about this one, perhaps we might just remove the docs here entirely as they're not easily visible at all, and I've not seen many crates do this.

ash/src/lib.rs
```diff
@@ -33,6 +33,12 @@
 //! Load the Vulkan library linked at compile time using [`Entry::linked()`], or load it at runtime
 //! using [`Entry::load()`], which uses `libloading`. If you want to perform entry point loading
 //! yourself, call [`Entry::from_static_fn()`].
+//!
+//! ## Create features
+//!
+//! * **linked** - Link the Vulkan loader at compile time.
```
MarijnS95, line 39: Shall we list the `(default)` features first? Perhaps put them under separate headings, like how clap does it:

https://docs.rs/clap/latest/clap/#feature-flags

Also, I think I prefer a colon (`:`) better than a hyphen (`-`) here.

ash/src/lib.rs
```diff
@@ -33,6 +33,12 @@
...
+//! * **loaded** (default) - Support searching for the Vulkan loader manually at runtime.
+//! * **debug** (default) - Whether Vulkan structs should implement Debug
```
MarijnS95, line 41: This sentence should also end with a period (`.`)... But that was mistaken in the original `Cargo.toml` comment too, feel free to add it there.

MarijnS95 (commented):
I personally prefer to write these in backticks, but it seems all crates use bold (`**`) so that's probably fine. Few small nits, for example we could possibly use separate headings like `clap` and use it to list the default features at the top.

MarijnS95:
> Currently you can see the list of ash features on [docs.rs](https://docs.rs/crate/ash/latest/features) but need to get Cargo.toml to see what they do.

I hadn't really looked at all the many docs.rs pages yet, didn't know this view exists. Any idea if there's a way to get useful documentation to show up on that page?

ash/Cargo.toml
lilly-lizard: Sounds good. No harm in leaving the original comments there though, some people might look in Cargo.toml for feature info.

ash/src/lib.rs
lilly-lizard: Sounds good. Wouldn't want a separate heading though because unlike clap there's only 3 features.

lilly-lizard:
> I hadn't really looked at all the many docs.rs pages yet, didn't know this view exists. Any idea if there's a way to get useful documentation to show up on that page?

Unfortunately not. There's currently no standardized feature documentation method in the rust ecosystem (discussion here https://github.com/rust-lang/api-guidelines/discussions/43)

ash/src/lib.rs
```diff
@@ -33,6 +33,12 @@
 //! Load the Vulkan library linked at compile time using [`Entry::linked()`], or load it at runtime
 //! using [`Entry::load()`], which uses `libloading`. If you want to perform entry point loading
 //! yourself, call [`Entry::from_static_fn()`].
+//!
+//! ## Create features
```
MarijnS95, line 37: That's a typo

ash/src/lib.rs
```diff
@@ -33,6 +33,12 @@
...
+//!
+//! * __debug__ (default): Whether Vulkan structs should implement Debug.
```
MarijnS95, line 39: These are underscores (`__`), not backticks. But as said, if other crates use `**` (same effect as `__`) we should probably too.

lilly-lizard: ah my bad, should have read your comment properly...

ash/src/lib.rs
```diff
@@ -33,6 +33,12 @@
 //! Load the Vulkan library linked at compile time using [`Entry::linked()`], or load it at runtime
 //! using [`Entry::load()`], which uses `libloading`. If you want to perform entry point loading
 //! yourself, call [`Entry::from_static_fn()`].
+//!
+//! ## Crate features
+//!
+//! * **debug** (default): Whether Vulkan structs should implement Debug.
```
MarijnS95, line 39:
```suggestion
//! * **debug** (default): Whether Vulkan structs should implement `Debug`.
```

MarijnS95 (approved):
Thanks!

PR #630, opened 2022-06-04
extensions/khr: Add VK_KHR_device_group_creation

MarijnS95:


ash/src/extensions/khr/device_group_creation.rs
```diff
@@ -0,0 +1,62 @@
...
+        vk::KhrDeviceGroupCreationFn::name()
+    }
+
+    pub fn fp(&self) -> &vk::KhrDeviceGroupCreationFn {
+        &self.fp
+    }
+
+    pub fn device(&self) -> vk::Instance {
+        self.handle
+    }
```
Ralith, line 61: We should probably inline these too, since they're utterly trivial.

Ralith (approved):

ash/src/extensions/khr/device_group_creation.rs
MarijnS95: We haven't done this for any of `ash/src/extensions` yet, shall I set that up in a separate PR?

Ralith: Sure! IIRC I just didn't have the energy to dig through them all, or maybe my regexes missed them, or something.

MarijnS95: There are many so it's going to take some effort to get this one right. `fp()`/`device()`/`name()` are easy, the rest is just a find-replace on `pub unsafe fn` I think.

MarijnS95: But again, I'll do that after the remaining extensions (except the ones blocking on the spec update) are merged, so that I won't miss any.

PR #606, opened 2022-03-29
Inline Default impls and trivial instance/device wrapper methods

Ralith:
I did not inline `load`, any of the wrappers that invoke `read_into_*_vector` helpers, or (for now) extensions methods. In theory this should be a nice clean overhead reduction at runtime. Data on buildtime impact would be great!

PR #631, opened 2022-06-04
extensions/khr: Add VK_KHR_device_group

MarijnS95:
Depends on #629, ~#630~.


ash/src/extensions/khr/device_group_creation.rs
```diff
@@ -0,0 +1,62 @@
...
+        vk::KhrDeviceGroupCreationFn::name()
+    }
+
+    pub fn fp(&self) -> &vk::KhrDeviceGroupCreationFn {
+        &self.fp
+    }
+
+    pub fn device(&self) -> vk::Instance {
+        self.handle
+    }
```
Ralith, line 61: inline

ash/src/extensions/khr/device_group.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub fn new(instance: &Instance, device: &Device) -> Self {
+        let handle = device.handle();
+        let fp = vk::KhrDeviceGroupFn::load(|name| unsafe {
+            mem::transmute(instance.get_device_proc_addr(handle, name.as_ptr()))
+        });
+        Self { handle, fp }
+    }
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetDeviceGroupPeerMemoryFeaturesKHR.html>
+    pub unsafe fn get_device_group_peer_memory_features(
```
Ralith, line 23: Looks like this and everything below is missing inline?

Ralith (changes requested):

Ralith:
> Depends on https://github.com/ash-rs/ash/pull/631

This is #631, though.

MarijnS95:
> > Depends on #631
> 
> This is #631, though.

Good catch, must have been #630.

ash/src/extensions/khr/device_group_creation.rs
MarijnS95: Let's do that in a separate PR, for all `extensions` at once.

ash/src/extensions/khr/device_group.rs
MarijnS95: Same here, let's get it over with separately.

ash/src/extensions/khr/device_group_creation.rs
MarijnS95: Regarding autogenerated code, I had backported #606 as 29cf7c2 but not #602 yet because it inlined together with merging the setters into the structs. Since you pushed that as a separate commit (https://github.com/ash-rs/ash/pull/602/commits/f14baf77702a85de6d416dcc7b273073ea9f6032) it was pretty trivial to backport into a6336a5, let me know if it looks allright.

Ralith: LGTM.

Ralith (approved):

ash/src/extensions/khr/device_group.rs
```diff
@@ -0,0 +1,159 @@
...
+        &self,
+        heap_index: u32,
+        local_device_index: u32,
+        remote_device_index: u32,
+    ) -> vk::PeerMemoryFeatureFlags {
+        let mut peer_memory_features = mem::zeroed();
```
i509VCB, line 32: Would `MaybeUninit` make more sense here and in the other uses?

PR #633, opened 2022-06-05
Inline trivial `Entry` wrapper methods and `impl` functions

MarijnS95:
Looks like these went missing from #606.

@Ralith where do you generally put the cutoff point?  That is, this file has some `Display`/`Error`/`From` `impl`s which could get the same treatment? How about `load_checked()`/`linked()`/`load()`? Anything else in `utils.rs` that we want to cover?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Changed
+
+- Replaced builders with lifetimes/setters directly on Vulkan structs (#602)
+- Inlined struct setters (#602)
+- Inlined `Default` impls and trivial `Instance`/`Device`/`Entry` wrapper methods (#606, #632)
+
 ### Added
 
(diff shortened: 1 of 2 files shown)


Ralith:
My rule of thumb is to `#[inline]` stuff for which the benefit is glaringly obvious (that is to say, for which the body should compile down to at most a handful of trivial instructions). For example, *any* case of an accessor that directly returns a (reference to a) field satisfies this.

> load_checked

Generic functions are de-facto inline due to monomorphization, so I wouldn't bother.

> linked/load

On the surface, these could be a good candidate, since they are trivial. However, they expand to calls to other functions, which are not all obviously trivial, so there may not be significant savings. The low likelihood of them being performance relevant acts as a tiebreaker, so I wouldn't bother.

Ralith (approved):

MarijnS95:
Sounds good.

> However, they expand to calls to other functions, which are not all obviously trivial, so there may not be significant savings.

I remember reading that for those functions to be included in cross-crate optimizations, they themselves have to be marked `#[inline]` as well - so this would only save some trivial extra method calls. But, since they are not as trivial as the others on the surface nor are part of any critical path, let's not bother for now.

PR #638, opened 2022-07-01
extensions: Inline all trivial functions

MarijnS95:
As with earlier PRs inlining public functions with the `#[inline]` attribute is desired to provide function bodies to the linker and allow end-users to get ever so slightly higher performance when their trivial implementations are inlined directly into target code, instead of ending up as an extra indirection through a call and arguments that have to be set up according to a calling convention.

At the same time this seems to not do / postpone certain codegen/optimization passes, resulting in ever so slightly reduced compilation times for the pure `ash` crate (which was part of the reason these have already been applied to other parts of the `ash` codebase already).  This hit is likely only postponed to the linker though, but should still result in a total compiletime reduction (even if it's so small that it is considered negligible) as a user can't possibly use each and every function - which the compiler/linker then won't waste any time on I hope/suppose :)

For actual performance numbers, on the `master` branch at 71d45e46:

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      6.749 s ±  0.076 s    [User: 7.723 s, System: 0.456 s]
  Range (min … max):    6.670 s …  6.872 s    10 runs
```

With this PR directly on top we get an ever so slightly increase of about ±60ms:

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      6.690 s ±  0.051 s    [User: 7.349 s, System: 0.398 s]
  Range (min … max):    6.618 s …  6.768 s    10 runs
```

---

@Ralith it seems you have intentionally omitted (`Instance`) functions that call into `read_into_uninitialized_vector()` in #606, presumably because their bodies are too complex?  Note that those `read_into_*_vector()` functions have generic and would be publicized/inlined when the `#[inline]` attribute is applied here, _as far as I know/understand_.

Undoing that `#[inline]` change for those and every other extension function that calls `read_into_uninitialized_vector()`/`read_into_defaulted_vector()` brings the timings right back to where we started, undoing our ±60ms improvement:

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      6.770 s ±  0.056 s    [User: 7.563 s, System: 0.464 s]
  Range (min … max):    6.710 s …  6.865 s    10 runs
```

Should we profile with an end-user application that actively uses _some_ of these functions, and see how the timings fare?


Ralith:
>  Note that those `read_into_*_vector()` functions have generic and would be publicized/inlined when the `#[inline]` attribute is applied here, as far as I know/understand.

Yes, my feeling is that this represents enough code in aggregate that the benefit is nonobvious.

> Should we profile with an end-user application that actively uses some of these functions, and see how the timings fare?

I think this would be much more reliable data, yeah. `wgpu` would be a good candidate, but a nontrivial direct consumer of `ash` might also be distinctly interesting. In particular I don't think a 60ms difference in compile time is actionable either way on its own.  I'm also happy to see this merged either way, but better data never hurts.

MarijnS95:
I can try against `wgpu`, not sure how soon though. I'd like to try:

1. Without any `#[inline]` at all;
2. The current state of ash on `master`;
3. With this PR;
4. With this PR, with `#[inline]` removed on those non-trivial `read_into_*_vector()` again.

MarijnS95:
With WGPU at https://github.com/gfx-rs/wgpu/commit/c20c86a9481b1585107cad6455652d434849147a:

Note that I made sure that `ash` (and all other deps) are up-to-date/compiled first with a `-w1` warmup run, and didn't clean ash through an extra `-p ash`. We'd need both to understand

1. Without any `#[inline]` at all, at 71d45e46:
   ```console
   $ hyperfine -w1 -p 'cargo clean -p wgpu-hal' 'cargo b -p wgpu-hal --features ash'
   Benchmark 1: cargo b -p wgpu-hal --features ash
     Time (mean ± σ):     576.8 ms ±   7.0 ms    [User: 648.0 ms, System: 87.0 ms]
     Range (min … max):   571.5 ms … 595.8 ms    10 runs
   ```
   Reminder: 71d45e46 compiles in `±6.737s`. Removing all `#[inline]` brings us back to `±7.059s`.
2. The current state of ash on `master` at 71d45e46:
   ```console
   $ hyperfine -w1 -p 'cargo clean -p wgpu-hal' 'cargo b -p wgpu-hal --features ash'
   Benchmark 1: cargo b -p wgpu-hal --features ash
     Time (mean ± σ):     567.2 ms ±   3.1 ms    [User: 650.0 ms, System: 73.7 ms]
     Range (min … max):   562.9 ms … 573.8 ms    10 runs
   ```
   That's clearly a win on `wgpu-hal` alone, even if it isn't much. In other words, we _definitely_ aren't paying for the winds of `#[inline]` on `ash` later during linking.
3. With this PR at e0f5cad7:
   ```console
   $ hyperfine -w1 -p 'cargo clean -p wgpu-hal' 'cargo b -p wgpu-hal --features ash'
   Benchmark 1: cargo b -p wgpu-hal --features ash
     Time (mean ± σ):     569.4 ms ±   3.3 ms    [User: 649.2 ms, System: 76.9 ms]
     Range (min … max):   564.6 ms … 576.9 ms    10 runs
   ```
   This fluctuates a bit between 569-571ms on re-runs, but it's definitely `±2ms` _slower_ than `master.
4. With this PR at e0f5cad7, with `#[inline]` removed on those non-trivial `read_into_*_vector()` again:
   ```console
   $ hyperfine -w1 -p 'cargo clean -p wgpu-hal' 'cargo b -p wgpu-hal --features ash'
   Benchmark 1: cargo b -p wgpu-hal --features ash
     Time (mean ± σ):     567.8 ms ±   2.7 ms    [User: 647.5 ms, System: 78.2 ms]
     Range (min … max):   563.9 ms … 571.8 ms    10 runs
   ```
   And we're back at `master`, showing that it is indeed beneficial to omit `#[inline]` on functions that call `read_into_*_vector()`; if only by a very, _very_ tiny margin.

MarijnS95:
I need to do some clockfixing on my PC, after a suspend-resume this PR in wgpu is much faster again:

```console
$ hyperfine -w1 -p 'cargo clean -p ash -p wgpu-hal' 'cargo b -p wgpu-hal --features ash'
Benchmark 1: cargo b -p wgpu-hal --features ash
  Time (mean ± σ):      6.608 s ±  0.053 s    [User: 7.869 s, System: 0.434 s]
  Range (min … max):    6.552 s …  6.697 s    10 runs
```

The new `enum` push seems to fluctuate around that.

Ralith (approved):
Nice analysis!

PR #640, opened 2022-07-11
Fix invalid data length calculated in `get_query_pool_results()` when called with `WITH_AVAILABILITY` flag

### LINKED ISSUE #639: `get_query_pool_results` buggy when called with `vk::QueryResultFlags::WITH_AVAILABILITY`
Hello,
I have a piece of code which according to my knowledge does not violate the Vulkan spec.
```rust
let first_query = 0;
let query_count = 2;
let flags = vk::QueryResultFlags::WITH_AVAILABILITY;
let mut data = [0u32; 4];
device.get_query_pool_results(
    query_pool,
    first_query,
    query_count,
    data.as_mut_slice(),
    flags,
)?;
```
This code gives a validation error:
```
[VK][Validation]"Validation Error: [ VUID-vkGetQueryPoolResults-dataSize-00817 ] Object 0: handle = 0x13105e0000000155, type = VK_OBJECT_TYPE_QUERY_POOL; | MessageID = 0x73375ed4 | vkGetQueryPoolResults() on querypool VkQueryPool 0x13105e0000000155[] specified dataSize 8 which is incompatible with the specified query type and options. The Vulkan spec states: dataSize must be large enough to contain the result of each query, as described here (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkGetQueryPoolResults-dataSize-00817)"
```

I managed to locate a bug in the implementation of this function. It passes `query_count` as a data length parameter forward. This is fine in most cases but here the flag`vk::QueryResultFlags::WITH_AVAILABILITY` requires the data to be twice as large as `query_count`.
I will open PR with a fix.

matszczygiel:
Fixes #639

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -28,6 +28,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Removed experimental AMD extensions (#607)
 
+### Fixed
+
+-  Fix invalid data length calculated inside of `get_query_pool_results` when called with `WITH_AVAILABILITY` flag (#640)
+
 ## [0.37.0] - 2022-03-23
 
 ### Changed
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -2025,7 +2025,7 @@ impl Device {
             data_length <= data.len(),
             "query_count was higher than the length of the slice"
         );
-        let data_size = mem::size_of::<T>() * data.len();
+        let data_size = mem::size_of_val(data);
         (self.device_fn_1_0.get_query_pool_results)(
             self.handle(),
             query_pool,
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -2025,7 +2025,7 @@ impl Device {
             data_length <= data.len(),
             "query_count was higher than the length of the slice"
         );
-        let data_size = mem::size_of::<T>() * data_length;
+        let data_size = mem::size_of::<T>() * data.len();
         (self.device_fn_1_0.get_query_pool_results)(
             self.handle(),
             query_pool,


Ralith:
Note that we're still passing `mem::size_of::<T>()` as the stride. This means that for `WITH_AVAILABILITY`, `T` should probably be `(u32, vk::Bool32)` or similar, unlike your example in #639. Might bear documenting this explicitly.

MarijnS95:
@Ralith I wondered if it makes sense to perhaps change the function signature to deal specifically with `WITH_AVAILABILITY` and `WITH_STATUS`? From your example I take it the stride should also be `8` (for u32 + bool32) here otherwise subsequent values start overwriting the availability/status result?

Ralith:
`mem::size_of::<(u32, vk::Bool32)>()` is 8, so using that as `T` should Just Work without any additional measures in the interface--even without the change proposed in this PR.

MarijnS95:
> `mem::size_of::<(u32, vk::Bool32)>()` is 8, so using that as `T` should Just Work without any additional measures in the interface--even without the change proposed in this PR.

Meaning that this PR together with the example proposed in #639 doesn't work as intended, since the use of `u32` there results in a stride of `4`? My question is whether query `N+1` would overwrite the `AVAILABILITY` bit of query `N` in that case.
Local testing indicates that that is indeed what happens; the availability result is overwritten by query value `N+1`.

Of course the caller can/should use a `Vec<(u32, vk::Bool32)>` here and immediately receive the type-safe results in a usable manner. Replace `u32` with `u64` for 64-bit results. And that immediately addresses / works around the "missing" `stride` field.

The request to come up with a more specific API is probably moot given that its use may be expanded over time - see `WITH_STATUS_KHR` which is being introduced through the `VK_KHR_video_queue` beta extension. 
We can't check for `T=u32` + `TYPE_64` either, but that's unnecessary and the job of the validation layer.

> I lean towards dropping `query_count` in the spirit of greater type safety personally. Adding (or dynamically computing) `stride` would make it easier to select flags dynamically and decode results after the fact, but it's difficult to imagine a case where that is preferable.

IIRC we discussed and approved/agreed on removing `query_count` before, since the user passes in a slice here and can hence easily change the length.

matszczygiel:
OK. So what seems to be the solution is to just simply use a slice of `(u32, vk::Bool32)`.
Should I close this PR ?

Ralith (approved):

MarijnS95 (changes requested):
@Ralith Recapping what we wrote above... What do we get when we _do not_ merge this change?

I think _not_ merging this change _forces_ the user to use the proper `(u32, vk::Bool32)` type for `T`, and _automatically_ get the right stride in what is currently `mem::size_of::<T>() as _`? As such there's not really anything to fix?

Then separately we can remove `query_count` and utilize slice length.

Ralith:
Fine by me.

PR #644, opened 2022-07-28
device: Replace `query_count` parameter in `get_query_pool_results` with `data.len()`

MarijnS95:
Supersedes #640

A slice is designed to be a cheap view into a continuous segment of memory and easily recreated with a different offset and size.  This can be used to our advantage to let the user set exactly how many items they want to request, while ensuring all returned slots are filled on success or otherwise return `NOT_READY` if not enough query results could be copied to user memory (or block if the `WAIT` flag is set).

This also opens the door to accepting `MaybeUninit` in the future and returning the initialized slice in the `VkResult::Ok(here)` case.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -2016,21 +2016,15 @@ impl Device {
         &self,
         query_pool: vk::QueryPool,
         first_query: u32,
-        query_count: u32,
         data: &mut [T],
         flags: vk::QueryResultFlags,
     ) -> VkResult<()> {
-        let data_length = query_count as usize;
-        assert!(
-            data_length <= data.len(),
-            "query_count was higher than the length of the slice"
-        );
-        let data_size = mem::size_of::<T>() * data_length;
+        let data_size = mem::size_of_val(data);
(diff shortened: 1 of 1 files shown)


ash/src/device.rs
```diff
@@ -2016,21 +2016,15 @@ impl Device {
...
         data: &mut [T],
         flags: vk::QueryResultFlags,
     ) -> VkResult<()> {
-        let data_length = query_count as usize;
-        assert!(
-            data_length <= data.len(),
-            "query_count was higher than the length of the slice"
-        );
-        let data_size = mem::size_of::<T>() * data_length;
+        let data_size = mem::size_of_val(data);
```
MarijnS95, line 2022: This effectively incorporates #640; not to fix a "bug" (there was no bug as far as I understood, the caller was using the wrong the type and doing it that way would still result in the wrong stride hence the wrong output for `WITH_AVAILABILITY` as result N+1 overwrites N), but simply because this otherwise reads `mem::size_of::<T>() * data.len()` which should be exactly what `mem::size_of_val()` returns.

Ralith (approved):

PR #645, opened 2022-07-28
ash-window: Upgrade to raw-window-handle 0.5.0

MarijnS95:
Depends on #505, #643, https://github.com/norse-rs/raw-window-metal/pull/5

### DIFF
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -11,6 +11,16 @@ use raw_window_handle::{
     HasRawDisplayHandle, HasRawWindowHandle, RawDisplayHandle, RawWindowHandle,
 };
 
+pub trait HasRawDisplayAndWindowHandle: HasRawDisplayHandle + HasRawWindowHandle {
+    fn raw_display_and_window_handle(&self) -> (RawDisplayHandle, RawWindowHandle);
+}
+
+impl<T: HasRawDisplayHandle + HasRawWindowHandle> HasRawDisplayAndWindowHandle for T {
+    fn raw_display_and_window_handle(&self) -> (RawDisplayHandle, RawWindowHandle) {
+        (self.raw_display_handle(), self.raw_window_handle())
+    }
+}
+
 /// Create a surface from a raw surface handle.
 ///
 /// `instance` must have created with platform specific surface extensions enabled.
@@ -23,15 +33,10 @@ use raw_window_handle::{
 pub unsafe fn create_surface(
     entry: &Entry,
     instance: &Instance,
-    // TODO: Cannot write `&dyn HasRawDisplayHandle + HasRawWindowHandle :/
-    display_handle: &dyn HasRawDisplayHandle,
-    window_handle: &dyn HasRawWindowHandle,
+    handle: &dyn HasRawDisplayAndWindowHandle,
     allocation_callbacks: Option<&vk::AllocationCallbacks>,
 ) -> VkResult<vk::SurfaceKHR> {
-    match (
-        display_handle.raw_display_handle(),
-        window_handle.raw_window_handle(),
-    ) {
+    match handle.raw_display_and_window_handle() {
         (RawDisplayHandle::Windows(_), RawWindowHandle::Win32(window)) => {
             let surface_desc = vk::Win32SurfaceCreateInfoKHR::default()
                 .hinstance(window.hinstance)
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -103,14 +103,18 @@ pub unsafe fn create_surface(
     }
 }
 
-/// Query the required instance extensions for creating a surface from a window handle.
+/// Query the required instance extensions for creating a surface from a display handle.
+///
+/// This [`HasRawDisplayHandle`] can typically be acquired from a window, but is usually also
+/// accessible earlier through an "event loop" concept to allow querying required instance
+/// extensions and creation of a compatible Vulkan instance prior to creating a window.
 ///
 /// The returned extensions will include all extension dependencies.
 pub fn enumerate_required_extensions(
-    window_handle: &dyn HasRawWindowHandle,
+    display_handle: &dyn HasRawDisplayHandle,
 ) -> VkResult<&'static [*const c_char]> {
-    let extensions = match window_handle.raw_window_handle() {
-        RawWindowHandle::Win32(_) => {
+    let extensions = match display_handle.raw_display_handle() {
+        RawDisplayHandle::Windows(_) => {
             const WINDOWS_EXTS: [*const c_char; 2] = [
                 khr::Surface::name().as_ptr(),
                 khr::Win32Surface::name().as_ptr(),
@@ -118,7 +122,7 @@ pub fn enumerate_required_extensions(
             &WINDOWS_EXTS
         }
 
-        RawWindowHandle::Wayland(_) => {
+        RawDisplayHandle::Wayland(_) => {
             const WAYLAND_EXTS: [*const c_char; 2] = [
                 khr::Surface::name().as_ptr(),
                 khr::WaylandSurface::name().as_ptr(),
@@ -126,7 +130,7 @@ pub fn enumerate_required_extensions(
             &WAYLAND_EXTS
         }
 
-        RawWindowHandle::Xlib(_) => {
+        RawDisplayHandle::Xlib(_) => {
             const XLIB_EXTS: [*const c_char; 2] = [
                 khr::Surface::name().as_ptr(),
                 khr::XlibSurface::name().as_ptr(),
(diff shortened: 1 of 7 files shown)


ash-window/src/lib.rs
```diff
@@ -20,77 +23,59 @@ use ash::extensions::ext; // portability extensions
 pub unsafe fn create_surface(
     entry: &Entry,
     instance: &Instance,
+    // TODO: Cannot write `&dyn HasRawDisplayHandle + HasRawWindowHandle :/
+    display_handle: &dyn HasRawDisplayHandle,
     window_handle: &dyn HasRawWindowHandle,
```
MarijnS95, line 28: @kchibisov what was the idea here?

MarijnS95:
We'll have to wait some for https://github.com/rust-windowing/winit/pull/2400 to back out of the 1.60 MSRV change.

ash-window/src/lib.rs
```diff
@@ -125,22 +110,14 @@ pub fn enumerate_required_extensions(
...
```
msiglreith, line 110: this could be switched to use the display handle instead, removing the requirement for a window - should help in particular with Androids `Suspend`/`Resume` limitations

MarijnS95: That's a nice suggestion! However, that means we should assume that whenever the user can (easily) get to a `HasRawWindowHandle`, they also have access to `HasRawDisplayHandle`. This is true in a "chronological" sense (you need a "display" to create a "window") and for Winit where both `Window` and `EventLoopWindowTarget` (and `EventLoop` which coerces into that) implement both traits.

It doesn't matter for `Android` since `winit::Window` is just an empty shell there that you can create well before receiving `Event::Resumed`.

MarijnS95: One thing to note is that @kchibisov simplified the names somewhat. `AndroiNdk` was generalized to `Android` [because it's only representing the Android platform and not specifically an "NDK" `NativeWindow` pointer (Java `Surface`)](https://github.com/rust-windowing/raw-window-handle/pull/93#discussion_r917000491).

The same was done for `Windows` where there's no separation between `Win32` and `WinRt` in the "display" handle. But as soon as Vulkan support for a `WinRt` handle starts trickling in ([efforts are silent/stale](https://github.com/KhronosGroup/Vulkan-Docs/issues/366)) `ash-window` needs to distinguish them on `enumerate_required_extensions()` as well.

Alas, I see no reason to preempt them now though; we can always update later when it is clear _how_. Just thought we should all be aware of this.

ash-window/src/lib.rs
kchibisov: I think you can combine them, or what is the issue here?

MarijnS95: > what is the issue here?

```rust
error[E0225]: only auto traits can be used as additional traits in a trait object
  --> ash-window/src/lib.rs:26:41
   |
26 |     handle: &(dyn HasRawDisplayHandle + HasRawWindowHandle),
   |                   -------------------   ^^^^^^^^^^^^^^^^^^ additional non-auto trait
   |                   |
   |                   first non-auto trait
   |
   = help: consider creating a new trait with all of these as supertraits and using that trait here instead: `trait NewTrait: HasRawDisplayHandle + HasRawWindowHandle {}`
   = note: auto-traits like `Send` and `Sync` are traits that have special properties; for more information on them, visit <https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits>
```

MarijnS95: This works:

```rust
pub trait RawHandles: HasRawDisplayHandle + HasRawWindowHandle {
    fn get(&self) -> (RawDisplayHandle, RawWindowHandle);
}

impl<T: HasRawDisplayHandle + HasRawWindowHandle> RawHandles for T {
    fn get(&self) -> (RawDisplayHandle, RawWindowHandle) {
        (self.raw_display_handle(), self.raw_window_handle())
    }
}

pub unsafe fn create_surface(
    ...,
    handle: &dyn RawHandles,
    ...,
) {
    match handle.get() {
        (RawDisplayHandle..., RawWindowHandle...) =>
    }
}
```

kchibisov: I think you can `Box` them?

MarijnS95: @kchibisov Same error... But even if it did that defeats the purpose of stuffing a cheap borrow of a `Window` in there as the user now has to _own_ (or clone?) it and temporarily move it into a heap allocation unnecessarily.

I'd then rather change the API to accept `RawDisplayHandle` and `RawWindowHandle` by-value.

msiglreith: switching to static dispatching would be also an option (which wgpu will do https://github.com/gfx-rs/wgpu/blob/ab4df8bcbfd937a9fb527f797590d559c13076bb/wgpu-core/src/instance.rs#L427)

MarijnS95: @msiglreith Cool, there's probably no harm in doing static dispatching in `ash-window` either?

MarijnS95: Hmm, I'm currently using `dyn` dispatch on my side within trait functions (i.e. `trait Renderer { fn create_swapchain(&dyn HasRawWindowHandle, ...) }`), that'd be annoying.

MarijnS95: @msiglreith Fwiw @i509VCB recently (3 days ago) changed this to use the owned/resolved `RawDisplayHandle`/`RawWindowHandle`): https://github.com/gfx-rs/wgpu/commit/9e3efd7019518d372516820c77d4456ae2882d99

That way the caller can simply call `.raw_display_handle()`/`.raw_window_handle()`, without having to worry that - for example - the `Has*` traits are not implemented for those types.

MarijnS95: Implemented it that way now, as originally proposed in https://github.com/ash-rs/ash/pull/645#discussion_r933060153 too, so we're back at static dispatch again at the detriment of requiring the caller to now add/use the `raw-window-handle` crate.

Fortunately these `Raw*Handle` types are `Copy`, so it's perfectly fine to move them around instead of borrowing too.

msiglreith (approved):

ash-window/src/lib.rs
```diff
@@ -21,54 +22,55 @@ use raw_window_handle::{HasRawWindowHandle, RawWindowHandle};
 pub unsafe fn create_surface(
     entry: &Entry,
     instance: &Instance,
-    window_handle: &dyn HasRawWindowHandle,
+    display_handle: RawDisplayHandle,
+    window_handle: RawWindowHandle,
```
Ralith, line 26: Do we need to add a safety precondition that these handles are live and associated with eachother?

MarijnS95: I haven't checked if the API even allows for this.

Ralith: I'm not sure I follow. This is part of ash-window's public API.

MarijnS95: I thought you were requesting to programmatically uphold said constraints, but I think this is a request to simply document it under `# Safety` in the docs no?

i509VCB: Wayland requires the surface must belong to the display. So yes the two must be associated.

MarijnS95: > Wayland requires the surface must belong to the display. So yes the two must be associated.

No way!

MarijnS95: @Ralith English-fu has never been my strong suit, but turned this into:

```rust
/// `window_handle` must be associated with `display_handle` for this call to be valid (e.g. the
/// given `RawWindowHandle` must have been created from the display connection described by
/// `RawDisplayHandle`).
```

Ralith: I'd cut that down to just "`window_handle` must be associated with `display_handle`" for concision, but that seems basically fine. Still need a note that the handle must be currently valid (e.g. not associated with a destroyed window or display connection).

MarijnS95:
```rust
/// The window represented by `window_handle` must be associated with the display connection
/// in `display_handle`.
///
/// `window_handle` and `display_handle` must be associated with a valid window and display
/// connection, that have both not been destroyed.
```

i509VCB: We should probably clarify that the window and display handles don't just need to be valid and not destroyed at creation time, but actually valid for the entire lifetime of the returned Surface.

MarijnS95: Sure, but it's already at a point where one can write a 1500 page manual on this. For the most part `ash`'s stance is that you're familiar with Vulkan and all its rules/caveats anyway - same applies for the window<->surface<->swapchain part IMO.

Ralith (approved):

PR #648, opened 2022-08-11
vk/prelude: Drop upstream-fixed `Self::` doc workaround

MarijnS95:
The report at https://github.com/rust-lang/rust/issues/93205 was closed as it has presumably been fixed in https://github.com/rust-lang/rust/pull/93805 which has long trickled down into stable releases, and I cannot reproduce the issue on `1.62.1` anymore (latest stable as of writing) 🎉

This workaround was originally added in #559.


Ralith:
Nice!

PR #651, opened 2022-08-17
Add optional SmallVec support

JoshBmillikan:
This pull request adds the [smallvec](https://crates.io/crates/smallvecl) crate as an optional dependency and makes most functions that would normally return a `Vec<T>` return a `SmallVec<[T;16]>` when the "smallvec" feature is enabled. Since many ash functions are likely to only return a relatively small number of elements, this can help avoid unnecessary heap allocation.

The API is exactly the same as the existing ash functions, and in fact I mostly copied the existing functions and just modified them to return a SmallVec instead of a Vec. Everything is gated behind the "smallvec" feature flag, so it is completely optional and does not change how ash works unless the feature is enabled.

I currently have chosen 16 as the size of the array for SmallVec, though it might be better to allow the user to choose an array size with a const generic parameter instead. This would be a bit more flexible, but I was trying to keep the API identical to the existing ash functions.

### DIFF
--- a/ash/Cargo.toml
+++ b/ash/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.59.0"
 
 [dependencies]
 libloading = { version = "0.7", optional = true }
-smallvec = { version = "1.9.0", features = ["union", "const_generics", "const_new", "write"], optional = true}
+smallvec = { version = "1.9.0", features = ["union"], optional = true}
 
 [features]
 default = ["loaded", "debug"]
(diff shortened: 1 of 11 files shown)


Ralith:
I like smallvec, but unfortunately changing the type signature of a function is breaking. If multiple crates in a project depend on ash, and one of them enables the feature, it will break the build for all the others. Hence, this cannot be merged. A more viable approach would be for you to offer these functions in a separate crate.

Also note that few (if any) of these functions are liable to be called in a performance-sensitive context, so the downstream benefit is unclear.

PR #656, opened 2022-09-16
extensions/khr: Add VK_KHR_external_fence_capabilities, VK_KHR_external_semaphore_capabilities and VK_KHR_external_memory_capabilities

i509VCB:


Ralith (commented):
These bindings seem fine, but are you sure you need them? They were all promoted to core in 1.1.

i509VCB:
I just added these in case someone needs these in Vulkan 1.0

MarijnS95:
Doesn't seem worth the maintenance effort if no one is actively using these, unlike the other wrappers TBH.

Ralith:
Yeah, let's hold off unless there's specific demand, but thanks for the contribution!

PR #658, opened 2022-09-21
extensions/khr/ray_tracing_pipeline: Set length of capture-replay shader handle buffer after filling

djkoloski:
Because the capacity is allocated but the length of the vector is not set, the vec returned from `get_ray_tracing_capture_replay_shader_group_handles` will always appear to be empty. I checked for other mistakes of this nature and did not locate any.

### DIFF
--- a/ash/src/extensions/khr/ray_tracing_pipeline.rs
+++ b/ash/src/extensions/khr/ray_tracing_pipeline.rs
@@ -113,7 +113,7 @@ impl RayTracingPipeline {
     ) -> VkResult<Vec<u8>> {
         let mut data: Vec<u8> = Vec::with_capacity(data_size);
 
-        (self
+        let err_code = (self
             .fp
             .get_ray_tracing_capture_replay_shader_group_handles_khr)(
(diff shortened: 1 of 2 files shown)


ash/src/extensions/khr/ray_tracing_pipeline.rs
```diff
@@ -113,7 +113,7 @@ impl RayTracingPipeline {
     ) -> VkResult<Vec<u8>> {
         let mut data: Vec<u8> = Vec::with_capacity(data_size);
```
MarijnS95, line 115:
```suggestion
```

MarijnS95 (approved):
Nice catch! That was a bad yet-easy-to-make oopsie in the original Ray Tracing PR :)

Can you add a changelog item for this under a new `### Fixed` designator?

ash/src/extensions/khr/ray_tracing_pipeline.rs
djkoloski: Added!

PR #659, opened 2022-09-22
ash-window: Link to the `ash` gitter instead of nonexistent `ash-window`

MarijnS95:
It seems there's no gitter channel for `ash-window` (404), and since everyone is in the `ash` channel already, just use the same one.

### DIFF
--- a/ash-window/README.md
+++ b/ash-window/README.md
@@ -7,7 +7,7 @@ Interoperability between [`ash`](https://github.com/MaikKlein/ash) and [`raw-win
 [![Build Status](https://github.com/MaikKlein/ash/workflows/CI/badge.svg)](https://github.com/MaikKlein/ash/actions?workflow=CI)
 [![LICENSE](https://img.shields.io/badge/license-MIT-blue.svg)](LICENSE-MIT)
 [![LICENSE](https://img.shields.io/badge/license-apache-blue.svg)](LICENSE-APACHE)
-[![Join the chat at https://gitter.im/MaikKlein/ash-window](https://badges.gitter.im/MaikKlein/ash-window.svg)](https://gitter.im/MaikKlein/ash-window?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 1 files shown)


MarijnS95:
@Ralith Perhaps because `MaikKlein/ash-window` hasn't been an existing GitHub repo? Alas, we moved `ash-rs/ash` and the gitter room needs to be updated for that. @MaikKlein can you pick up the [steps described here](https://gitlab.com/gitterHQ/webapp/-/blob/develop/docs/faq.md#what-happens-if-i-rename-something-on-github-org-repo)?

MaikKlein:
>Perhaps because MaikKlein/ash-window hasn't been an existing GitHub repo? Alas, we moved ash-rs/ash and the gitter room needs to be updated for that. @MaikKlein can you pick up the [steps described here](https://gitlab.com/gitterHQ/webapp/-/blob/develop/docs/faq.md#what-happens-if-i-rename-something-on-github-org-repo)?

Just to be clear. You want me to create a new gitter room for ash-rs? I created https://gitter.im/ash-rs/ash just now. 

The docs say

> Create a new room tied to the repo and we can move the messages over manually. Please note, this will not move room members over.

Is that okay? Just moving messages over and not members? If so I can send the mail to make the transition.

MarijnS95:
> > Perhaps because MaikKlein/ash-window hasn't been an existing GitHub repo? Alas, we moved ash-rs/ash and the gitter room needs to be updated for that. @MaikKlein can you pick up the [steps described here](https://gitlab.com/gitterHQ/webapp/-/blob/develop/docs/faq.md#what-happens-if-i-rename-something-on-github-org-repo)?
> 
> Just to be clear. You want me to create a new gitter room for ash-rs? I created https://gitter.im/ash-rs/ash just now.
> 
> The docs say
> 
> > Create a new room tied to the repo and we can move the messages over manually. Please note, this will not move room members over.
> 
> Is that okay? Just moving messages over and not members? If so I can send the mail to make the transition.

Oops, I read over that and only skimmed the "org rename", which isn't what happened here, I should have read the "transfer" docs instead.

I think that's fine though, there's not nearly enough activity in there to really be worried about existing users. Just make sure to ask if we can post one final "this room has been moved over to https://gitter.im/ash-rs/ash" and then disable message sending for everyone (and perhaps disable joins etc).

PR #661, opened 2022-09-23
extensions/ext: Rename `debug_utils_set_object_name` to `set_debug_utils_object_name` and `debug_utils_set_object_tag` to `set_debug_utils_object_tag` for consistency and deprecate old ones

### LINKED ISSUE #660: `VK_EXT_debug_utils` function inconsistency
Inside the Debug Utils loader, the function is called `debug_utils_set_object_name`, but should be called `set_debug_utils_object_name` in my opinion. Is there a good reason for that?
The same is with `debug_utils_set_object_tag`.

Just wanted to point that out in case it was accidental.

https://github.com/ash-rs/ash/blob/b4fc3340836c9b304c2018324e4f9175052f606a/ash/src/extensions/ext/debug_utils.rs#L24

BeastLe9enD:
Fixes #660

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replaced builders with lifetimes/setters directly on Vulkan structs (#602)
 - Inlined struct setters (#602)
 - Inlined `Default` impls and trivial `Instance`/`Device`/`Entry` wrapper methods (#606, #632)
-- Renamed `debug_utils_set_object_name` and `debug_utils_set_object_tag` for consistency and deprecated old ones (#661)
+- Renamed `debug_utils_set_object_name` to `set_debug_utils_object_name` and `debug_utils_set_object_tag` to `set_debug_utils_object_tag` for consistency and deprecated old ones (#661)
 
 ### Added
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replaced builders with lifetimes/setters directly on Vulkan structs (#602)
 - Inlined struct setters (#602)
 - Inlined `Default` impls and trivial `Instance`/`Device`/`Entry` wrapper methods (#606, #632)
+- Renamed `debug_utils_set_object_name` and `debug_utils_set_object_tag` for consistency and deprecated old ones (#661)
 
 ### Added
 
--- a/ash/src/extensions/ext/debug_utils.rs
+++ b/ash/src/extensions/ext/debug_utils.rs
@@ -20,8 +20,19 @@ impl DebugUtils {
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkSetDebugUtilsObjectNameEXT.html>
+    #[deprecated]
     #[inline]
     pub unsafe fn debug_utils_set_object_name(
+        &self,
+        device: vk::Device,
+        name_info: &vk::DebugUtilsObjectNameInfoEXT,
+    ) -> VkResult<()> {
+        self.set_debug_utils_object_name(device, name_info)
+    }
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkSetDebugUtilsObjectNameEXT.html>
+    #[inline]
+    pub unsafe fn set_debug_utils_object_name(
         &self,
         device: vk::Device,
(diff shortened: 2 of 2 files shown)


ash/src/extensions/ext/debug_utils.rs
```diff
@@ -20,21 +20,43 @@ impl DebugUtils {
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkSetDebugUtilsObjectNameEXT.html>
+    #[deprecated]
```
MarijnS95, line 23: This lacks a message.

MarijnS95 (commented):
Can you fix the title of this PR, add a changelog entry, and target `0.37-stable`? For the `master` branch I'd remove the offending function entirely.

Changelog.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replaced builders with lifetimes/setters directly on Vulkan structs (#602)
 - Inlined struct setters (#602)
 - Inlined `Default` impls and trivial `Instance`/`Device`/`Entry` wrapper methods (#606, #632)
+- Renamed `debug_utils_set_object_name` and `debug_utils_set_object_tag` for consistency and deprecated old ones (#661)
```
MarijnS95, line 14: Renamed to what?

ash/src/extensions/ext/debug_utils.rs
```diff
@@ -20,6 +20,7 @@ impl DebugUtils {
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkSetDebugUtilsObjectNameEXT.html>
+    /// Use `set_debug_utils_object_name` instead for consistency
     #[deprecated]
```
MarijnS95, line 24: `deprecated = "..."`...

MarijnS95:
For the deprecation message, `vk.xml` uses the following: `Backwards-compatible alias containing a typo`. I don't think we should use exactly that, but something like it has my preference.

ash/src/extensions/ext/debug_utils.rs
BeastLe9enD: Oh sorry, I have never used `#[deprecated]` before

MarijnS95:
> Also, can you double-check if intradoc links in deprecation messages now work?

Even after fixing the reference (must be prefixed by `Self::` or `DebugUtils::`) they still don't work:

image

Can you remove the `[]` link brackets around it, and add `()` at the end of the name?

MarijnS95:
@BeastLe9enD Also, after targeting `0.37-stable` feel free to submit a new PR removing the functions entirely for `master` (with appropriate changelog item). I'll sort out the changelog on `master` when we eventually perform a `0.37` release.

ash/src/extensions/ext/debug_utils.rs
```diff
@@ -20,7 +20,7 @@ impl DebugUtils {
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkSetDebugUtilsObjectNameEXT.html>
-    #[deprecated = "Backwards-compatible alias containing a typo, use [`set_debug_utils_object_name`] instead"]
+    #[deprecated = "Backwards-compatible alias containing a typo, use `Self::set_debug_utils_object_name()` instead"]
```
MarijnS95, line 23: Sorry, unclear on my part: drop the `Self::` bit, that's only relevant if this were an intradoc link (which, as we clarified before, doesn't work).

BeastLe9enD: Ok, the thing is I don't know how to check if it works or not since the IDE I'm using does not support it :)

MarijnS95: Nothing supports this, apparently; both `rustdoc` and `rust-analyzer` don't show an intradoc links for this (per screenshot above).

To explain again: backticks render the text as monospace. For intradoc links - when adding `[]` around it - `rustdoc` looks in the current scope for the symbol, but these associate methods are only available under `DebugUtils::` (and because we're "in" `DebugUtils`, `Self` works as an alias to that). However, since intradoc links are not resolved in `deprecated = "..."` messages there shouldn't be `[]` _nor_ `Self::` / `DebugUtils::`, that only looks funky to the user.

MarijnS95 (approved):

MarijnS95:
Since this hasn't been rebased on `0.37-stable` (only the target branch was updated) I'll cherry-pick it to that branch (like I do with any other PR for now, however inconvenient that workflow may be) and push a commit to `master` that removes the deprecated functions entirely.

PR #663, opened 2022-09-25
ban wildcard imports for non-vulkan imports

i509VCB:
The wildcard clean up will help in the future with no_std support to avoid needing to also untangle wildcard imports along with getting the equivalent core and alloc types for imports.

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -365,7 +365,7 @@ mod loaded {
     use std::error::Error;
     use std::fmt;
 
-    use super::*;
+    use super::MissingEntryPoint;
 
     #[derive(Debug)]
     #[cfg_attr(docsrs, doc(cfg(feature = "loaded")))]
--- a/ash/src/lib.rs
+++ b/ash/src/lib.rs
@@ -1,4 +1,5 @@
 #![deny(clippy::use_self)]
+#![deny(clippy::wildcard_imports)]
 #![warn(trivial_casts, trivial_numeric_casts)]
 #![allow(
     clippy::too_many_arguments,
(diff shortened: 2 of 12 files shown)


MarijnS95:
Not a fan of this. You're adding a ton of `#[allow(clippy::wildcard_imports)]` just to flatten out five-or-so `use std::os::raw::*;` statements. I don't see how this helps with `no_std`, just switch these imports to `core::ffi::...`.

Also, why can't this use curly braces (`{}`) to import multiple items like `std::os::raw::{c_char, c_int, c_void}` in one go?

MarijnS95:
Fwiw to guarantee `no_std` support we'd just add a CI step that validates it?

i509VCB:
> Not a fan of this. You're adding a ton of `#[allow(clippy::wildcard_imports)]` just to flatten out five-or-so `use std::os::raw::*;` statements.

I'll probably remove the lint changes.

> I don't see how this helps with `no_std`, just switch these imports to `core::ffi::...`.

If I did the easy route like that, ash would need to have the MSRV changed to 1.64. The only way to keep ash with 1.59 is to use a buildscript only feature and reexport the ffi module from std or core at the crate root.

> Also, why can't this use curly braces (`{}`) to import multiple items like `std::os::raw::{c_char, c_int, c_void}` in one go?

Ash has no single convention for this. I just seemed to use multiple items because that is what appeared in most of the extension wrappers.

MarijnS95:
> If I did the easy route like that, ash would need to have the MSRV changed to 1.64. The only way to keep ash with 1.59 is to use a buildscript only feature and reexport the ffi module from std or core at the crate root.

This is about wildcard imports, isn't it? You can still have those with `#[cfg(...)] use ...` regardless of whether that's buildscript driven?

> Ash has no single convention for this. I just seemed to use multiple items because that is what appeared in most of the extension wrappers.

Prefer the shorter, concise version. The extension wrappers (at least those added recently) seem to mostly import modules into scope anyway.

i509VCB:
I will deal with this in the proper no_std pull request. I think I have a better approach.

PR #664, opened 2022-09-25
Add `no_std` support

i509VCB:
no_std support is enabled by disabling the default `std` feature.


.github/workflows/ci.yml
```diff
@@ -29,6 +29,17 @@ jobs:
           command: check
           args: --workspace --all-targets --all-features
 
+  check_no_std:
+    name: Check no_std
```
MarijnS95, line 33: _and_ `1.64` MSRV.

.github/workflows/ci.yml
```diff
@@ -29,6 +29,17 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: dtolnay/rust-toolchain@1.64.0
+      - uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --package=ash --no-default-features --features=debug,linked
```
MarijnS95, line 41: To match the format of the rest of this file:

```suggestion
          args: -p ash --no-default-features --features=debug,linked
```

Changelog.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added new default `std` feature. Disabling this feature makes ash no_std capable and raises the MSRV to 1.64
```
MarijnS95, line 18: PR identifier? And as evident from the other entries, these are mostly chronological / ordered by PR number.

README.md
```diff
@@ -223,6 +223,10 @@ The default `loaded` cargo feature will dynamically load the default Vulkan libr
 
 If, on the other hand, your application cannot handle Vulkan being missing at runtime, you can instead enable the `linked` feature, which will link your binary with the Vulkan loader directly and expose the infallible `Entry::linked`.
 
+### No std
+
+Ash can be used in no_std environments (with alloc). Disabling the default `std` feature. Note: this will raise the MSRV to 1.64
```
MarijnS95, line 228:
```suggestion
### Use in `no_std` environments

Ash can be used in `no_std` environments (with `alloc`) by disabling the default `std` feature. This [raises MSRV to 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#c-compatible-ffi-types-in-core-and-alloc).
```

ash-window/Cargo.toml
```diff
@@ -27,4 +27,4 @@ ash = { path = "../ash", version = "0.37", default-features = false, features =
 
 [[example]]
 name = "winit"
-required-features = ["ash/linked"]
+required-features = ["ash/linked", "ash/std"]
```
MarijnS95, line 30: This doesn't make the example work out-of-the-box:

```console
$ cargo r -p ash-window --example winit
error: target `winit` in package `ash-window` requires the features: `ash/linked`, `ash/std`
Consider enabling them by passing, e.g., `--features="ash/linked ash/std"`
```

Just like `"linked"`, add `"std"` to the list of features for the `ash` dependency...

ash/Cargo.toml
```diff
@@ -15,11 +15,13 @@ rust-version = "1.59.0"
 libloading = { version = "0.7", optional = true }
 
 [features]
-default = ["loaded", "debug"]
+default = ["loaded", "debug", "std"]
+# Whether ash depends on the standard library. Disabiling this raises the MSRV to 1.64
```
MarijnS95, line 19: `Disabiling`?

```suggestion
# Whether ash depends on the standard library. Disabling this raises the MSRV to 1.64.
```

And add a period at the end to match the other feature-flag comments.

ash/src/device.rs
```diff
@@ -2,9 +2,15 @@
...
-use std::ptr;
+use core::mem;
+use core::ptr;
+
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+#[cfg(not(feature = "std"))]
+use core::ffi::{c_void, CStr};
+#[cfg(feature = "std")]
+use std::ffi::{c_void, CStr};
```
MarijnS95, line 13: Can you document with a comment that this nastiness is only required prior to 1.64 (`std::ffi::` is just a reexport on `core::ffi::` after), that way we can remove it when MSRV is inevitably bumped at some point down the road.
(Everywhere, not just in the generator)

FWIW I wouldn't mind doing this _once_ in a prelude (existing or new module) and importing from that.

MarijnS95 (changes requested):
This seems to be in early stages, I'll mark this as WIP until you deem it ready to review; here are a few basal points I started writing on while realizing that.

MarijnS95:
@i509VCB is this still something you want to pursue?

ash-window/Cargo.toml
i509VCB: Though having ash-window enable `ash/std` would be valid, it would force enable `std` just be depending on ash-window.

I think the proper solution here is to give ash-window a default `std` feature.

ash-window can be no-std capable since raw-window-handle is no-std.

raw-window-metal however requires std which complicates this a bit.

ash/src/lib.rs
```diff
@@ -59,6 +65,8 @@ pub mod vk;
 /// Wrappers for Vulkan extensions
 pub mod extensions;
 
+pub(crate) mod prelude_internal;
```
i509VCB, line 68: really could do with a better name

i509VCB:
other than the issue with `ash-window`, the rest of the pull request is ready.

ash-window/Cargo.toml
MarijnS95: > Though having ash-window enable `ash/std` would be valid, it would force enable `std` just be depending on ash-window.

That's why you have to add it to `DEV-dependencies`, next to `"linked"` like I said (and **not** in the `[dependencies]` table like you did now).

However, it'd be great if `ash-window` could be `no_std` capable too.

> raw-window-metal however requires std which complicates this a bit.

Can that be addressed? @msiglreith owns that crate so we can iterate quite easily.

ash/Cargo.toml
```diff
@@ -15,11 +15,13 @@ rust-version = "1.59.0"
...
 [features]
-default = ["loaded", "debug"]
+default = ["loaded", "debug", "std"]
+# Whether ash depends on the standard library. Disabling this raises the MSRV to 1.64.
+std = []
 # Link the Vulkan loader at compile time.
 linked = []
 # Support searching for the Vulkan loader manually at runtime.
-loaded = ["libloading"]
+loaded = ["std", "libloading"]
```
MarijnS95, line 24: Libloading and/or our logic requires `std` because of the `Arc`?

ash/src/device.rs
MarijnS95: As requested, can you document this? Once we bump MSRV for the non-`no_std` case to 1.64 too we can just get rid of the `std::ffi` import altogether and unconditionally use `std::ffi`.

ash/src/vk/platform_types.rs
```diff
@@ -1,5 +1,8 @@
 #![allow(non_camel_case_types)]
 
+#[cfg(not(feature = "std"))]
+use core::ffi::*;
+#[cfg(feature = "std")]
 use std::os::raw::*;
```
MarijnS95, line 6: Use prelude?

ash/src/device.rs
MarijnS95, line 13:
```suggestion
use crate::prelude_internal::{c_void, CStr};
```

ash/src/instance.rs
```diff
@@ -2,9 +2,17 @@ use crate::device::Device;
...
+use core::ptr;
+
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+#[cfg(not(feature = "std"))]
+use core::ffi::{c_char, CStr};
+#[cfg(feature = "std")]
+use std::ffi::CStr;
+#[cfg(feature = "std")]
 use std::os::raw::c_char;
```
MarijnS95, line 15:
```suggestion
use crate::prelude_internal::{c_char, CStr};
```

ash/src/lib.rs
MarijnS95: Yeah, I was thinking the same, not sure what to call it though. Perhaps folding this in with `prelude` and only exporting these items as `pub(crate)` might be enough?

After all 60% of `ash/src/prelude.rs` (_not_ `ash/src/vk/prelude.rs`) is already `pub(crate)` for internal use only.

But if you do, a submodule of that prelude might be nicest, since it's a wildcard reexport.

ash/src/lib.rs
```diff
@@ -40,6 +42,10 @@
 //! * **loaded** (default): Support searching for the Vulkan loader manually at runtime.
 //! * **linked**: Link the Vulkan loader at compile time.
 
+#[cfg(not(feature = "std"))]
+#[macro_use]
+extern crate alloc;
```
MarijnS95, line 47: Are allocations something that we might later want to make entirely optional (separate change/PR)?

ash/src/util.rs
```diff
@@ -28,10 +30,9 @@ pub struct AlignIter<'a, T: 'a> {
 
 impl<T: Copy> Align<T> {
     pub fn copy_from_slice(&mut self, slice: &[T]) {
-        use std::slice::from_raw_parts_mut;
         if self.elem_size == size_of::<T>() as u64 {
             unsafe {
-                let mapped_slice = from_raw_parts_mut(self.ptr.cast(), slice.len());
+                let mapped_slice = slice::from_raw_parts_mut(self.ptr as *mut T, slice.len());
```
MarijnS95, line 35: We just cleaned these loose `as` casts up :angry:

MarijnS95 (changes requested):

ash/src/extensions/ext/acquire_drm_display.rs
```diff
@@ -1,8 +1,8 @@
 use crate::prelude::*;
+use crate::prelude_internal::CStr;
```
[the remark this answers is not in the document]
i509VCB: I guess I forgot

ash/src/entry.rs
```diff
@@ -2,19 +2,25 @@ use crate::instance::Instance;
...
+use core::ptr;
+
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+#[cfg(not(feature = "std"))]
+use core::ffi::{c_char, c_void, CStr};
+#[cfg(feature = "std")]
 use std::ffi::CStr;
+#[cfg(feature = "std")]
+use std::os::raw::{c_char, c_void};
```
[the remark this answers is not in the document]
i509VCB: Forgot

ash/src/util.rs
i509VCB: I recall some clippy or rust lint exists to prevent these from creeping back in?

ash/src/lib.rs
i509VCB: Possibly.

This would however require radically changing how ash approaches the two call pattern in Vulkan unless we want to have those not using alloc call the raw vulkan functions themselves. It's a much larger discussion that needs it's own issue.

For now I required alloc so the same ash api works with no_std

ash/Cargo.toml
i509VCB: libloading requires `std`, `Arc` is actually in `alloc` assuming your system has the requisite atomics support

ash/src/extensions/ext/acquire_drm_display.rs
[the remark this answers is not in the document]
MarijnS95: Ack, I thought you said it was ready.

As a general reminder, self-review your PR after creating/updating it, instead of putting that burden on the reviewer/maintainer(s).

ash/src/lib.rs
MarijnS95: Not now, but we could do it later. You are right that the two-call-pattern becomes increasingly complex when we do this, effectively having to split all those functions in two.

ash/src/util.rs
MarijnS95: Perhaps [`ptr_as_ptr`](https://rust-lang.github.io/rust-clippy/master/#ptr_as_ptr) might help us with this but I don't recall seriously looking into it via #685. With that, would you still be able to implement this using `core`? Since this cast function is available on the `core` primitive pointer type?

MarijnS95: I think we cannot use this because there are a few `p_next` variants that store `*const T` in a `*mut T`, and `cast_mut()` hasn't been stable for long enough.

ash-window/Cargo.toml
msiglreith: I quickly looked into it and seems a bit tricky due to `objc` dependency

nical:
Note that Gecko is currently at rustc 1.63 let's please wait until Gecko updates to 1.64 before raising the msrv. Gecko tends to lag behind the latest stable a bit but usually not too much, so it shouldn't be long.

MarijnS95:
@nical Ash won't blindly bump its MSRV, at least not in regular patch releases.

nical:
Sorry, I got confused, I though I was on the wgpu repo. That said, bumping the msrv will prevent wgpu from updating ash until gecko catches up.

MarijnS95:
> It would be great if a version containing #695 would be published before an msrv bump.

Again, non-breaking releases with bugfixes and improvements won't include an MSRV bump (that makes them... breaking...). That said, I'd like to be requested new releases explicitly, otherwise they might sit for a while.

i509VCB:
> Note that Gecko is currently at rustc 1.63 let's please wait until Gecko updates to 1.64 before raising the msrv. Gecko tends to lag behind the latest stable a bit but usually not too much, so it shouldn't be long.

The MSRV bump here only applies if the std feature is disabled. So ash effectively has an std MSRV and an alloc MSRV

MarijnS95:
@nical fwiw #695 was published in https://github.com/ash-rs/ash/releases/tag/0.37.2.

MarijnS95:
@nical it looks like `wgpu` meanwhile has an MSRV of 1.65, and I https://github.com/gfx-rs/wgpu#msrv-policy, and we should be well okay bumping to or past 1.64 IMHO (that's older than half a year - even though we didn't document our policy explicitly yet).

@i509VCB is it time to revive this on 1.64, and drop all the workarounds for older MSRV?

nical:
Yes, thanks a lot for being mindful of wgpu's constraints!

i509VCB:
> @nical it looks like `wgpu` meanwhile has an MSRV of 1.65 (https://github.com/gfx-rs/wgpu#msrv-policy), and we should be well okay bumping to or past 1.64 IMHO (that's older than half a year - even though we didn't document our policy explicitly yet).
> 
> @i509VCB is it time to revive this on 1.64, and drop all the workarounds for older MSRV?

Yeah I can reapply the changes into 1.64, just not going to be today.

i509VCB:
@MarijnS95 I have rebuilt the pull request. Ash appears to have raised it's MSRV to 1.69 already, so I went with that.

ash/src/extensions/nv/ray_tracing.rs
```diff
@@ -208,7 +209,7 @@ impl RayTracing {
         (self.fp.get_acceleration_structure_handle_nv)(
             self.handle,
             accel_struct,
-            std::mem::size_of_val(&handle),
+            core::mem::size_of_val(&handle),
```
MarijnS95, line 212:
```suggestion
            mem::size_of_val(&handle),
```

ash/src/entry.rs
```diff
@@ -337,7 +338,7 @@ impl vk::StaticFn {
                 if val.is_null() {
                     return Err(MissingEntryPoint);
                 } else {
-                    ::std::mem::transmute(val)
+                    ::core::mem::transmute(val)
```
MarijnS95, line 341:
```suggestion
                    mem::transmute(val)
```

MarijnS95 (commented):
Any clue when `unnecessary_qualifications` broke?

ash/src/lib.rs
```diff
@@ -45,6 +50,11 @@
 //! * **debug** (default): Whether Vulkan structs should implement `Debug`.
 //! * **loaded** (default): Support searching for the Vulkan loader manually at runtime.
 //! * **linked**: Link the Vulkan loader at compile time.
+//! * **std** (default): Whether ash depends on the standard library (otherwise alloc is required)
+
+#[cfg_attr(feature = "std", allow(unused_extern_crates))]
+#[macro_use]
```
Ralith, line 56: Is `#[macro_use]` actually required here?

ash/src/lib.rs
```diff
@@ -45,6 +50,11 @@
 //! * **debug** (default): Whether Vulkan structs should implement `Debug`.
 //! * **loaded** (default): Support searching for the Vulkan loader manually at runtime.
 //! * **linked**: Link the Vulkan loader at compile time.
+//! * **std** (default): Whether ash depends on the standard library (otherwise alloc is required)
+
+#[cfg_attr(feature = "std", allow(unused_extern_crates))]
```
Ralith, line 55: Is this `allow` needed? Looks like we have unconditional references to `alloc::` scattered around, which is the norm for no_std crates.

Ralith (approved):

ash/src/lib.rs
MarijnS95: That makes `vec![]` available without qualified `alloc::vec![]` references or imports.

How about adding `use alloc::vec` across the code-base? There's only one non-`std` use of the macro anyway.

Ralith: Grep tells me there's exactly three callsites for that, so `use`ing/qualifying seems fine.

ash/src/entry.rs
```diff
@@ -351,7 +352,8 @@ impl fmt::Display for MissingEntryPoint {
         write!(f, "Cannot load `vkGetInstanceProcAddr` symbol from library")
     }
 }
-impl Error for MissingEntryPoint {}
+#[cfg(feature = "std")]
+impl std::error::Error for MissingEntryPoint {}
```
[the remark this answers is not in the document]
Ralith: I see no reason not to, though I wouldn't stress tracking every little unstable feature we might benefit from.

ash/src/lib.rs
```diff
@@ -182,6 +190,7 @@ macro_rules! match_in_struct {
 #[cfg(test)]
 mod tests {
     use super::vk;
+    use alloc::vec::Vec;
     #[test]
     fn test_ptr_chains() {
         let mut variable_pointers = vk::PhysicalDeviceVariablePointerFeatures::default();
```
MarijnS95, line 196: The test code below also calls `vec![]`, I'm still surprised why I can run `cargo t` with and without the _default_ `std` feature _without_ importing `use alloc::vec`, qualifying `alloc::vec![]`, _nor_ `#[macro_use]`... :confused:

Ralith: Yeah, I think it really likes to give you std implicitly. Not sure what the best way to test this really is -- for hecs someone contributed a CI step that sets `--target x86_64-unknown-none`, which is probably reliable but seems like a big hammer.

MarijnS95: Not really interested to use this test as a `no_std` hammer, just curious why it works when it "shouldn't" :)

MarijnS95 (approved):

PR #667, opened 2022-09-29
Update Vulkan-Headers to 1.3.235

MarijnS95:
~Depends on #666~


Ralith (approved):
Approved conditional on confirmation that the video extensions are still intended to be semver-exempt (do we document this anywhere?).

MarijnS95:
Yeah, it's semver-exempt but looks like we (I :eyes:) didn't even document it anywhere outside of #417, not even in the README...

Ralith:
Bonus points for adding docs and even a feature gate, then.

MarijnS95:
@Ralith Done in `README.md`! Conciseness isn't really my thing, feel free to suggest improvements.

While at it, maybe we should soup-up our module docs as well :sweat_smile:

Ralith (approved):

MarijnS95:
Rebased on GH and of course, needed some manual attention to integrate cleaned-up casts from #685.

PR #668, opened 2022-10-11
extensions/ext: Add VK_EXT_acquire_drm_display

AidoP:
Note that https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4136 prevents the use of `vkGetDrmDisplayEXT` while using  `VK_LAYER_KHRONOS_validation`.

### DIFF
--- a/ash/src/extensions/ext/acquire_drm_display.rs
+++ b/ash/src/extensions/ext/acquire_drm_display.rs
@@ -26,14 +26,9 @@ impl AcquireDrmDisplay {
         &self,
         physical_device: vk::PhysicalDevice,
         drm_fd: i32,
-        display: vk::DisplayKHR
+        display: vk::DisplayKHR,
     ) -> VkResult<()> {
-        (self.fp.acquire_drm_display_ext)(
-            physical_device,
-            drm_fd,
-            display
-        )
-        .result()
+        (self.fp.acquire_drm_display_ext)(physical_device, drm_fd, display).result()
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetDrmDisplayEXT.html>
@@ -42,16 +37,11 @@ impl AcquireDrmDisplay {
         &self,
         physical_device: vk::PhysicalDevice,
         drm_fd: i32,
-        connector_id: u32
+        connector_id: u32,
     ) -> VkResult<vk::DisplayKHR> {
         let mut display = mem::MaybeUninit::uninit();
-        (self.fp.get_drm_display_ext)(
-            physical_device,
-            drm_fd,
-            connector_id,
-            display.as_mut_ptr()
-        )
-        .result_with_success(display.assume_init())
+        (self.fp.get_drm_display_ext)(physical_device, drm_fd, connector_id, display.as_mut_ptr())
+            .result_with_success(display.assume_init())
     }
 
     #[inline]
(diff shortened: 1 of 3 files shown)


ash/src/extensions/ext/acquire_drm_display.rs
```diff
@@ -0,0 +1,61 @@
+use crate::prelude::*;
+use crate::vk;
+use crate::{Entry, Instance};
+use std::ffi::CStr;
+use std::mem;
+
+/// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_EXT_acquire_drm_display.html>
+#[derive(Clone)]
+pub struct AcquireDrmDisplay {
+    handle: vk::Instance,
```
MarijnS95, line 10: No need to store `handle` if it isn't used by any of the functions wrapped by this struct. In turn drop `fn instance()` as well.

AidoP: removed in d855ed1

ash/src/extensions/ext/acquire_drm_display.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetDrmDisplayEXT.html>
+    #[inline]
+    pub unsafe fn get_drm_display(
+        &self,
+        physical_device: vk::PhysicalDevice,
+        drm_fd: i32,
+        connector_id: u32,
+    ) -> VkResult<vk::DisplayKHR> {
+        let mut display = mem::MaybeUninit::uninit();
```
MarijnS95, line 41: We've only sparsely used `MaybeUninit` before (only ever in `VK_KHR_display`), and only with `zeroed()` instead of `uninit()`.

Doesn't seem like there's any harm in doing that here, so I'm willing to try that out (and hopefully schedule to replace more `MaybeUninit::zeroed()`/`mem::zeroed()`/`default()` calls at some point).

MarijnS95 (approved):
Thanks for taking the other extension implementations by example, making this super easy to review!

ash/src/extensions/ext/acquire_drm_display.rs
```diff
@@ -0,0 +1,55 @@
...
+        (self.fp.get_drm_display_ext)(physical_device, drm_fd, connector_id, display.as_mut_ptr())
+            .result_with_success(display.assume_init())
```
Ralith, line 43: This is unsound because it calls `assume_init()` even if `get_drm_display_ext` returned an error, in which case the data is not necessarily initialized.

MarijnS95: Maybe that's why we'd been using `mem::zeroed()` before?

Ralith: I think we've mostly used it because it's less error-prone. It's perfectly reasonable to use `MaybeUninit`, it just requires a slightly different pattern where the result is only read on success.

MarijnS95: We can add a `.result_with_success` version that takes a closure instead?

Ralith: Right, that'd work well.

MarijnS95: @Ralith Taken care of in https://github.com/ash-rs/ash/pull/669.

PR #671, opened 2022-10-17
extensions/ext: Add VK_EXT_extended_dynamic_state3

MarijnS95:
Depends on #667, cc @wrightwriter


ash/src/extensions/ext/extended_dynamic_state3.rs
```diff
@@ -0,0 +1,405 @@
...
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCmdSetSampleMaskEXT.html>
+    #[inline]
+    pub unsafe fn cmd_set_sample_mask(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        samples: vk::SampleCountFlags,
+        sample_mask: &[vk::SampleMask],
+    ) {
+        // TODO: Assert on the size of sample_mask vs samples, or derive `samples` from slice length?
+        (self.fp.cmd_set_sample_mask_ext)(command_buffer, samples, sample_mask.as_ptr())
```
MarijnS95, line 69: Unsure about this one. We don't do it for the struct constructor because we can't, but here we could/should?

Also, `SampleCountFlags` is a bitflags of powers of two, theoretically multiple bitflags could be specified making it a regular number? Is that even possible/allowed/sensible? (I've never used this part of the API)

ash/src/extensions/ext/extended_dynamic_state3.rs
```diff
@@ -0,0 +1,405 @@
...
+        (self.fp.cmd_set_logic_op_enable_ext)(command_buffer, logic_op_enable.into())
+    }
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCmdSetColorBlendEnableEXT.html>
+    #[inline]
+    pub unsafe fn cmd_set_color_blend_enable(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        first_attachment: u32,
+        color_blend_enables: &[vk::Bool32],
```
MarijnS95, line 111: Only place where we can't do `&[bool]`. Requesting a `vk::Bool32` saves an unnecessary iteration and collect, which we anyway only do in `acceleration_structure.rs`.

ash/src/extensions/ext/extended_dynamic_state3.rs
filnet: Multiple bits can be set but not in this particular context.
See https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSampleCountFlagBits.html#_c_specification

filnet: This function is supposed to take a `VkSampleCountFlagBits` and not a `VkSampleCountFlag`. Was that change deliberate ?

filnet: PS: same issue with `vkCmdSetRasterizationSamples` ?

filnet: Please ignore my comments... There is no `VkSampleCountFlagBits` vs `VkSampleCountFlags` in ash. Both have been conflated in `SampleCountFlags` which acts both as an enum and a mask.

MarijnS95: So the question is: should `ash` treat these as separate entities _if_ `vk.xml` properly uses `*Flags` or `*FlagBits` to distinguish variables where only a single flag is allowed (i.e. some API/function call) versus variables where multiple flags can be set (typically where allowed flags are queried for a specific implementation).

coffeenotfound:
When is this going to be merged? I would really need ext_dynamic_state3

MarijnS95:
@coffeenotfound Whenever it receives review.

ash/src/extensions/ext/extended_dynamic_state3.rs
Ralith: LGTM.

Ralith (approved):

wrightwriter:
i'm wondering if it makes sense to have the dynamic state extensions as a traits on device.

MarijnS95:
Hardly any of the functions take a `VkDevice`, I think you're more so intending this to be on `VkCommandBuffer`.

Unless you're intending for a model where `ash::Device` loads _all_ device functions, I think Erupt does something like that.

wrightwriter:
Ok, my logic was that core vkCmd are on the device. But what you are saying makes sense.

MarijnS95:
Most of these functions are loaded _for_ a device (and `vk::CommandBuffer` is a special child specific to a single `vk::Device`), and all the standardized `.cmd_` function are hence loaded for and available on a device.

The main point is that we didn't force this for extensions, so that the `ash::Device` object doesn't get too bloated with _all_ extensions and their optionally-loaded/available functions.

PR #676, opened 2022-11-07
generator: Update list of vendor names

charles-lunarg:
Noticed that the vendor list is a bit on the lean side. I'm not sure what impact this will have (in fact, I didn't even run the code because I wasn't sure how to set up the rust environment to test it, and it seems like a 'no brain required' change.)

Regardless, while I would of liked to address the TODO to locate the vendors from the code directly, my rust isn't capable of that at the moment. Rather than ask one of the ash maintainers to do it, this lets any new vendor extensions not cause ash builds to fail.


MarijnS95:
> I'm not sure what impact this will have

This list is only used while stripping enum variant types (`FlagBits`), and my general stance has been to not add vendors until they start shipping flags in `vk.xml`; we'd get a clear generator error otherwise which is really only relevant for those who update the bindings (also me).

charles-lunarg:
That makes sense. 

I did make this change due to an upcoming additional vendor adding flag bits, but since its not in the public vk.xml I didn't want to add *only* the one that is needed. 

Should I close the PR? seems this commit goes against the established policy for this repo.

MarijnS95:
@charles-lunarg That is the kind of issue I had in mind, after adding `ash` to the Khronos CI. Happy to pull this in to unblock upstream, there's no real gain in not listing all the vendors here however.

MarijnS95:
Ugh, just saw this in another PR half an hour ago: the freshly released `bytemuck` crate bumps our MSRV to 1.60: https://github.com/Lokathor/bytemuck/pull/140#issuecomment-1304404474

I hope it's only the `generator` though, allow me to fix the CI first to _only_ lint our published crates as the `generator` can rely on `nightly` for what I'm concerned.

charles-lunarg:
I don't think this is *blocking* upstream per-say, as the upstream CI checks don't fail if ash fails, it just documents that it didn't succeed. (Aka, a warning not an error).

I will fix the formatting error. I'll just copy/paste what the formatter tells me it should be. But I'll wait till the bytemuck issue is resolved :)

MarijnS95:
Sure, I set it up that way to not truly block everyone, but seeing a red/orange "CI failed with warnings" is never useful/nice :wink: 

Just run `cargo fmt` locally, set up your Rust environment while we eagerly await more contributions :partying_face:

MarijnS95:
@charles-lunarg You can rebase now :)

MarijnS95 (approved):
Good, but one request: can we sort these alphabetically? :grimacing:

charles-lunarg:
Alphabetical order is no issue, and it is much more pleasing to look at.

MarijnS95:
@charles-lunarg Upstream CI should now be unblocked, which picks directly from the `master` branch.

charles-lunarg:
Thanks!

PR #679, opened 2022-11-18
extensions/ext: Add VK_EXT_descriptor_buffer

MarijnS95:
Depends on #667

### DIFF
--- a/ash/src/extensions/ext/descriptor_buffer.rs
+++ b/ash/src/extensions/ext/descriptor_buffer.rs
@@ -1,0 +1,211 @@
+use crate::prelude::*;
+use crate::vk;
+use crate::{Device, Instance};
+use std::ffi::CStr;
+use std::mem;
+
+/// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_EXT_descriptor_buffer.html>
+#[derive(Clone)]
+pub struct DescriptorBuffer {
+    handle: vk::Device,
+    fp: vk::ExtDescriptorBufferFn,
+}
+
+impl DescriptorBuffer {
+    pub fn new(instance: &Instance, device: &Device) -> Self {
+        let handle = device.handle();
+        let fp = vk::ExtDescriptorBufferFn::load(|name| unsafe {
+            mem::transmute(instance.get_device_proc_addr(handle, name.as_ptr()))
+        });
+        Self { handle, fp }
+    }
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetDescriptorSetLayoutSizeEXT.html>
+    #[inline]
+    pub unsafe fn get_descriptor_set_layout_size(
+        &self,
+        layout: vk::DescriptorSetLayout,
+    ) -> vk::DeviceSize {
+        let mut count = 0;
+        (self.fp.get_descriptor_set_layout_size_ext)(self.handle, layout, &mut count);
+        count
+    }
+
+    /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetDescriptorSetLayoutBindingOffsetEXT.html>
+    #[inline]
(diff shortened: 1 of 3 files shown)


ash/src/extensions/ext/descriptor_buffer.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub unsafe fn get_descriptor(
+        &self,
+        descriptor_info: &vk::DescriptorGetInfoEXT,
+        descriptor: &mut [u8],
+    ) {
+        (self.fp.get_descriptor_ext)(
+            self.handle,
+            descriptor_info,
+            descriptor.len(),
+            descriptor.as_mut_ptr().cast(),
```
MarijnS95, line 75: For this and other functions in this API: we've always been using `u8` types, and even allocating a `Vec` for the user while receiving a `data_size: usize` argument. Not sure if we should keep following that pattern here, or allow the user to perform the allocation.

They may not even have to initialize it, it's perhaps even better to accept `&mut MaybeUninit<[u8]>`?

Separately we can perform a breaking change on main with this, if people feel strongly about cutting down on allocations within Ash.

ash/src/extensions/ext/descriptor_buffer.rs
```diff
@@ -0,0 +1,224 @@
...
+        instance: &Instance,
+        pdevice: vk::PhysicalDevice,
+    ) -> vk::PhysicalDeviceDescriptorBufferPropertiesEXT {
+        let mut props_db = vk::PhysicalDeviceDescriptorBufferPropertiesEXT::default();
+        {
+            let mut props = vk::PhysicalDeviceProperties2::default().push_next(&mut props_db);
+            instance.get_physical_device_properties2(pdevice, &mut props);
+        }
+        props_db
+    }
```
MarijnS95, line 34: After checking, most of the extensions _do not_ implement this helper. I feel like we should push users to _not_ call it in favour of retrieving all their needed properties in a single up-front call with all the necessary structures.

Shall I remove them as a breaking change on `main`?

Ralith: I wasn't aware we had any of these, and I agree that a single call makes more sense, if only because it's comparably ergonomic.

MarijnS95: Removed from this PR, will set up a PR (that obviously won't be backported to 0.37) to remove the rest.

ash/src/extensions/ext/descriptor_buffer.rs
MarijnS95: > I think taking passed-in memory is more in keeping with the spirit of the API in this instance. Note in particular that we don't have the size of memory written to readily available here. 

We have the exact same situation in `ray_tracing_pipeline.rs` with `get_ray_tracing_(capture_replay_)shader_group_handles` where an additional `data_size: usize` is passed in, we can replace those too as a breaking change. All other `Vec::with_capacity` instances aren't returning raw data but known structs, and have their size deducible from other arguments (i.e. `CreateInfo` slices).

> `&mut [MaybeUninit<u8>]` (note _inside_ the slice) would be sensible.

Good catch, yeah the slice size nor pointer part shouldn't be part of `MaybeUinit`, only the data it points o.

MarijnS95: Changed `u8` to `MaybeUninit<u8>` on all occurrences, no other changes necessary.

MarijnS95: One downside, as mentioned in IRC, is that various helper functions for `[MaybeUninit<T/u8>]` are still experimental: https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.slice_assume_init_ref

MarijnS95: And as discussed on IRC, removed these again until working with slices of `MaybeUninit` becomes better supported.

Ralith (approved):

PR #688, opened 2022-12-01
Update Vulkan-Headers to 1.3.238

MarijnS95:


ash/src/vk/definitions.rs
```diff
@@ -520,6 +525,11 @@ pub type PFN_vkDeviceMemoryReportCallbackEXT = Option<
         p_user_data: *mut c_void,
     ),
 >;
+#[allow(non_camel_case_types)]
+#[doc = "<https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/PFN_vkGetInstanceProcAddr.html>"]
+pub type PFN_vkGetInstanceProcAddr = Option<
+    unsafe extern "system" fn(instance: Instance, p_name: *const c_char) -> PFN_vkVoidFunction,
+>;
```
MarijnS95, line 532: Note to self: Khronos found this duplicate definition (we already have it in `features.rs` because of implicit type generation on the `vkGetInstanceProcAddr` `<command>`) to be problematic in C, and it is disallowing us to import it from Rust as well thanks to our glob reexports:

```
error[E0432]: unresolved import `ash::vk::PFN_vkGetInstanceProcAddr`
 --> examples/src/lib.rs:9:5
  |
9 |     vk::PFN_vkGetInstanceProcAddr,
  |     ^^^^-------------------------
  |     |   |
  |     |   help: a similar name exists in the module: `PFN_vkGetDeviceProcAddr`
  |     no `PFN_vkGetInstanceProcAddr` in `vk`
```

(Unsure why the compiler doesn't clearly complain about having found a duplicate, though)

MarijnS95: Fixed now!

ash/src/vk/extensions.rs
```diff
@@ -23945,25 +23990,32 @@ impl StructureType {
...
+        }
     }
-    pub const SPEC_VERSION: u32 = 0u32;
+    pub const SPEC_VERSION: u32 = 1u32;
 }
 #[derive(Clone)]
-pub struct ExtExtension460Fn {}
-unsafe impl Send for ExtExtension460Fn {}
-unsafe impl Sync for ExtExtension460Fn {}
-impl ExtExtension460Fn {
```
Ralith, line 23959: Would be nice if we could stop generating these non-bindings in the first place.

Ralith (approved):

ash/src/vk/extensions.rs
MarijnS95: #448, iirc.

Ralith (approved):

Issue #573, opened 2022-02-03
Writing validation layers in Rust.


cheako:
There are a few new strucks not included:

https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/blob/25d5884746a2de7b51a8ef3ec88e1cd8066460e8/include/vulkan/vk_layer.h#L112

It becomes a pattern of descending chains of pnext looking for a particular strict type.

Many tools used in implementing vulkan in rust would be handy.

I am working on an example layer, I expect it to segfault a lot.

cheako:
The magic:
```rust
#[repr(transparent)]
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct VkLayerFunction(pub(crate) vk::Flags);
ash::vk_bitflags_wrapped!(VkLayerFunction, vk::Flags);
impl VkLayerFunction {
    pub const VALUE: Self = Self(0b0);
}

#[repr(C)]
#[cfg_attr(feature = "debug", derive(Debug))]
#[derive(Copy, Clone)]
pub struct VkLayerInstanceCreateInfo {
    pub s_type: vk::StructureType,
    pub p_next: *const c_void,
    pub function: VkLayerFunction,
    pub p_layer_info: *const VkLayerInstanceLink,
}

#[repr(C)]
#[cfg_attr(feature = "debug", derive(Debug))]
#[derive(Copy, Clone)]
pub struct VkLayerInstanceLink {
    pub p_next: *const c_void,
    pub pfn_next_get_instance_proc_addr: vk::PFN_vkGetInstanceProcAddr,
    pub pfn_next_get_device_proc_addr: vk::PFN_vkGetDeviceProcAddr,
}

#[repr(C)]
#[cfg_attr(feature = "debug", derive(Debug))]
#[derive(Copy, Clone)]
pub struct VkLayerDeviceLink {
    pub p_next: *const c_void,
    pub p_instance: vk::PFN_vkGetInstanceProcAddr,
    pub p_device: vk::PFN_vkGetDeviceProcAddr,
}
```

https://gitlab.com/cheako/vk-layer-cache/-/snippets/2248646

cheako:
First viable product: https://gitlab.com/cheako/vk-layer-cache/-/blob/359e97aac7539298d21a06c09ee5604d53943979/src/lib.rs

i509VCB:
[...]

PR #694, opened 2022-12-20
Add `Handle::is_null()`

i509VCB:
This is related to a comment I made: https://github.com/ash-rs/ash/issues/573#issuecomment-1272222609

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added `VK_EXT_pipeline_properties` device extension (#622)
+- Added `Handle::is_null()` to allow checking if a handle is a `NULL` value (#694)
 - Update Vulkan-Headers to 1.3.246 (#697, #723)
 - Added `VK_KHR_performance_query` device extension (#726)
 - Added `VK_EXT_shader_object` device extension (#732)
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -45,8 +45,21 @@ pub(crate) unsafe fn ptr_chain_iter<T>(ptr: &mut T) -> impl Iterator<Item = *mut
         Some(old)
     })
 }
-pub trait Handle {
+pub trait Handle: Sized {
     const TYPE: ObjectType;
     fn as_raw(self) -> u64;
     fn from_raw(_: u64) -> Self;
+
+    /// Returns whether the handle is a `NULL` value.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use ash::vk::{Handle, Instance};
+    /// let instance = Instance::null();
+    /// assert!(instance.is_null());
+    /// ```
+    fn is_null(self) -> bool {
+        self.as_raw() == 0
+    }
 }


Changelog.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Added `Handle::is_null` to allow checking if a handle is a null handle (#694)
```
MarijnS95, line 11: Maybe better under `### Added`?

ash/src/vk.rs
```diff
@@ -49,4 +49,17 @@ pub trait Handle {
     const TYPE: ObjectType;
     fn as_raw(self) -> u64;
     fn from_raw(_: u64) -> Self;
+
+    /// Returns true if the handle is a null handle.
```
MarijnS95, line 53:
```suggestion
    /// Returns [`true`] if the handle is a null handle.
```

Or maybe just "Returns whether the handle is ..."

Changelog.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `Handle::is_null` to allow checking if a handle is a null handle (#694)
 - Update Vulkan-Headers to 1.3.238 (#688)
```
MarijnS95, line 17: I typically keeps these sorted somewhat chronologically, by PR ID ascending.

MarijnS95 (approved):

ash/src/vk.rs
```diff
@@ -49,4 +49,16 @@ pub trait Handle {
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use ash::vk::{Handle, Instance};
+    ///
+    /// let instance = Instance::null();
+    /// assert!(instance.is_null());
+    /// ```
+    fn is_null(&self) -> bool;
```
Ralith, line 63: Like the other methods, this should take `self` by value (see e.g. [`pointer::is_null`](https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.is_null).

Ralith (changes requested):

ash/src/vk.rs
MarijnS95: Good catch, I did not even notice that!

ash/src/vk.rs
```diff
@@ -49,4 +49,15 @@ pub trait Handle {
...
+    /// Returns whether the handle is a `NULL` value.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use ash::vk::{Handle, Instance};
+    /// let instance = Instance::null();
+    /// assert!(instance.is_null());
+    /// ```
+    fn is_null(self) -> bool;
```
Ralith, line 62: Maybe have a provided impl of `self.as_raw() == 0`?

Ralith (approved):

ash/src/vk.rs
MarijnS95: Thought about that as well, but ended up not doing that. Wouldn't hurt to have it though, I guess?

MarijnS95: Oh apparently this is only allowed if we add `Sized` to the trait bound of `Handle`:

```rust
error[E0277]: the size for values of type `Self` cannot be known at compilation time
  --> ash/src/vk.rs:62:16
   |
62 |     fn is_null(self) -> bool {
   |                ^^^^ doesn't have a size known at compile-time
   |
help: consider further restricting `Self`
   |
62 |     fn is_null(self) -> bool where Self: std::marker::Sized {
   |                              ++++++++++++++++++++++++++++++
help: function arguments must have a statically known size, borrowed types always have a known size
   |
62 |     fn is_null(&self) -> bool {
   |                +
```

Which is fine.

Ralith: The main benefit of a provided impl is that now we can drop the explicit impls from the macro.

MarijnS95: D'oh, I forgot to save.

Ralith (approved):

PR #695, opened 2022-12-21
Fix extension-provided `draw_indirect_count`

teoxoy:


MarijnS95 (approved):
Nice catch! It has been this way since the extension was introduced in 2020 via https://github.com/ash-rs/ash/pull/262 (CC @gabdube), and nobody noticed :) (likely because it was stabilized).

Can you document this in the readme?:

```diff
diff --git a/Changelog.md b/Changelog.md
index 944fec8..943e6f7 100644
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 
+### Fixed
+
+- extensions/khr/draw_indirect_count: use `cmd_draw_indirect_count_khr` instead of `cmd_draw_indexed_indirect_count_khr` for non-indexed draw call. (#695)
+
 ### Added
 
 - Update Vulkan-Headers to 1.3.238 (#688)
```

teoxoy:
Added a changelog entry.

> Nice catch! It has been this way since the extension was introduced in 2020 via https://github.com/ash-rs/ash/pull/262 (CC @gabdube), and nobody noticed :) (likely because it was stabilized).

Yeah, it seems so, we only noticed because we recently started using the extension as opposed to the one in core 1.2.
Crediting @pwouik (https://github.com/gfx-rs/wgpu/issues/3079#issuecomment-1361092358).

Changelog.md
```diff
@@ -21,6 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed experimental AMD extensions (#607)
 - Removed misnamed, deprecated `debug_utils_set_object_name()` and `debug_utils_set_object_tag()` entirely, use `set_debug_utils_object_name()` and `set_debug_utils_object_tag()` instead (#661)
 
+### Fixed
+
+- `VK_KHR_draw_indirect_count`: use `cmd_draw_indirect_count_khr` instead of `cmd_draw_indexed_indirect_count_khr` for non-indexed draw call (#695)
```
MarijnS95, line 26: Gah, I should start cleaning up our `<subsystem>:` prefix in the changelog, it's all over the place everywhere :)

teoxoy: Want me to change it to something else? I made it look a bit more consistent with the others.

MarijnS95: I think I'll fix `khr::RayTracingPipeline` separately, or feel free to replace that with `VK_KHR_ray_tracing_pipeline`. We don't have many **prefixes** like this, but most changelog entries are referring to an extension by its name rather than the file path or Rust module path anyway.

MarijnS95: In that sense removing the colon (`:`) and making this a complete sentence is fine by me as well (maybe even has my preference now).

teoxoy: Yeah,  `khr::RayTracingPipeline` is the odd one out of the bunch.

I think it's worth doing as a separate PR though.
I'd touch a bunch of changelog entries that will end up pointing to this unrelated PR.

MarijnS95: Feel free to submit one (and feel free to make this one a sentence as part of this PR), otherwise I'll try to fix them up with issuing the next release, if I don't forget.

gabdube:
Ouch, sorry about that

cwfitzgerald:
What's the status over here?

Changelog.md
MarijnS95: > In that sense removing the colon (`:`) and making this a complete sentence is fine by me as well (maybe even has my preference now).

@teoxoy what's your thought, stick to `` `VK_KHR_draw_indirect_count`: `` or turn this into a more naturally flowing sentence without that prefix?

teoxoy: Tbh, I think it's fine as it is. I'd imagine most people using ash would know what to look for and the ext name makes it easy to spot.

MarijnS95: Ack, I'll refactor these all at once later then.

PR #697, opened 2023-01-19
Update Vulkan-Headers to 1.3.244

### LINKED ISSUE #670: Deprecation notice of VkDeviceCreateInfo enabledLayerCount and ppEnabledLayerNames
(apologies for using the C structure names)

Recently had a user try to enable validation layers through the device. This does nothing, and makes users that are unaware of its uselessness think they are enabling validation.

Maybe it isn't possible, but it would be lovely if there was a 'deprecated' tag for this.

### LINKED ISSUE #698: FYI - we will be making use of 'api' attribute in Vulkan XML, you will need a small change
This has been part of the schema for a while, but we will start using it soon to allow combining the Vulkan and Vulkan SC XML in a single file. You will need to adapt your loading code to this new use.

The simplest way is to globally pre-filter the XML to remove all tags with explicit 'api' elements not matching the API being generated, as the Vulkan-Docs scripts already do. For this you can either use the scripts/stripAPI.py script that's provided, or do the equivalent

> import reg  
> keepAPI = 'vulkan'
> reg.stripNonmatchingAPIs(tree.getroot(), keepAPI, actuallyDelete=True)

after loading your ElementTree from the XML, where 'reg' is scripts/reg.py from the Vulkan-Docs repository. Although since you are probably doing the loading and parsing in Rust, you would have to figure out the equivalent of this Python command. The essence is to recursively traverse the XML tree removing any tags with explicit non-matching 'api' attributes at the point they're encountered, prior to any other XML parsing.

Let me know if there are any questions. There isn't a specific timescale for this, but probably will appear starting sometime in February.

MarijnS95:
Depends on #719

Closes #698, closes #670

Lots of churn to use the new `api` attribute, but all worth it in the end. Also notice the shiny new `deprecated` attribute :tada:


Ralith (approved):

Lokathor:
Parts of the PR update to 1.3.241 while the title says 1.3.242. I don't know which is correct but they should be made to agree before this merges.

MarijnS95:
Don't worry about it. I sometimes forget to update the value in the changelog and `Cargo.toml` but end up double-checking before merging (and definitely before releasing).

In this particular case there's still _lots_ to do on the generator side to make the new format work. What appear to be simple changes can only be passed through to the `vk-parse` structs, while we still _partly_ rely on the unmaintained `vkxml` crate. Desire has always been to migrate to `vk-parse` (and lately: rewrite the generator altogether) but that has yet to materialize.

Lokathor:
Well, I mean I've got a whole new parser that works on the latest XML, but since it's totally separate data types you'd (probably?) have to adjust your output stage a lot to use it.

MarijnS95:
It seems that everyone is writing a new parser, generator, or minimal Vulkan wrapper library nowadays, but that doesn't help us if it takes significant time to integrate with our generator/rewrite. Feel free to join in on https://github.com/ash-rs/ash/discussions/662 as we try to work out how (and when?) to merge Erupt's generator and approach to Vulkan with Ash.

ash/src/vk/bitflags.rs
```diff
@@ -252,12 +252,6 @@ impl FenceCreateFlags {
 }
 #[repr(transparent)]
 #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-#[doc = "<https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkSemaphoreCreateFlagBits.html>"]
-pub struct SemaphoreCreateFlags(pub(crate) Flags);
-vk_bitflags_wrapped!(SemaphoreCreateFlags, Flags);
-impl SemaphoreCreateFlags {}
```
MarijnS95, line 258: Note that these silly things are moving to `definitions.rs` because we're skipping the `SemaphoreCreateFlagBits` type now (not in a `<require>`) and then only later coming across `SemaphoreCreateFlags` which is handled by an _almost identical_ codepath (this is why we have the `const_cache` to deduplicate them): :vomiting_face:.

MarijnS95: This was caused by the addition of:

https://github.com/ash-rs/ash/blob/57de82d5f842b56a79060b9716ea98b9c0c3f2b8/generator/src/lib.rs#L2777-L2782

But now simply averted by using `.replace("FlagBits", "Flags")` :nauseated_face:

ash/src/vk/definitions.rs
```diff
@@ -1192,7 +1192,9 @@ pub struct DeviceCreateInfo<'a> {
     pub flags: DeviceCreateFlags,
     pub queue_create_info_count: u32,
     pub p_queue_create_infos: *const DeviceQueueCreateInfo<'a>,
+    #[deprecated = "functionality described by this member no longer operates"]
     pub enabled_layer_count: u32,
+    #[deprecated = "functionality described by this member no longer operates"]
```
charles-lunarg, line 1197: :+1:

ash/src/vk/extensions.rs
```diff
@@ -1487,8 +1487,6 @@ impl KhrSamplerMirrorClampToEdgeFn {
 impl SamplerAddressMode {
     #[doc = "Note that this defines what was previously a core enum, and so uses the 'value' attribute rather than 'offset', and does not have a suffix. This is a special case, and should not be repeated"]
     pub const MIRROR_CLAMP_TO_EDGE: Self = Self(4);
-    #[deprecated = "Alias introduced for consistency with extension suffixing rules"]
-    pub const MIRROR_CLAMP_TO_EDGE_KHR: Self = Self::MIRROR_CLAMP_TO_EDGE;
```
MarijnS95, line 1491: Looks like we forgot about this nonstandard deprecation message when yeeting out all other `deprecated` alias in the previous breaking Ash release. Merging this to `master` is fine, but I'll apply the following workaround when backporting this upgrade to `0.37`.

https://github.com/ash-rs/ash/commits/for-0.37-keep-deprecated-alias

MarijnS95:
@Ralith can you check out #719 and this PR? Upstream CI is waiting for us to be compatible with the new vk.xml format again.

Ralith (approved):

PR #703, opened 2023-02-02
examples: Upgrade `winit` to `0.28`

MarijnS95:


Ralith (approved):

MarijnS95:
@Ralith this requires an MSRV bump (for the examples _but also `ash-window`_) - I've yet to clear out the CI to make that work.

PR #705, opened 2023-02-05
fix `clippy::uninlined_format_args`

### LINKED ISSUE #704: Fix or opt out of `clippy::uninlined_format_args`?
https://rust-lang.github.io/rust-clippy/master/#uninlined_format_args

Clippy now warns on this by default in Rust 1.66, while it was only added in [Rust 1.58](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html) 48 weeks prior. I've been keeping it out of my own projects to not bump MSRV unnecessarily and to keep peers happy.

We'd be fine doing this though (in `ash/build.rs`!, the generator and examples) as our MSRV is already 1.59, but the question is whether we should. @Ralith how do you feel about the new format?

filnet:
Closes #704

Note that there is an unrelated `clippy::seek_to_start_instead_of_rewind` fix.

~And one instance of `clippy::uninlined_format_args` in generator.rs cannot be fixed and must be allowed.~

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -587,7 +587,7 @@ fn convert_c_literal(lit: Literal) -> Literal {
         // If expression rewriting succeeds this should parse into a single literal
         match (stream.next(), stream.next()) {
             (Some(TokenTree::Literal(l)), None) => l,
-            x => panic!("Stream must contain a single literal, not {:?}", x),
+            x => panic!("Stream must contain a single literal, not {x:?}"),
         }
     } else {
         lit
@@ -627,7 +627,7 @@ fn convert_c_expression(c_expr: &str, identifier_renames: &BTreeMap<String, Iden
     }
     let c_expr = c_expr
         .parse()
-        .unwrap_or_else(|_| panic!("Failed to parse `{}` as Rust", c_expr));
+        .unwrap_or_else(|_| panic!("Failed to parse `{c_expr}` as Rust"));
     rewrite_token_stream(c_expr, identifier_renames)
 }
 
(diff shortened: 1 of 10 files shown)


generator/src/lib.rs
```diff
@@ -2379,6 +2379,11 @@ pub fn extract_native_types(registry: &vk_parse::Registry) -> (Vec<(String, Stri
         match ty.category.as_deref() {
             Some("include") => {
                 // `category="include"` lacking an `#include` directive are generally "irrelevant" system headers.
+                // using inlined format args in panic messages is not yet supported and will generate this warning:
+                // warning: panic message contains an unused formatting placeholder
+                // note: this message is not used as a format string when given without arguments, but will be in Rust 2021
+                // note: `#[warn(non_fmt_panics)]` on by default
+                #[allow(clippy::uninlined_format_args)]
```
Ralith, line 2386: Could we fix this by updating the edition to 2021?

filnet: Done.

generator/src/lib.rs
```diff
@@ -1772,7 +1768,7 @@ pub fn derive_setters(
                         #[inline]
                         pub fn #param_ident_short(mut self, #param_ident_short: &'a #mutable #slice_param_ty_tokens) -> Self {
                             #set_size_stmt
-                            self.#param_ident = #param_ident_short#ptr;
+                            self.#param_ident = #param_ident_short #ptr;
```
filnet, line 1771: See https://doc.rust-lang.org/edition-guide/rust-2021/reserving-syntax.html

Ralith (approved):
Thanks!

filnet:
`clippy::uninlined_format_args` is going to be downgraded to pedantic : https://github.com/rust-lang/rust/blob/stable/RELEASES.md#version-1671-2023-02-09

MarijnS95 (approved):

PR #706, opened 2023-02-15
impl `From` for `ClearValue` union possibilities

lilly-lizard:
Thought the way to declare `ClearValue` instances was kinda annoying so I wrote `From` implementations for it.

### DIFF
--- a/ash/src/vk/prelude.rs
+++ b/ash/src/vk/prelude.rs
@@ -53,6 +53,20 @@ impl From<vk::Extent2D> for vk::Rect2D {
     }
 }
 
+impl From<vk::ClearColorValue> for vk::ClearValue {
+    fn from(value: vk::ClearColorValue) -> Self {
+        Self { color: value }
+    }
+}
+
+impl From<vk::ClearDepthStencilValue> for vk::ClearValue {
+    fn from(value: vk::ClearDepthStencilValue) -> Self {
+        Self {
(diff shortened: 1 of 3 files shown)


Ralith:
Seems harmless, but the impact on callers is pretty trivial, and I worry that it might oblige us to do the same for any other unions that come up.

lilly-lizard:
yeah good point. looks like all the other unions are for extensions right now, so their use would be negligible compared to clear values. but it does seem like the sort of thing that should be automated with the generator...

idk tbh, but I won't mind you determine this isn't compatible with the design goals of the project.

lilly-lizard:
took a look at the union generator and it wasn't too hard to write up code to automate this for all unions. but it fells pretty weird to do this for something like
```rust
pub union DeviceOrHostAddressKHR {
    pub device_address: DeviceAddress,
    pub host_address: *mut c_void,
}
```
which would result in a `From` for a `c_void` pointer...

imo `ClearValues` are a special case compared to the others.

Ralith:
I guess it feels sort of like second-guessing the language design for unions. I don't feel super strongly though, maybe @MarijnS95 has an opinion.

MarijnS95:
Might be me, but I find the changes to the examples less explicit/readable, more ugly.

lilly-lizard:
@MarijnS95  fair enough!

PR #708, opened 2023-02-21
generator: Convert `#define` generation from `vkxml` to `vk-parse`

MarijnS95:
For upcoming `vk.xml` features (the new `api` attribute) some of our codegen has to be converted to work on `vk-parse` types to make this ergonomic (and there's a longstanding plan of factoring out `vkxml` regardless).  Start with converting the `#define` code and showcasing that it does not affect the output (beyond removing the unneeded edgecase for `VK_HEADER_VERSION` resulting in a doc link).


Ralith (approved):

MarijnS95:
> Start with converting the `#define` code and showcasing that it does not affect the output (beyond removing the unneeded edgecase for `VK_HEADER_VERSION` resulting in a doc link).

Unfortunately this is not entirely true anymore. The other big requirement for conversion is `struct`s, and `vkmxl` (the conversion in `vk-parse`, really) provides _all_ the necessary boilerplate that I do not feel like whipping up in Ash just now, with a rewrite in mind.

PR #712, opened 2023-02-28
utils: add getter for the aligned size

HugoPeters1024:
I was following the famous tutorials from Sascha Willems. At some point he uses a utility function called `alignedSize` when building the shader binding table for a raytracing pipeline:

```cpp
uint32_t alignedSize(uint32_t value, uint32_t alignment)
{
        return (value + alignment - 1) & ~(alignment - 1);
}
```

 I was hoping to find that same function in `ash::utils` and I did find the `Align` struct. From what I understand its `elem_size` field yields the same result. (as you probably know `& ~(alignment-1) === % alignment` if `alignment % 2 == 0`)

### DIFF
--- a/ash/src/util.rs
+++ b/ash/src/util.rs
@@ -59,6 +59,10 @@ impl<T> Align<T> {
         }
     }
 
+    pub fn aligned_size(&self) -> vk::DeviceSize {
+        self.elem_size
+    }
+
     pub fn iter_mut(&mut self) -> AlignIter<T> {
         AlignIter {
             current: 0,


MarijnS95:
Not sure about this: it is typically a function everyone implements themselves as it's useful so much more often outside the context of Vulkan/Ash as well. In what context are you constructing an `Align` with valid pointer where you need to know the aligned element size, when extracting SBT handles? Seems to me that you'd like to have `alignedSize()` without the `unsafe` pointer construction?

@Ralith what are your thoughts on this drive-by functionality of `struct Align`?

HugoPeters1024:
Yeah I suppose doing it directly makes more sense

Ralith:
Frankly I'd like to get rid of `Align` entirely; it's of very limited usefulness.

filnet:
The ash mission statement is "A very lightweight wrapper around Vulkan".
It is not the first time that a QoL or utility additions are not accepted because they are outside of this strict scope.

HugoPeters1024:
There appears to be a consensus about not providing this functionality within ash, closing

PR #713, opened 2023-03-02
rewrite: Absolute basics 

Friz64:
This PR intends to lay a groundwork for the rewrite.

Here are my changes

- Added an analysis crate, so that in the future code other than ash's generator can work with the API registry.
- Other small miscellaneous changes.


Friz64:
Responding to [early feedback](https://github.com/ash-rs/ash/discussions/662#discussioncomment-5176823) about this PR, I've changed it to keep the old generator in the code base as a reference point and I've changed the base branch to `master` so that we have a working version of the current generator to build off of. Now we even might not even need to have a `rewrite` branch.

Jerrody:
@Friz64 
Very cool!

.github/workflows/ci.yml
```diff
@@ -27,8 +27,9 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: check
-          args: -p ash -p ash-window --all-features
+          args: -p ash -p ash-rewrite -p ash-window --all-features
 
+  # TODO: add a similiar for the rewrite once that generates code
```
MarijnS95, line 32: a similar... job?

MarijnS95 (approved):
Nice and simple, this should allow us to more easily compare the new generated version to the old one. Which of course only makes sense if the new one looks similar to the old one... and that is exactly the feedback I have heard/received over time: `ash` works well for many users as a low-level no-frills and they really wouldn't want to deal with (much) more API breakage nor see any need for it. In other words, and as said before I really don't intend to rewrite/redesign `ash`, only the generator needs to be rewritten to deal with increasing complexity in `vk.xml` resulting in an unmaintainable hunk of code. Changes to `ash` itself should mostly be incremental, such as perhaps autogenerating the hand-written helpers, including more/better documentation, properly guarding experimental APIs behind `cfg` and so forth.

For this PR that mainly comes down to leaving `README.md` intact, _especially_ if we merge this to `master` instead of instantiating a special `rewrite` branch.

And if folks don't align with my view here, let's discuss. Incremental changes/improvements to `ash` is still rather broad, and I am open to breaking changes.

.github/workflows/ci.yml
Friz64: Typo, fixed.

Friz64:
Okay! Now it probably makes more sense to directly work from `master`, so: Reverted the `README.md` change.

Friz64:
Looks like CI broke because `raw-window-handle` 0.5.1 got released with https://github.com/rust-windowing/raw-window-handle/pull/107 which bumps its MSRV to 1.64.

MarijnS95:
Ugh, I'm also running into that MSRV break on the `ndk` crate. Don't feel like bumping the MSRV for a patch dependency update (that we "could" opt out of) but `raw-window-handle 0.5.1` provides some new features and drops `cty` in favour of the stabilized `ffi` types.

MarijnS95:
@Friz64 I'm tackling the MSRV bump in https://github.com/ash-rs/ash/pull/716 for just `ash-window`. In fact I think we might bump `ash` back down to 1.59 as I only bumped it to `1.60` for `winit 0.28` initially which isn't used in `ash`'s crate graph anyway (and 1.59 is still enough for #715 :partying_face:).

Friz64:
FYI I rebased this.

MarijnS95:
Let's get it in then and get this show on the road!

PR #715, opened 2023-03-09
Replace `const fn name()` with associated `NAME` constants

MarijnS95:
Depends on #716

`CStr::from_bytes_with_nul_unchecked` is `const`-stable since Rust 1.59 which is already required for `ash` so it is high time to finally turn these inlined `name()` functions into associated constants (which is a breaking change in itself that cannot be backported).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Replaced builders with lifetimes/setters directly on Vulkan structs (#602)
 - Inlined struct setters (#602)
 - Bumped MSRV from 1.59 to 1.60 (#709)
+- Replaced `const fn name()` with associated `NAME` constants (#715)
 
 ### Removed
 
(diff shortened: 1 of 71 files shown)


Ralith (approved):
:tada:

Any compiletime impact? Not sure which way I'd expect this to go.

MarijnS95:
e288193 (this PR):

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      6.105 s ±  0.057 s    [User: 6.905 s, System: 0.529 s]
  Range (min … max):    6.046 s …  6.228 s    10 runs
```

197bc30 (`HEAD~`, i.e. #716):

```console
hyperfine -p 'cargo clean -p ash' 'cargo b -p ash'
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      6.126 s ±  0.107 s    [User: 6.916 s, System: 0.517 s]
  Range (min … max):    5.994 s …  6.373 s    10 runs
```

That's identical.

PR #709, opened 2023-02-22
Bump MSRV to 1.60 to accomodate for `winit 0.28`/`objc` requirement

MarijnS95:
CC #703

These dependencies fail to resolve on older Rust/cargo versions before even complaining about them being a too-high MSRV.

PR #716, opened 2023-03-09
ash-window: Bump MSRV to 1.64 to match `raw-window-handle 0.5.1`

MarijnS95:
`raw-window-handle 0.5.1` bumped from 1.60 to 1.64 in a semver-compatible release, failing our CI infrastructure overnight.

~Keep the `ash` version at `1.60` for now.~ Lower `ash` to 1.59 because the original bump in #709 was only to match `winit 0.28`/`objc`, both of which are not used by the `ash` crate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
           args: --workspace --all-targets --all-features
 
   check_msrv:
-    name: Check MSRV (1.60.0)
+    name: Check ash MSRV (1.60.0)
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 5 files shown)


MarijnS95:
> Lower `ash` to 1.59 because the original bump in #709 was only to match `winit 0.28`/`objc`, both of which are not used by the `ash` crate.

Seems like I did not document it, but even if we use `-p ash` to only compile Ash on 1.59, the compiler first tries to understand the entire crate graph and somehow fails on `objc` yet again, while it is fine on 1.60 (https://github.com/ash-rs/ash/actions/runs/4374927457) despite having MSRV-1.64 elsewhere in the not-actively-compiled dependency graph.

We could solve this by either `sed`'ing those crates out of the workspace, or dropping the second commit again and sticking to `1.60` for Ash (wasn't released yet). Thoughts?

Friz64:
By saying "`sed`'ing those crates out of the workspace", are you suggesting to remove their reference in the Cargo.toml workspace members section? If I'm understanding this correctly I'd rather opt to remove the second commit and take the hit of one Rust version.

MarijnS95:
> By saying "`sed`'ing those crates out of the workspace", are you suggesting to remove their reference in the Cargo.toml workspace members section?

Yes.

> If I'm understanding this correctly I'd rather opt to remove the second commit and take the hit of one Rust version.

Sure, let's keep it at `1.60`, otherwise users cloning this repo on `1.59` can't repro a successful build either.

Ralith (approved):

PR #717, opened 2023-03-10
Rename some function arguments to match the vulkan spec

lilly-lizard:
Found it annoying that some functions have arguments called `create_info` when the struct type isn't actually a `create_info` e.g.
```rust
    pub unsafe fn cmd_begin_render_pass(
        &self,
        create_info: &vk::RenderPassBeginInfo,
        contents: vk::SubpassContents,
    )
```

So I looked through all the `create_info` function arguments and renamed the ones which didn't make sense to match whatever the vulkan spec calls them e.g. `memory_reqirements`

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -465,34 +465,42 @@ impl Device {
     #[inline]
     pub unsafe fn get_device_buffer_memory_requirements(
         &self,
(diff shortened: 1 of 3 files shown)


MarijnS95 (approved):
Thanks! You are right, that is a really poor copy-paste (was likely me).

PR #720, opened 2023-03-13
CI: Replace deprecated `actions-rs` with simple `run: cargo ...`

MarijnS95:
There's no reason to use these steps anymore: besides being old, unmaintained, and spitting out NodeJS deprecation warnings, GitHub's `runner-images` come preloaded with all Rust tools and components we need, and the syntax to run commands is more efficient, much shorter and more apprehensible by simply matching what we'd use on our own command-line, too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,15 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


Ralith (approved):
Man, I have a *ton* of CI scripts to update...

MarijnS95:
At least it's lots of removals, for simplicity :)

PR #724, opened 2023-03-27
Generate generic builder setters for fields taking an `objecttype`

MarijnS95:
We already do this for hand-written extension functions but can now also implement it for setters since `vk_parse` fields are available within the builder generator code: when a field refers to another field for setting its `objecttype`, that `VkObjectType` field setter is omitted and instead assigned when the object is set, based on a type generic that implements the `Handle` trait instead of an untyped `u64`.

---

Note that most of this diff is merely code motion: some necessary cleanups such as replacing tuples with infinitely-more-self-descriptive structs.


Ralith (approved):
Nice!

PR #726, opened 2023-03-28
extensions/khr: Add VK_KHR_performance_query

MarijnS95:
CC @wolfiestyle


wolfiestyle:
Couldn't test locally. patch doesn't work applied to 0.37.2, and trying to use HEAD w/patch breaks other deps.
The error:
```
$ cargo check
    Checking ash v0.37.2+1.3.238 (/home/wolfie/progs/rust/vktest/ash/ash)
error[E0599]: no associated item named `NAME` found for struct `KhrPerformanceQueryFn` in the current scope
    --> ash/ash/src/extensions/khr/performance_query.rs:107:64
     |
107  |     pub const NAME: &'static CStr = vk::KhrPerformanceQueryFn::NAME;
     |                                                                ^^^^ associated item not found in `KhrPerformanceQueryFn`
     |
    ::: ash/ash/src/vk/extensions.rs:7017:1
     |
7017 | pub struct KhrPerformanceQueryFn {
     | -------------------------------- associated item `NAME` not found for this struct
```

MarijnS95:
@wolfiestyle I'll address that after merging and backporting this to the `0.37-stable` branch, and for the time being you can replace this associated constant with:

```rust
    #[inline]
    pub const fn name() -> &'static CStr {
        vk::KhrPerformanceQueryFn::name()
    }
```

Ralith (approved):

ash/src/extensions/khr/performance_query.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    pub const NAME: &'static CStr = vk::KhrPerformanceQueryFn::NAME;
+
+    #[inline]
+    pub fn fp(&self) -> &vk::KhrPerformanceQueryFn {
+        &self.fp
+    }
+
+    #[inline]
+    pub fn device(&self) -> vk::Instance {
```
MarijnS95, line 115: Lol, same copy-paste bug as `VK_KHR_device_group_creation`, pointed out in #727.

PR #730, opened 2023-03-30
prelude: Add loop limit for VK_INCOMPLETE

idanraiter:
Bad validation layers or drivers can end up always giving VK_INCOMPLETE. Stop trying after 100 tries instead of looping forever.

### DIFF
--- a/ash/src/prelude.rs
+++ b/ash/src/prelude.rs
@@ -26,7 +26,7 @@ impl vk::Result {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ash/src/prelude.rs
```diff
@@ -26,6 +26,9 @@ impl vk::Result {
     }
 }
 
+/// The amount of times we will attempt to call a function that returns [`vk::Result::INCOMPLETE`].
```
filnet, line 29: nit: should the `The number of times [...]`

PR #733, opened 2023-04-03
generator: Borrow all `command` names and disentangle "aliases"

MarijnS95:
Turns out we were doing the wrong thing for the right reason: the `aliases` here aren't `vk.xml` aliases: they are renames.  When generating function pointers for extensions, a list of command _definitions_ is collected, which can only ever be "root" `command`s.  Extensions typically reference stabilized `command`s under an alias with the vendor tag suffixed, which the `Fn` struct field name is renamed to using this `aliases` - now replaced with `rename_commands` - list, while generating the rest of the "function pointer" command bits using the "root" `command` (as this mostly pertains the parameters and return type).  With that explanation it becomes clear why `generate_extension_commands()` was creating an "alias" mapping from stabilized name to vendor-suffixed extension name, and calls `generate_function_pointers()` with that mapping - and a list of stabilized/root `command`s - rather than passing `cmd_aliases` directly.  (This `cmd_aliases` list exists because the rename always happens in the root `<commands>` element: extensions then `<require>` the aliased rather than the stabilized name, so the base for this alias is found first to look up the base command, and then stored in `rename_commands` to rename it back to the aliased name).

With improved clarity we can now also borrow the name strings rather than cloning them in many places.

---

In the end this was largely unnecessary, but it took me a while (too long) to understand why we were deduplicating (`fn_cache`) the `PFN_` types on _extension command name_ rather than the _base aliased-to name_, because the alias was inverted with respect to `vk.xml`.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -893,7 +893,7 @@ pub type CommandMap<'a> = HashMap<vkxml::Identifier, &'a vk_parse::CommandDefini
 fn generate_function_pointers<'a>(
     ident: Ident,
     commands: &[&'a vk_parse::CommandDefinition],
-    aliases: &HashMap<String, String>,
+    rename_commands: &HashMap<&'a str, &'a str>,
     fn_cache: &mut HashSet<&'a str>,
 ) -> TokenStream {
     // Commands can have duplicates inside them because they are declared per features. But we only
@@ -903,10 +903,10 @@ fn generate_function_pointers<'a>(
         .unique_by(|cmd| cmd.proto.name.as_str())
         .collect::<Vec<_>>();
 
-    struct Command {
+    struct Command<'a> {
         type_needs_defining: bool,
         type_name: Ident,
-        function_name_c: String,
+        function_name_c: &'a str,
         function_name_rust: Ident,
         parameters: TokenStream,
         parameters_unused: TokenStream,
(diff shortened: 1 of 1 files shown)


generator/src/lib.rs
```diff
@@ -919,11 +919,14 @@ fn generate_function_pointers<'a>(
             let name = &cmd.proto.name;
             let type_name = format_ident!("PFN_{}", name);
 
-            let function_name_c = if let Some(alias_name) = aliases.get(name) {
-                alias_name.to_string()
+            // We might need to generate a function pointer for an extension, where we are given the original
+            // `cmd` and a rename back to the extension alias (typically with vendor suffix) in `rename_commands`:
```
filnet, line 923: is the trailing colon meant ?

generator/src/lib.rs
```diff
@@ -919,11 +919,14 @@ fn generate_function_pointers<'a>(
...
+            let function_name_c = if let Some(&renamed) = rename_commands.get(name.as_str()) {
```
filnet, line 924: You could use `Option::map_or()` on the `get(...)` return value and get rid of the `if let`.

Ralith (approved):

generator/src/lib.rs
MarijnS95: An `unwrap_or()` with `cloned()` ;)

generator/src/lib.rs
MarijnS95: I typically use trailing colons in comments to indicate that it relates to the next line, specifically. See other instances of `//.*:$` within the codebase.

Issue #731, opened 2023-03-31
How to deal with large ash::Device type in libraries


i509VCB:
I opened https://github.com/Traverse-Research/gpu-allocator/issues/159 a week ago and while thinking how to resolve that issue on a larger scale across the ecosystem where ash is used, the ash::Device type is quite large. In Smithay's WIP vulkan renderer by using gpu_allocator and having the renderer own a ash::Device I essentially double the size of my VulkanRenderer type. If I wanted to bring in a descriptor pool allocator it would be a third ash::Device.

Clearly tripling the size of the type (without any Boxes) is quite bad. I believe ash is probably the best spot to find a solution so that the ecosystem can adopt it.

## Possible solutions

Most libraries and apps use a fraction of the vulkan commands. It would make sense then to reduce the size of types by only loading what the app uses. A few solutions to the problem come to mind:

1. Allow generating a type like ash::Device with the commands the app/library needs. This would help with apps that need a fraction of the API but code generation for this would be needed and will be quite complicated.
2. Give every command a function loader type like the extensions. This would mean a safer wrapper can be provided for every command. This is similar to how extension function loading is done in ash. However the current method of storing a vk::Instance or vk::Device will cost 8 bytes per instance on 64-bit systems. It's far less than 100s of commands going unused but it can add up quickly.
3. Load the raw function pointers by hand and use the function pointers. This is what C libraries do and it works. But I'd really like something better.
4. Change libraries to not own a device and instead borrow. This can cause some problems with borrow checking.
5. Just Box/Arc it and call it a day. This works but it does waste heap memory. Plus libraries need to coordinate who destroys the device.

Ralith:
I pass `&Device` around. I've never had any issue with borrow checking as a result. Don't try to capture the reference.

> Just Box/Arc it and call it a day. This works but it does waste heap memory

It does not waste memory.

i509VCB:
> Arc does not waste memory.

If the library owns a ash::Device and a other owns a Arc<ash::Device> you'll have to clone the device and therefore double the memory use.

MarijnS95:
> If the library owns a ash::Device and another owns a `Arc<ash::Device>` you'll have to clone the device and therefore double the memory use.

Sounds like the library in question (`gpu-allocator`) needs to stop owning the handle and function pointers, perhaps via an `Arc` or whatever "standard" we decide to settle on in this issue.

---

[...]

PR #734, opened 2023-04-03
Separate function-pointer structs between instance and device

### LINKED ISSUE #727: Some _device_ extensions are loading (and failing) _instance_ functions via `get_device_proc_addr()`
Found a couple extensions that are erroneously loading _all_ functions through `get_device_proc_addr()` despite containing some instance-level functions (i.e. not taking a device or device child as first argument, see also `fn function_type()` in the generator for fishing this out in a crude way), which [cannot be loaded](https://vulkan.lunarg.com/doc/view/1.3.211.0/linux/LoaderInterfaceArchitecture.html#user-content-device-functions). Concretely found these extensions (by looking for explicit `PhysicalDevice` argument, and later [validated using the generator](https://github.com/ash-rs/ash/issues/727#issuecomment-1494969910)):

- `VK_KHR_device_group`: #631
- `VK_EXT_full_screen_exclusive`: #399
- `VK_KHR_swapchain` (new `get_physical_device_present_rectangles`): #629

#### Quick solution

Unfortunately the fix (a pattern which we apply more rigorously in other extensions) implies a slower path for device-level functions which now imply an extra dispatch table per device, "injected" by the ICD. Worse, it's a breaking change so it might be about time to start thinking about `ash 0.38`.

#### Long-term correctness

As for the fix itself, I've said it before and I'll use this issue to track it: I'd rather have the generator distinguish instance-level and device-level _extension_ functions, either in the loader or via separate `*Fn` structs entirely (anything to force us to load both in a separate way). @oddhack what's the desired way to figure this out if a function operates on device (or childs thereof)? I'm not confident this hardcoded list of `"VkDevice" | "VkCommandBuffer" | "VkQueue"` for the first-parameter-type will hold, does it need an extra attribute in `vk.xml`?

https://github.com/ash-rs/ash/blob/a9fbc7147bd2eef124d8b2153e94bbf313885420/generator/src/lib.rs#L561-L566

### Random weirdness
`VK_KHR_device_group_creation`, added in #630, has a function named `fn device()` that returns `vk::Instance` :sweat_smile:

MarijnS95:
~Depends on #733~
Fixes #727

Not only is it more efficient to load specialized device functions via `get_device_proc_addr()` instead of `get_instance_proc_addr()` (saves a device-based jump in the ICD), loading instance functions via `get_device_proc_addr()` (which was the case in `VK_KHR_swapchain`, `VK_KHR_device_group` and `VK_EXT_full_screen_exclusive`) always results in a `NULL` causing their instance functions panic no matter what.

Low-level `*Fn` structs are separated out between instance and device pointers for every extension that contains both, forcing users (typically high-level extension writers) to explicitly account for this.


ash/src/extensions/ext/calibrated_timestamps.rs
```diff
@@ -55,15 +41,54 @@ impl CalibratedTimestamps {
         .result_with_success((timestamps, max_deviation))
     }
 
-    pub const NAME: &'static CStr = vk::ExtCalibratedTimestampsFn::NAME;
+    pub const NAME: &'static CStr = vk::ExtCalibratedTimestampsDeviceFn::NAME;
```
MarijnS95, line 44: I've been contemplating keeping the original non-suffixed struct and leaving the two associated constants there...

Ralith:
Having large numbers of types with one of two suffices feels a bit like we're reimplementing the module system. A ZST purely for the sake of the name constant also feels a bit weird.

Possible alternative approaches:
- Three types per extension, with FP tables canonically exposed via associated types on a single extension ZST, alongside an associated `const NAME`
- The above, except the root extension type has fields `instance: InstanceFn, device: Option<DeviceFn>`
- One module per extension, containing `const NAME`, `struct Device`, and/or `struct Instance`
- Two structs per extension, each having the same name but inside `ash::extensions::device::...` and `ash::extensions::instance::...` respectively (and similarly for the raw layer)

Ralith:
I also wonder if we should scrap the panicing stub functions and fail loaders if any function doesn't load, since that's already bitten us pretty badly by making the errors subtle (and confused users/contributors).

MarijnS95:
> Having large numbers of types with one of two suffices feels a bit like we're reimplementing the module system. A ZST purely for the sake of the name constant also feels a bit weird.

We already have modules but only for the sake of `ash` readability it seems: all these `struct`s are reexported from the module hierarchy. If we go ahead with a module system we'd have to drop the reexports and make the modules public (or only for these few specific extensions? Then I'm more in favour of associated types on a ZST to at least look similar to other extensions).

> Possible alternative approaches:
> 
> * Three types per extension, with FP tables canonically exposed via associated types on a single extension ZST, alongside an associated `const NAME`

I've been thinking the same thing for the suggested ZST above, but then would wonder if we could keep the struct name "private" in favour of always referring to it as `ExtensionName::Device`?

Unfortunately these so-called `inherent associated types` are still unstable:

```rust
pub struct CalibratedTimestamps;

impl CalibratedTimestamps {
    // inherent associated types are unstable
    // see issue #8995 <https://github.com/rust-lang/rust/issues/8995> for more information
    pub type Device = CalibratedTimestampsDevice;
    pub type Instance = Instance;
}
```

> * The above, except the root extension type has fields `instance: InstanceFn, device: Option<DeviceFn>`

The downside here is that it forces reloading instance functions, if you already have them or only care about device functions. OTOH we can find a middle-ground here a chain of constructors either loads `instance`, or takes/moves an existing value for it so that you can _upgrade_ the struct with "device" support.

Additionally we could keep a mode or constructor where the device functions are loaded via `get_instance_proc_addr` (and make it all available via a single type) but nobody complained about the current setup requiring `Device` for device-specific extensions so this isn't useful and only adds lots complexity/code.

> * One module per extension, containing `const NAME`, `struct Device`, and/or `struct Instance`

That'd be nice and generic, if a bit harder to manage for the folks that import structs rather than modules, but nothing that can't be fixed. Perhaps worth investigating given that associated types on ZSTs are the "clearer" (?) variant of this but not yet stabilized.

In this case I'd name them `DeviceFn` and `InstanceFn` or something similar, to disambiguate between a real Vulkan device/instance? (Not all extensions hold the device/instance handle after all)

> * Two structs per extension, each having the same name but inside `ash::extensions::device::...` and `ash::extensions::instance::...` respectively (and similarly for the raw layer)

Doing the `device`/`instance` split on the module level rather than on the `struct` level? That may be interesting but really pulls the codebase apart and makes it again more prone to import the same item names in scope (solving it either with `as DeviceXXXExtension` or qualifying in scope with `device::XXXExtension`).

MarijnS95:
> I also wonder if we should scrap the panicing stub functions and fail loaders if any function doesn't load, since that's already bitten us pretty badly by making the errors subtle (and confused users/contributors). I guess that's tricky for versioned stuff.

We should definitely have a mode where this is caught up-front, yes. Just thinking about how we should represent `ash::Device` containing four function pointer groups now, or start with the extensions first (which are typically an all-or-nothing, except the broken `get_physical_device_present_rectangles()` in this PR which requires Vulkan 1.1 in addition to the `Swapchain` extension. And there are probably more such combinations out in the wild now).

MarijnS95:
@Ralith WDYT? This is one of the few PRs remaining before cutting `0.38`, I think :)

Ralith:
This needs a proper think, will try to get back to you on it later today. Thanks for driving everything forwards!

MarijnS95:
@Ralith lovely, let's surely think this through properly before merging and releasing any hasty implementation (though we can neatly use this PR to experiment and reflect, it was opened specifically to spark such a discussion). We can also defer to #727 to discuss this.

Ralith:
> Unfortunately these so-called inherent associated types are still unstable:

Aw.

> we can find a middle-ground here a chain of constructors either loads instance, or takes/moves an existing value for it so that you can upgrade the struct with "device" support.

That's what I was thinking. Something like:

```rust
struct SomeExt {
    instance_handle: vk::Instance,
    instance_fn: InstanceFn,
    device: Option<(vk::Device, DeviceFn)>,
}

impl SomeExt {
    fn load(instance: &Instance, device: &Device) -> Self { ... }
    fn load_instance(instance: &Instance) -> Self { ... }
    fn load_device(&mut self, device: &Device) { ... }
}
```

This has the drawback of forcing dynamic checks on every device fn, of course, so I'm not sold on it.

> If we go ahead with a module system we'd have to drop the reexports and make the modules public

That was my thinking. We'd set a convention of always importing at most the modules, never their contents, similar to idiomatic use of e.g. `std::{io, mpsc, thread, ...}`, and do so consistently for all extensions. Having some extensions be `FooFn` and some be `foo::{InstanceFn, DeviceFn}` would be too confusing, and being clear about the stage an extension is usable at seems nice even if it happens to only have one.

> [module per extension]'d be nice and generic if a bit harder to manage for the folks that import structs rather than modules

What's the difference? An item's an item, as far as scoping is concerned.

> In this case I'd name them DeviceFn and InstanceFn or something similar, to disambiguate between a real Vulkan device/instance? (Not all extensions hold the device/instance handle after all)

Makes sense to me.

> given that associated types on ZSTs are the "clearer" (?) variant

It's not obvious to me that ZSTs are clearer in this application. Generally we use types to describe values, and in that case there's no values, so why have a type at all?

> [splitting on device/instance first] may be interesting but really pulls the codebase apart

Yeah, I think that's probably the wrong way to go. It makes it difficult to survey the contents of an extension, and there's no single graceful place to hang the `NAME`.

MarijnS95:
> > [module per extension]'d be nice and generic if a bit harder to manage for the folks that import structs rather than modules
> 
> What's the difference? An item's an item, as far as scoping is concerned.

It seems to be more common to import _items_ in scope rather than strategically importing modules (and this may also be driven by how `rust-analyzer` helps you do that, though it might also help you import modules?). So for example, a user may start out with:

```rust
use ash::extensions::khr::swapchain::{NAME, DeviceFn};
```

And now have a meaningless `DeviceFn` item in scope throughout their source. This gets worse when using a second or more extensions (But is equivalently easy to solve by importing up until `swapchain;`.

> > given that associated types on ZSTs are the "clearer" (?) variant
> 
> It's not obvious to me that ZSTs are clearer in this application. Generally we use types to describe values, and in that case there's no values, so why have a type at all?

Given the above, inherent associated types cannot be imported in scope yet (e.g. `use ...::Swapchain::DeviceFn`), so the above problem is forcefully solved.
(However, one can import/reexport enum _variants_ so this might be enabled before the feature is stabilized?)

MarijnS95:
Re. "how `std` does it", the new model feels like we're having the following layout:

```rust
use std::collections::hash_map::Collection;
use std::collections::hash_set::Collection;
use std::collections::vec::Collection;
```

(Disregard the double `collections` module)

because all the modules have identically-named items.

Ralith:
Contrast `std::{format, io, thread}::Result`, the *many* instances of `...::{Iter, IterMut}`, `std::{io, iter}::Take`, etc. It's a well established pattern. Even rust-analyzer will only import `DeviceFn` if the user starts out by writing unqualified `DeviceFn`; starting with the extension name, as both exploring users and those referring to examples will tend to, will behave sensibly.

> `use std::collections::hash_map::Collection;`

This doesn't exist because there's necessarily a sensible "main" type for any collection.

>  (However, one can import/reexport enum variants so this might be enabled before the feature is stabilized?)

Yeah, there's already some pressure for this today for associated constants, which would help bitfields and similar patterns feel first class.

Ralith:
> Besides Result those types are typically never imported in scope in user code

I've named them explicitly a number of times (e.g. `slice::Iter` comes in handy in APIs fairly regularly), but I'll allow that it's not common. `mpsc::Receiver` and `thread::spawn` are additional good examples of something that, while not heavily overloaded if we consider std in isolation, would be awfully vague unqualified. Free functions in `std::{mem, iter}` are similarly too terse to be comfortable to use unqualified.

> This is different for our DeviceFn/InstanceFn types which do count as a "main" type similar to the collections.

The point was that since there's two of them, neither has inherent primacy.

> How should we do this for the generated code, by the way? I was thinking to just have a bunch of in-line:

It's appealing to group everything defined by an extension inside the same module. This minimizes reference-chasing and makes it easy to inventory, and maps closely to idiomatic Rust. Might be nice to isolate the function pointer definitions a little since usually folks aren't looking for those, but e.g. having miscellaneous struct definitions right there alongside the new methods sounds nice. Perhaps `ash::ext::pfn::*`? `...::signatures::*`?

Unfortunate that this proposal transposes the current organization of the generator output, though... seems like that might be churny to implement.

MarijnS95:
> `mpsc::Receiver` and `thread::spawn` are additional good examples of something that, while not heavily overloaded if we consider std in isolation, would be awfully vague unqualified. Free functions in `std::{mem, iter}` are similarly too terse to be comfortable to use unqualified.

Agreed.

> The point was that since there's two of them, neither has inherent primacy.

Indeed, either has it, but it's not about the primacy between these two elements within the module.

> It's appealing to group everything defined by an extension inside the same module. This minimizes reference-chasing and makes it easy to inventory, and maps closely to idiomatic Rust. Might be nice to isolate the function pointer definitions a little since usually folks aren't looking for those, but e.g. having miscellaneous struct definitions right there alongside the new methods sounds nice. Perhaps `ash::ext::pfn::*`? `...::signatures::*`?
> 
> Unfortunate that this proposal transposes the current organization of the generator output, though... seems like that might be churny to implement.

Main problem here is that PFNs are used outside of and across extensions as well; that idea might fly but we have to be more careful where the first definition of the PFN is going to be (currently it is all over the place) and adjust the generator to import it from the right place.

We'll need to do `mod` scoping around `DeviceFn`, `InstanceFn`, `NAME` and `SPEC_VERSION`, let's start with that.

Good point on the churny-ness: I've already received comments about `ash` changing too much for "being a simple sys wrapper" before even releasing these changes. And while I agree that the builder changes and now the `Fn` separation is very churny, it's all for the better (and we're much more than just a `sys` crate with bindgen-like content).

Ralith:
Do we have any open questions here?

MarijnS95:
@Ralith I think I was going to `mod`-ify the low-level generated bindings, and then we see how to do tackle the hand-written bindings from there?

Still unsure what the best approach / naming would be.

Ralith:
I stand by my preference for module-per-extension as discussed, but I don't feel strongly enough to block things.

MarijnS95:
Yeah I think that is totally reasonable, just hope the rename from `CamelCase` types to `snake_case` modules for _every extension_ aren't going to be too much of a hassle for folks!

Still okay to keep `InstanceFn` and `DeviceFn` purely separate, or do you rather keep the combined structure proposed in https://github.com/ash-rs/ash/pull/734#issuecomment-1537293633?

Ralith:
> just hope the rename from CamelCase types to snake_case modules for every extension aren't going to be too much of a hassle for folks!

I think it's mitigated by a few things: most people are probably mostly using core functions, and extension struct types don't need to be named very often. Hard to see how we'd get out of this hole without renaming them *somehow*, and with a small number of references, the cost of having to make a change at all dominates the specifics of the change.

> Still okay to keep InstanceFn and DeviceFn purely separate, or do you rather keep the combined structure proposed in https://github.com/ash-rs/ash/pull/734#issuecomment-1537293633?

I don't really like the `Option` there. I'm more tempted by the idea of a single struct that populates device fns with panicing stubs (or even unchecked `Option`s), but if we were to do that, we might as well go all the way with a single monolithic table containing all extensions. Separate instance/device structs is most consistent with our existing, less error-prone paradigm of independently loaded tables.

MarijnS95:
My plan was still to have the separate `InstanceFn` and `DeviceFn` structs, and obviously separated per-extension as we already do (which is still complicated when multiple extensions are required, or when multiple extensions individually enable the same function, but those cases are few and far between).

Ralith:
Right, so on that backdrop, I think avoiding a combined structure makes the most sense. Relatively few extensions will end up needing both, anyway.

MarijnS95:
Pushed another revision that still needs some work. In particular:

- create modules for vendor prefixes, i.e. `mod amdx_shader_enqueue { ... }` -> `mod amdx { mod shader_enqueue { ... } }`;
- Make `extensions/<vendor>/<extension>/` modules public so that they can be referenced by the user, and remove `pub use` reexports here;
- Rename hand-written extension structs to not repeat the name already covered in the extension struct (e.g. to `DeviceFn` and `InstanceFn`?);
- Move `///` `struct` link comment to module (`//!`);
- Clean up `PFN_` definitions to reside in the "core" where possible, rather than having the core reference extensions (relies on the order in which things are running and get added to `fn_cache`).

@Ralith let me know if there's anything off with the above plans!

Changelog.md
```diff
@@ -33,6 +33,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bumped MSRV from 1.59 to 1.60 (#709)
 - Replaced `const fn name()` with associated `NAME` constants (#715)
 - Generic builders now automatically set `objecttype` to `<T as Handle>::ObjectType` (#724)
+- Separated low-level `*Fn` structs an high-level extension wrappers between instance and device functions, for the following extensions: (#734)
```
Ralith, line 36: typo: s/an/and/

Ralith (approved):

Ralith:
Plans sound good to me overall. A few thoughts:

> create modules for vendor prefixes

This sounds neat!

> Make extensions/<vendor>/<extension>/ modules public

I wonder if we should actually make these separate files. Maybe rustc will like smaller files better? Bit of a shot in the dark, I guess. Would make reviewing diffs of generated code a bit easier, at least -- github won't even render this one.

> Rename hand-written extension structs to not repeat the name

Should we forego renaming extensions that only have one class of function? That would reduce downstream churn, though I feel like the reduced consistency might be worse.

> Clean up PFN_ definitions to reside in the "core" where possible

Is this strongly motivated? It seems kinda graceful for extension PFNs to be defined in the extension module.

MarijnS95:
> I wonder if we should actually make these separate files. Maybe rustc will like smaller files better? Bit of a shot in the dark, I guess. Would make reviewing diffs of generated code a bit easier, at least -- github won't even render this one.

Definitely the plan now that we're doing module separation for the generated bindings anyway.

Still on the fence whether to do that for vendor prefixes only, or for every extension. Leaning towards the latter because some vendor prefixes (e.g. `KHR` and `EXT`) are rather heavy already.

> Should we forego renaming extensions that only have one class of function? That would reduce downstream churn, though I feel like the reduced consistency might be worse.

You sure like shoving lots of work my way :grimacing:, after previously expressing the desire to perform this rename consistently:

https://github.com/ash-rs/ash/pull/734#issuecomment-1537293633

> Having some extensions be `FooFn` and some be `foo::{InstanceFn, DeviceFn}` would be too confusing, and being clear about the stage an extension is usable at seems nice even if it happens to only have one.

---

> > Clean up PFN_ definitions to reside in the "core" where possible
> 
> Is this strongly motivated? It seems kinda graceful for extension PFNs to be defined in the extension module.

Seems like you cut out the most important "rather than having the _core reference extensions_" part in the quote. Said in a different way: this is about `PFN_`s that are used in extensions as well as in the core (after e.g. stabilization): currently it seems to be rather inconsistent where they end up being defined, but it did not matter to the user as everything is reexported anyway. Not so much now when `PFN_`s will end up in extension modules when they are generated ad-hoc (which we can no longer do, fwiw).

MarijnS95:
Converted some but not all of the extension wrapper struct names to `struct Instance` or `struct Device`, let's see how this flies.

Ralith:
> You sure like shoving lots of work my way 😬, after previously expressing the desire to perform this rename consistently:

Just wanted to make sure the option was considered, and wasn't sure if we'd discussed it yet. As I said, I do lean towards consistency, but I was interested in your thoughts.

> currently it seems to be rather inconsistent where they end up being defined

Ah, I missed that. Definitely want to be consistent here.

Friz64:
Hey, I want to weigh in on this to make sure we settle on a definite loader design and avoid any additional potential breaking changes once the new generator goes into action.

I think it's pretty confusing to have both `ash::vk::khr_swapchain` and `ash::extensions::khr::swapchain`. Everything from the former should be moved into the latter. Along with that, merge `Device` and `DeviceFn`. Is there a good justification for why the wrapper functions shouldn't be directly defined on the struct that also holds the function pointers?

I've always found the "fallback functions" that get called when the original function fails to load weird. I understand that the motivation behind them is to avoid a null-check every time a function gets called, as the alternative would be to store each function pointer in an `Option`. I once investigated this because someone brought it up in an erupt issue: https://gitlab.com/Friz64/erupt/-/issues/27. The conclusion was that one branch instruction is insignificant compared to the cost of a whole Vulkan API call.

But overall, I really like this!

---

Other than that... I want to throw this out there: Why not have a table for Entry, Instance, and Device each, that holds all function pointers? [`erupt` did this.](https://docs.rs/erupt/0.23.0+213/erupt/struct.DeviceLoader.html) The loading API becomes really pretty, as we can figure out which extensions to load by reading the `ppEnabledExtensionNames` array behind the scenes. I encourage poking around erupt to see what I mean by this:
- User facing API for 98% of use-cases: [`DeviceLoader::new`](https://docs.rs/erupt/0.23.0+213/erupt/struct.DeviceLoader.html#method.new)
- [`DeviceLoaderBuilder` implementation](https://docs.rs/erupt/0.23.0+213/src/erupt/lib.rs.html#572-690) 
- [`DeviceEnabled::new` implementation](https://docs.rs/erupt/0.23.0+213/src/erupt/generated/mod.rs.html#1594) 
- [`DeviceLoader::custom` implementation](https://docs.rs/erupt/0.23.0+213/src/erupt/generated/mod.rs.html#2867) 
  - example: [loading `get_physical_device_present_rectangles_khr`](https://docs.rs/erupt/0.23.0+213/src/erupt/generated/mod.rs.html#1091-1102)

Users would also e.g. be able pass around one single object for all their device functions needs. And regarding the stack size of this, I think that we should just store it in an `Arc` by default (related discussion @ #731).

I understand that this would be a major breaking change, but I don't want to leave it unmentioned.

Ralith:
> I think it's pretty confusing to have both ash::vk::khr_swapchain and ash::extensions::khr::swapchain. Everything from the former should be moved into the latter.

I think this distinction arises from, and preserves, the pre-existing isolation of raw Vulkan types from high-level wrappers. We shouldn't remove it unless we also remove that elsewhere, which I think should be out of scope of this PR.

> Along with that, merge Device and DeviceFn

My knee-jerk reaction is that this makes sense. Function pointer tables aren't really raw Vulkan types, and hence it's unclear why someone might want to use one in isolation.

> one branch instruction is insignificant compared to the cost of a whole Vulkan API call.

This is also out of scope for this PR, and has been discussed at length in the past. You may be able to find the discussion in an issue somewhere. Open a new issue if you really want to re-hash it.

> Why not have a table for Entry, Instance, and Device each, that holds all function pointers?

This is also out of scope for this PR, and has also been discussed at length in the past. We concluded that using extension tables to represent extension availability is a valuable pattern to nudge downstream code towards, as it's far less error prone than representing optional extension presence with a side channel and trying to remember to check if any time you use a function from it, if you can even remember which functions come from which extension.

Friz64:
Seeing a lot of "out of scope of this PR" 😅😅 Fair enough, as this was more of a general comment. I've opened these issues so that there is an fitting place for discussion:
- https://github.com/ash-rs/ash/issues/876
- https://github.com/ash-rs/ash/issues/877

> This is also out of scope for this PR, and has also been discussed at length in the past. We concluded that using extension tables to represent extension availability is a valuable pattern to nudge downstream code towards, as it's far less error prone than representing optional extension presence with a side channel and trying to remember to check if any time you use a function from it, if you can even remember which functions come from which extension.

Got it, thanks :)

Ralith:
> Got it, thanks :)

To be clear, I don't think it's necessarily a waste of time to revisit. Our current separate wrappers approach is imperfect (some functions may be dynamically absent even if an extension is loaded successfully, `ash::Device` is verison-independent, etc) and I've gone back and forth myself. Feel free to open an issue for that too if you like; at least it'll centralize the discussion.

MarijnS95:
Yeah, glad that you figured out yourself that any and all of these new points are outside the scope of this already-massive PR and will only serve as followup-improvements or are unrelated altogether. I've flagged them as off-topic (and will flag this reply too) so that is's a tad easier to keep up with the latest desire and progress of _this_ PR.

Ralith:
Fixed up the CI failures, and implemented the "create modules for vendor prefixes" task. This required a hack for modules whose names begin with a number (e.g. `VK_KHR_16bit_storage`); I resolved this by prefixing a `_` but I'm open to alternatives.

generator/src/lib.rs
```diff
@@ -1331,58 +1340,32 @@ pub fn generate_extension_commands<'a>(
...
     });
 
-    quote! {
-        pub mod #extension_ident {
-            use super::*;
+    (
+        vendor,
+        quote! {
+            pub mod #extension_ident {
+                use crate::vk::extensions::*;
```
MarijnS95: Any reason to not use `super::super`?

generator/src/lib.rs
```diff
@@ -1278,8 +1283,12 @@ pub fn generate_extension_commands<'a>(
...
     let byte_name_ident = Literal::byte_string(format!("{extension_name}\0").as_bytes());
 
-    let extension_name = extension_name.strip_prefix("VK_").unwrap().to_lowercase();
-    let extension_ident = format_ident!("{}", extension_name);
+    let extension_name = extension_name.strip_prefix("VK_").unwrap();
+    let (vendor, extension_ident) = extension_name.split_once('_').unwrap();
+    let extension_ident = match extension_ident.chars().next().unwrap().is_ascii_digit() {
+        false => format_ident!("{}", extension_ident.to_lowercase()),
+        true => format_ident!("_{}", extension_ident.to_lowercase()),
+    };
```
MarijnS95, line 1291: Re. your question: how about adding a little comment here explaining this?

generator/src/lib.rs
Ralith: Multiple levels of relative references make for poor readability, IMO. I'm also not actually sure if that works.

MarijnS95: It does, and there's a comment now (on the generator side) explaining that this is strictly for including things imported at the top of the file (not outside of the `extenions.rs` file at least).

We could also define a `use super::*` on both `mod`s and get the same effect.

Ralith: It's generated code, and it works, so I'm happy regardless. No need for more churn.

PR #735, opened 2023-04-05
generator: Run `rustfmt` over generated output before writing to disk

### LINKED ISSUE #653: generator doesn't run `rustfmt`
Hey, I want to add support for `VK_EXT_mesh_shader` to ash. Unfortunately I have a problem running the generator. There are extra spaces between the code.
```Rust
use std :: fmt ; use crate :: vk :: bitflags :: * ; use crate :: vk :: definitions :: * ; use crate :: vk :: enums :: * ;
```
I did the following to get the ash generator running:
* I downloaded `LLVM` for `bindgen` and set `LIBCLANG_PATH` to `C:\Program Files\LLVM\bin`.
* I cloned the newest version of ash
* I ran `cargo run --release --bin generator` from the root directory.

The generator runs without issues, but the code looks weird. Running `cargo fmt` works, I just was curious if its a bug or just intended behaviour.

MarijnS95:
Fixes #653

Some users are confused by seeing wildly different output after running the generator, which is simply solved by running `rustfmt`.  As this is both confusing and "somewhat" slow, invoke `rustfmt` directly within the generator by piping string contents through it before redirecting to disk.  This not only makes the output consistent, it is the fastest way to reformat generator changes by omitting the round-trip to disk entirely, nor having `rustfmt` recursively go through the workspace and all files (including those that are not generated).

On a many-core machine these times are a bit skewed, but I want to include them to prove the "speed" point nevertheless, even if simplicity and consistency is the main reason to make this change:

Before:

    time ./target/debug/generator && time cargo fmt --all
    ./target/debug/generator  3.51s user 1.25s system 99% cpu 4.769 total
    cargo fmt --all  0.79s user 0.06s system 99% cpu 0.853 total

After:

    time ./target/debug/generator
    ./target/debug/generator  4.51s user 0.41s system 99% cpu 4.931 total

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -2937,19 +2937,18 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
     let vk_dir = src_dir.join("vk");
     std::fs::create_dir_all(&vk_dir).expect("failed to create vk dir");
 
-    let mut vk_features_file = File::create(vk_dir.join("features.rs")).expect("vk/features.rs");
-    let mut vk_definitions_file =
+    let vk_features_file = File::create(vk_dir.join("features.rs")).expect("vk/features.rs");
+    let vk_definitions_file =
         File::create(vk_dir.join("definitions.rs")).expect("vk/definitions.rs");
-    let mut vk_enums_file = File::create(vk_dir.join("enums.rs")).expect("vk/enums.rs");
-    let mut vk_bitflags_file = File::create(vk_dir.join("bitflags.rs")).expect("vk/bitflags.rs");
-    let mut vk_constants_file = File::create(vk_dir.join("constants.rs")).expect("vk/constants.rs");
-    let mut vk_extensions_file =
-        File::create(vk_dir.join("extensions.rs")).expect("vk/extensions.rs");
-    let mut vk_feature_extensions_file =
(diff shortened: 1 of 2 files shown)


generator/src/lib.rs
```diff
@@ -3019,18 +3018,32 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
...
+            .stdin(std::process::Stdio::piped())
+            .stdout(out)
+            .spawn()
+            .expect("Failed to spawn `rustfmt`");
+        let mut stdin = child.stdin.take().expect("Failed to open stdin");
+        stdin.write_all(text).unwrap();
+        drop(stdin);
+
+        let status = child.wait().unwrap();
+        assert!(status.success());
```
MarijnS95, line 3032: Probably even faster if we batch this wait for all the files below... But that's a bit over the top.

MarijnS95:
```
Benchmark 1: ./target/release/generator
  Time (mean ± σ):      2.165 s ±  0.041 s    [User: 1.795 s, System: 0.363 s]
  Range (min … max):    2.090 s …  2.218 s    10 runs
```

```
Benchmark 1: ./target/release/generator
  Time (mean ± σ):      1.577 s ±  0.025 s    [User: 1.784 s, System: 0.369 s]
  Range (min … max):    1.532 s …  1.612 s    10 runs
```

Trivial and significant wins after all.

Ralith (approved):
Nice!

Issue #737, opened 2023-04-08
get_calibrated_timestamps should only return one value for maxDeviation


kanashimia:
It should return `VkResult<(Vec<u64>, u64)>` instead of `VkResult<(Vec<u64>, Vec<u64>)>` as per spec

Bug here:
```rust
/// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetCalibratedTimestampsEXT.html>
///
/// Returns a tuple containing `(timestamps, max_deviation)`
#[inline]
pub unsafe fn get_calibrated_timestamps(
    &self,
    device: vk::Device,
    info: &[vk::CalibratedTimestampInfoEXT],
) -> VkResult<(Vec<u64>, Vec<u64>)> {
    let mut timestamps = vec![0u64; info.len()];
    let mut max_deviation = vec![0u64; info.len()];
    (self.fp.get_calibrated_timestamps_ext)(
        device,
        info.len() as u32,
        info.as_ptr(),
        timestamps.as_mut_ptr(),
        max_deviation.as_mut_ptr(),
    )
    .result_with_success((timestamps, max_deviation))
}
```

Spec xml:
```xml
<command successcodes="VK_SUCCESS" errorcodes="VK_ERROR_OUT_OF_HOST_MEMORY,VK_ERROR_OUT_OF_DEVICE_MEMORY">
    <proto><type>VkResult</type> <name>vkGetCalibratedTimestampsEXT</name></proto>
    <param><type>VkDevice</type> <name>device</name></param>
    <param><type>uint32_t</type> <name>timestampCount</name></param>
    <param len="timestampCount">const <type>VkCalibratedTimestampInfoEXT</type>* <name>pTimestampInfos</name></param>
    <param len="timestampCount"><type>uint64_t</type>* <name>pTimestamps</name></param>
    <param><type>uint64_t</type>* <name>pMaxDeviation</name></param>
</command>
```

Spec: pMaxDeviation is a pointer to a 64-bit unsigned integer value in which the strictly positive maximum deviation, in nanoseconds, of the calibrated timestamp values is returned.
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetCalibratedTimestampsEXT.html

Ralith:
Thanks; I confirm your analysis. This is not a soundness issue since we initialize the data regardless, but it is misleading. Want to make a PR with a fix?

MarijnS95:
[...]

PR #738, opened 2023-04-09
extensions/ext: Make `get_calibrated_timestamps()` return single value for `max_deviation`

kanashimia:
Per spec `max_deviation` is a single value, and not a vector.

Closes https://github.com/ash-rs/ash/issues/737


Ralith (approved):

MarijnS95:
Can you make the changelog entry similar to the changed title? And do you want this to be backported to 37.3, in which case we need to `deprecate` the current function and add one with the updated function signature?

MarijnS95 (approved):

MarijnS95:
I'm fine not backporting this to `0.37` for now.

PR #739, opened 2023-04-11
Bump libloading to 0.8

djkoloski:


MarijnS95:
That's pretty rapid, the update is only out for 3 hours 😅... Any update on what's new in this semver-breaking release, why we need it, and if it needs to be backported to the latest stable release?

djkoloski:
Oh! Sorry, I didn't mean to make it seem like this was urgent. 😅 I just happened to be in the right place at just the right time to notice this bump.

The only major change I'm aware of is that `libloading` switched from `winapi` to `windows-sys`, which I'm excited for since it slims down my dependencies. I figured it was a quick change to make and wanted to be helpful. 🙂

MarijnS95:
Okay thanks, ditching `winapi` is definitely worth upgrading for. I'll do some extra scanning of the changelog to make sure this doesn't change any other semantics. However, as `libloading` is technically part of our public API in `LoadingError::LibraryLoadFailure` we cannot backport this to the current stable release.

The next breaking `ash` release is something I'd like to slowly start creeping towards though as we've accumulated quite a few useful breaking features now, but it is also right around this time that a lot of technically breaking changes are flowing in and I'd rather see that stabilize first.

Ralith:
> libloading switched from winapi to windows-sys

Hmm, isn't the latest guidance to use windows-bindgen?

> However, as libloading is technically part of our public API in LoadingError::LibraryLoadFailure

Perhaps we should fix this.

MarijnS95:
@Ralith `windows-bindgen` is the way to go, yes, but it's fairly new and I hope `libloading` can migrate to that without yet another breaking bump. Regardless, I've asked about it at https://github.com/nagisa/rust_libloading/pull/120#issuecomment-1503920648 though Jake might perform this conversion soon enough.

I'm fine dropping `libloading` from the public thiserror-like `enum` fwiw, but can only do that on `main` again.

AngelicosPhosphoros:
@MarijnS95 If you closed my PR for this, when this PR would be merged? It's been half a year already.

MarijnS95:
@AngelicosPhosphoros it's been merged now?

AngelicosPhosphoros:
Hooray!

PR #741, opened 2023-04-13
Update Vulkan-Headers to 1.3.251

MarijnS95:


gents83:
Would this possibly bump a new release version?

PR #743, opened 2023-04-13
README: Remove deprecated `builder()` snippets and guidelines

### LINKED ISSUE #742: `*Info::builder()` still mentioned in the README.md despite being removed.
Title.

MarijnS95:
Fixes #742

#602 introduced builder functions directly on the raw Vulkan struct types by using lifetime borrows which are FFI compatible (ABI is identical) wuth raw pointers, simplifying the whole system and protecting the user against losing lifetimes upon calling `.build()`.  However, this change wasn't propagated through to the `README` so the code snippets were still showcasing removed `::builder()` and `.build()` functions and documenting "the `.build()` footgun" which doesn't even exist anymore 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,29 +70,14 @@ Each Vulkan handle type is exposed as a newtyped struct for improved type safety
 `T::null()`, and handles may be freely converted to and from `u64` with `Handle::from_raw` and `Handle::as_raw` for
 interop with non-Ash Vulkan code.
 
-### Default implementation for all types
-
-```rust
-// No need to manually set the structure type
-let desc_alloc_info = vk::DescriptorSetAllocateInfo {
-    descriptor_pool: self.pool,
-    descriptor_set_count: self.layouts.len() as u32,
-    p_set_layouts: self.layouts.as_ptr(),
-    ..Default::default()
-};
-```
-
 ### Builder pattern
 
 ```rust
-// We lose all lifetime information when we call `.build()`. Be careful!
-let queue_info = [vk::DeviceQueueCreateInfo::builder()
+let queue_info = [vk::DeviceQueueCreateInfo::default()
     .queue_family_index(queue_family_index)
-    .queue_priorities(&priorities)
-    .build()];
+    .queue_priorities(&priorities)];
 
-// We don't need to call `.build()` here because builders implement `Deref`.
-let device_create_info = vk::DeviceCreateInfo::builder()
+let device_create_info = vk::DeviceCreateInfo::default()
     .queue_create_infos(&queue_info)
     .enabled_extension_names(&device_extension_names_raw)
     .enabled_features(&features);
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -95,56 +87,20 @@ let device: Device = instance
...
-
-```rust
-let queue_info = vk::DeviceQueueCreateInfo::builder()
-    .queue_family_index(queue_family_index)
-    .queue_priorities(&priorities);
-
-let device_create_info = vk::DeviceCreateInfo::builder()
-    .queue_create_infos(std::slice::from_ref(&queue_info))
-    ...;
-```
```
MarijnS95, line 108: Not sure if it's still useful to mention `std::slice::from_ref`? I'm unsure if `let x = [vk::...]` or `let x = &[vk::...]` make an extra stack allocation or if there are other cases where moving a singular Vulkan struct into an array of length-1 is not feasible (i.e. if you only have a borrow of a struct as function parameter, but the Vulkan API requires a slice).

Ralith: Note inline arrays can be used just fine if the structure doesn't need to outlive the enclosing statement, e.g. when constructed directly in a Vulkan function's argument list, which is my personal preferred pattern.

I think `slice::from_ref` is a bit niche and there's no harm in constructing intermediate arrays if you need to. It's not really meaningful to discuss whether there's additional stack allocation there, since LLVM will happily shuffle such straightforward things around into whatever form makes the most sense.

Ralith (approved):

README.md
MarijnS95: Fair enough, I do think `let` bindings have their place in massive nested structs like the render passes listed above, but we at least have the full borrow-checking now to validate soundness and users can write it however they prefer :+1: 

Good, then I'll leave out the `from_ref` mention entirely.

README.md
```diff
@@ -95,56 +87,20 @@ let device: Device = instance
...
-let mut device_create_info = vk::DeviceCreateInfo::builder()
+let mut device_create_info = vk::DeviceCreateInfo::default()
     .push_next(&mut corner)
     .push_next(&mut variable_pointers);
 ```
 
-Pointer chains in builders differ from raw Vulkan. Instead of chaining every struct manually, you instead use `.push_next` on the struct that you are going to pass into the function. Those structs then get *prepended* into the chain.
-
-`push_next` is also type checked, you can only add valid structs to the chain. Both the structs and the builders can be passed into `push_next`. Only builders for structs that can be passed into functions will implement a `push_next`.
+Pointer chains in Ash differ from raw Vulkan. Instead of chaining every struct manually, you instead use `.push_next()` on the struct that you are going to pass into the function. Those structs then get _prepended_ into the chain.
```
MarijnS95, line 101: Nit: I don't think "Pointer chains in Ash differ from raw Vulkan" is the right wording, but it was already there. They are still the same pointer chains, it's just that our `.push_next()` helper function has a specific way to "`.insert(1)`" rather than `.append()` to be simple and efficient (no need to walk to the end of the chain, though we do need to do that for the struct that is being appended, but the common case is that it's just one item).

At the same time [the spec says](https://registry.khronos.org/vulkan/specs/1.3/styleguide.html#extensions-interactions) that specific `pNext` ordering is allowed (but strongly discouraged) if drivers/extensions absolutely need it for deterministic behaviour.

Ralith: I agree this wording could be improved, both for clarity and concision. Perhaps:

> Ash provides helpers for assembling pointer chains. Use `.push_next()` on the struct ...

MarijnS95: Updated and reordered. Left out the "refer to the `push_next()` documentation for more detailed insertion order" or something along those lines.

Note that our `fn push_next()` docs don't cover the case where a hypothetical `D->E` chain is appended to an `A->B->C` chain, which will end up looking like `A->D->E->B->C`.

README.md
```diff
@@ -95,56 +87,21 @@ let device: Device = instance
...
+let mut device_create_info = vk::DeviceCreateInfo::default()
     .push_next(&mut corner)
     .push_next(&mut variable_pointers);
 ```
 
-Pointer chains in builders differ from raw Vulkan. Instead of chaining every struct manually, you instead use `.push_next` on the struct that you are going to pass into the function. Those structs then get *prepended* into the chain.
-
-`push_next` is also type checked, you can only add valid structs to the chain. Both the structs and the builders can be passed into `push_next`. Only builders for structs that can be passed into functions will implement a `push_next`.
-
+The templated argument of `.push_next()` only allows valid structs to extend a given struct (known as [`structextends` in the Vulkan registry](https://registry.khronos.org/vulkan/specs/1.3/styleguide.html#extensions-interactions), mapped to `Extends*` traits).
```
Ralith, line 103:
```suggestion
The generic argument of `.push_next()` only allows valid structs to extend a given struct (known as [`structextends` in the Vulkan registry](https://registry.khronos.org/vulkan/specs/1.3/styleguide.html#extensions-interactions), mapped to `Extends*` traits).
```

Ralith (approved):
Some nits, otherwise LGTM.

README.md
MarijnS95: @Ralith sure thing, I've been weighing this exact wording and indeed opted for the wrong one, thanks!

PR #744, opened 2023-04-16
extensions/khr: Take the remaining `p_next`-containing structs as `&mut`

### LINKED ISSUE #740: get_physical_device_surface_capabilities2 should take a mutable reference to vk::SurfaceCapabilities2KHR
The function `vkGetPhysicalDeviceSurfaceCapabilities2KHR` can be extended by use of `VkSurfaceCapabilitiesKHR.pNext`, however the current wrapper `get_physical_device_surface_capabilities2` doesn't allow this since it creates and returns the struct  `vk::SurfaceCapabilities2KHR`. 

To allow the use of the `pNext` pointer,  the function `get_physical_device_surface_capabilities2` should be changed from this:
```rust
  pub unsafe fn get_physical_device_surface_capabilities2(
      &self,
      surface_info: &vk::PhysicalDeviceSurfaceInfo2KHR,
  ) -> VkResult<vk::SurfaceCapabilities2KHR>
```
To something like this:
```rust
  pub unsafe fn get_physical_device_surface_capabilities2(
      &self,
      surface_info: &vk::PhysicalDeviceSurfaceInfo2KHR,
      surface_capabilities: &mut vk::SurfaceCapabilities2KHR,
  ) -> VkResult<()>
```

MarijnS95:
Fixes #740

Version 2 of `get_physical_device_surface_capabilities` and the matching `vk::SurfaceCapabilitiesKHR` struct exist solely to provide an `sType` and `pNext` field to allow extending the original query with additional data via extension structs.  However, this API when introduced in #530 only returns the `default()`-initialized struct making it just as constrained as `get_physical_device_surface_capabilities()`.  Solve this by taking `vk::SurfaceCapabilities2KHR` as `&mut` just like any similar API.

And just like this, do the same for the remaining:

- `AccelerationStructure::get_acceleration_structure_build_sizes()`
- `ExternalMemoryFd::get_memory_fd_properties()`
- `ExternalMemoryWin32::get_memory_win32_handle_properties()`

In case these structs get extended somewhere down the line, which the Vulkan API allows for.

### DIFF
--- a/ash/src/extensions/khr/acceleration_structure.rs
+++ b/ash/src/extensions/khr/acceleration_structure.rs
@@ -259,7 +259,7 @@ impl AccelerationStructure {
         &self,
         version: &vk::AccelerationStructureVersionInfoKHR,
     ) -> vk::AccelerationStructureCompatibilityKHR {
-        let mut compatibility = vk::AccelerationStructureCompatibilityKHR::default();
+        let mut compatibility = mem::zeroed();
 
         (self.fp.get_device_acceleration_structure_compatibility_khr)(
             self.handle,
--- a/ash/src/extensions/khr/acceleration_structure.rs
+++ b/ash/src/extensions/khr/acceleration_structure.rs
@@ -5,6 +5,7 @@ use crate::{Device, Instance};
 use std::ffi::CStr;
 use std::mem;
 
+/// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_acceleration_structure.html>
 #[derive(Clone)]
 pub struct AccelerationStructure {
     handle: vk::Device,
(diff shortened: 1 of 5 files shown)


MarijnS95:
Searching for `default`, we only have:

- `AccelerationStructure::get_acceleration_structure_build_sizes`;
- `ExternalMemoryFd::get_memory_fd_properties`;
- and `ExternalMemoryWin32::get_memory_win32_handle_properties`

that are `default()`-initializing a `pNext` struct, perhaps those should get the same treatment (in this PR) to get rid of this problem altogether?

(On a side-note `get_device_acceleration_structure_compatibility()` is the only other function calling `default()` but for a flag where we typically use `mem::zeroed()` instead)

ash/src/extensions/khr/acceleration_structure.rs
```diff
@@ -258,7 +258,7 @@ impl AccelerationStructure {
         &self,
         version: &vk::AccelerationStructureVersionInfoKHR,
     ) -> vk::AccelerationStructureCompatibilityKHR {
-        let mut compatibility = vk::AccelerationStructureCompatibilityKHR::default();
```
Ralith, line 261: Wrong branch?

Ralith (approved):

Ralith:
> perhaps those should get the same treatment (in this PR) to get rid of this problem altogether?

SGTM.

Ralith (approved):

PR #745, opened 2023-04-20
analysis: Introduce Vulkan XML parser

Friz64:
This adds a parser for Vulkan XML registries, namely `vk.xml` and `video.xml`. It doesn't have a 100% coverage of all possible attributes, and doesn't aim to do so. Any additional attributes that are not extracted will simply be ignored. It doesn't try to process the XML in any way at all, even for simple things such as parsing a number.


Jerrody:
@Friz64 
Yet another wonderful PR! Thank you!

.github/workflows/ci.yml
```diff
@@ -46,6 +46,8 @@ jobs:
       - name: Install Vulkan loader
         run: sudo apt-get install libvulkan-dev
       - uses: actions/checkout@v1
+        with:
+          submodules: true
```
MarijnS95, line 50: We don't have any submodules, right?

analysis/src/lib.rs
```diff
@@ -1,9 +1,33 @@
...
+}
+
+#[derive(Debug)]
+pub struct Library {
+    _xml: xml::Registry,
+}
+
+impl Library {
+    fn new(xml_path: impl AsRef<Path>) -> Library {
+        let xml_input = Box::leak(fs::read_to_string(xml_path).unwrap().into_boxed_str());
```
MarijnS95, line 28: Why leak this, for roxml lifetimes? Should we clean it up in `Drop`, or is this intentionally skipped to keep all the string names `'static` (in which case that detail should be documented here)?

analysis/src/xml.rs
```diff
@@ -0,0 +1,715 @@
...
+                                Some("handle") => {
+                                    registry.handle_aliases.push(Alias::from_node(type_node));
+                                }
+                                Some("enum") => {
+                                    registry.enum_aliases.push(Alias::from_node(type_node));
+                                }
+                                Some("struct") => {
+                                    registry.struct_aliases.push(Alias::from_node(type_node));
+                                }
+                                _ => (),
```
MarijnS95, line 106: Warn about new/unhandled items? Same below.

analysis/src/xml.rs
```diff
@@ -0,0 +1,715 @@
...
+                        }
+                    }
+                }
+                "enums" => match registry_child.attribute("type") {
+                    Some("enum") => registry.enums.push(Enum::from_node(registry_child, api)),
+                    Some("bitmask") => registry
+                        .bitmasks
+                        .push(BitMask::from_node(registry_child, api)),
+                    _ => {
+                        if matches!(registry_child.attribute("name"), Some("API Constants")) {
```
MarijnS95, line 145: No `==` with perhaps a `.as_deref()`?

analysis/src/xml.rs
```diff
@@ -0,0 +1,715 @@
...
+                        .push(Feature::from_node(registry_child, api));
+                }
+                "extensions" => {
+                    for extension_node in registry_child
+                        .children()
+                        .filter(|node| node.has_tag_name("extension"))
+                        .filter(|node| {
+                            node.attribute("supported")
+                                .map(|values| values.split(',').any(|support| support == api))
+                                .unwrap_or(true)
```
MarijnS95, line 189: Deduplicate with `api_matches`? Only the attribute name is different.

analysis/src/xml.rs
```diff
@@ -0,0 +1,715 @@
...
+            name: attribute(node, "name").unwrap(),
+            requires: attribute(node, "requires"),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct BaseType {
+    pub name: XmlStr,
+    /// `None` indicates this being a platform-specific type.
```
MarijnS95, line 238: `None` (and `false` for that matter) can be intradoc links.

analysis/src/xml.rs
```diff
@@ -0,0 +1,715 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct BitMask {
+    pub name: XmlStr,
+    pub bits: Vec<BitMaskBit>,
+    /// Some bitmask variants represent literal values instead of specific
+    /// individual bits, e.g. a combination of bits, or no bits at all. An good
```
MarijnS95, line 447:
```suggestion
    /// individual bits, e.g. a combination of bits, or no bits at all. A good
```

analysis/src/xml.rs
```diff
@@ -0,0 +1,715 @@
...
+    /// example for this is in `VkCullModeFlagBits`.
```
MarijnS95, line 448:
```suggestion
    /// example for this is `VkCullModeFlagBits`.
```

Perhaps point to `FRONT_AND_BACK`?

analysis/src/xml.rs
```diff
@@ -0,0 +1,715 @@
...
+fn attribute(node: Node, name: &str) -> Option<XmlStr> {
+    node.attribute_node(name)
+        .map(|attr| match attr.value_storage().clone() {
+            StringStorage::Borrowed(s) => Cow::Borrowed(s),
+            StringStorage::Owned(s) => Cow::Owned((*s).into()),
+        })
+}
+
+/// Retrieves the text inside the next child element of `node` named `name`.
+fn child_text(node: Node, name: &str) -> Option<XmlStr> {
```
MarijnS95, line 22: Does `descendant_text()` probably do the same, and is this function only a simpler (borrowing) variant that assumes there's never extra (text/comment) elements within the given `name` descendant?

MarijnS95 (commented):
Looking very clean and simple, I really like it thus far!

Curious to see how much processing we're going to have to do on top and how sane we can make it (consider e.g. recent issues about multiple features/extensions depending on and providing the same functions and types, etc).

analysis/src/xml.rs
Friz64: There's a difference between children and descendants in the sense that "children" only describe the nodes **directly** below the given one, whereas "descendants" will give you every children _and_ their sub-children.

This function will find the first direct child of the given node, and will then return you its text content in borrowed form. With "text content" I am referring to the node's first text child (which is what roxmltree's `.text_storage()` will give you). This is why there is no concern for extra comment elements.

This is all a bit unclear, we should add more comments to this code.

analysis/src/xml.rs
Friz64: I designed this parser around it only filtering out data WE want, and ignoring everything else. If new nodes with an unknown category get added, and we want to use that data, we can easily expand this. It's not often that there are new types of data points which are of use to our generator, and adding checks everywhere throughout the code to look for these feels unnecessary in my opinion.

analysis/src/xml.rs
Friz64: Let's change it if you prefer it this way.

analysis/src/xml.rs
Friz64: This is the only place in the code where we have to filter for the string "supported" instead of the much more common "api", so I decided not to disturb the `api_matches` function and just have a special case here. We could also have `api_matches` call another function `attribute_matches` with this code, and then use that here.

analysis/src/xml.rs
MarijnS95: Sure, the point was more so that `child_text()` is only intended for simple nodes that contain some text and no descendant nodes with text nor comments (and if they do, that is omitted). In that sense calling `descendant_text()` on the `child` node found here returns the same value. But the simplicity and borrowing nature of this function, together with the ability to select a child element of `node` probably warrants its existence.

(Note: I did check that calling `descendant_text(child)` is always identical to `child.text_storage()` here)

analysis/src/xml.rs
```diff
@@ -0,0 +1,715 @@
...
+    let child = node.children().find(|node| node.has_tag_name(name));
```
MarijnS95, line 23: Do you think we need more `StringStorage` -> `Cow` conversions, and should we factor it out into a function?

analysis/src/lib.rs
MarijnS95: I think it's worthwhile too, but let's document that for future readers.

MarijnS95: (In the current generator, wherever I've been able to clean up `String` copies, I just ended up threading through an explicit lifetime. Little less convenient to write, but skips the lifetime issues)

analysis/src/xml.rs
Friz64: > though a `debug!()` stating all the features that we haven't implemented (for sensible reasons) might still be nice

Implemented this. I have also set up `env_logger` to go along with this.

analysis/src/xml.rs
Friz64: It's not exactly the same as `api`, because it's value can also be `disabled`.

Side note, this also means that we automatically skip all disabled extensions as a consequence of this.

Friz64: > We could also have `api_matches` call another function `attribute_matches` with this code, and then use that here.

So, to be clear, did you suggest that this should be done?

analysis/src/xml.rs
Friz64: There probably won't be any more conversions. We could factor it out, but I don't think it's really worth it currently.

Friz64:
Trying to get back into working on this. Sorry for the silence. Life is hard.

> I wonder if this crate should be renamed to `parser` or if you intend it to also contain the `analysis` stage (e.g. linking everything together) within it? (Probably doesn't make much sense to only have a purpose-built parser crate for Ash)

The intention was for this crate to contain the `analysis` stage, yes.

analysis/src/xml.rs
Friz64: > Implemented this. I have also set up `env_logger` to go along with this.

Hm, CI fails because `env_logger` 0.11, which came out just hours ago, has an MSRV of 1.71. It's probably not worth bumping our MSRV over this? So, is it fine depending on `env_logger` 0.10? Or maybe use another logging library?

Friz64:
The next layer on top of this should take care of all remaining parsing. 

For this, I was reading back on the conversation in #662, and the conclusion we seemingly came to for C parsing was to use a parsing library, not on the whole spec at once, but on small chunks, lifted from the XML. One idea was to leverage the `lang-c` library I had used in erupt for this, but after some testing [I learned that it is not really fit for that purpose](https://github.com/ash-rs/ash/discussions/662#discussioncomment-4066569).

Today, I again searched for other C parsing libraries we could use, and one stuck out to me: [`tree-sitter-c`](https://docs.rs/tree-sitter-c/). Tree-sitter is an parsing library written in C, with well-written and official Rust bindings. In terms of C libraries, it is pretty harmless with all its code included in the crate, so compiling it as part of the generator should not be an issue, even on Windows. `tree-sitter-c` is the official C grammar for it. I did some quick experiments throwing various code snippets at it, and it seems to be pretty much exactly what we are looking for. It was developed for use within code editors, so it does not require its input to be run through an pre-processor. It also doesn't require any context for parsing. This means that even something like `uint8_t pipelineIdentifier[VK_UUID_SIZE]`, where `lang-c` would have previously errored out because it doesn't know what `VK_UUID_SIZE` is supposed to be exactly, parses successfully. The downside of this on the other hand is that is only produces an _Concrete_ Syntax Tree, not a more higher-lever _Abstract_ Syntax Tree, but this is enough for our purposes.

eddyb:
> This means that even something like `uint8_t pipelineIdentifier[VK_UUID_SIZE]`, where `lang-c` would have previously errored out because it doesn't know what `VK_UUID_SIZE` is supposed to be exactly,

Sorry if this was discussed before (I couldn't easily find this specific point brought up), but:
```xml
<member><type>uint8_t</type>                <name>pipelineIdentifier</name>[<enum>VK_UUID_SIZE</enum>]</member>
```
The actual XML has `<type>`, `<name>` and `<enum>`, fully disambiguating the identifiers used in the declaration.

Even without the `<enum>` on `VK_UUID_SIZE`, the syntax is unambiguously `Type · "[" Expr "]"`.

<sub>(if you'll excuse my shorthand - `·` would be like the "declaration focus" or "declaree", trivial here but more important for e.g. pointer-to-array/function, where the type syntax turns inside-out with parenthesization - e.g. `X (Z ·) Y` => `(X Y) Z ·`, where X/Y/Z are any tokens, if you could parenthesize types in C, or if you use a C++ noop type alias to emulate it etc.)</sub>

---

If there are counter-examples to that idea, I'd love to see them, because it's the direction I would've went in, for my own coincidental `vk.xml` parsing needs (before I remembered this effort might end up being a reusable way to generate `ash`-compatible automation - [in theory even `vk-layer-for-rust`'s Python scripts](https://github.com/google/vk-layer-for-rust/tree/a2363607922d97a4dfc5b62d05c54848c73d671c/scripts) could potentially be fully replaced by something based on it, but that's out of scope for me for now).

I'm really tempted to try to at least categorize the shapes of child nodes (and text between them), which are currently flattened in this PR's `StructureMember`'s `c_declaration` field - I'll post an update if I get there.

<sub>(disclaimer: back in the C11 days, I wrote [a C declaration parser/compiler](https://github.com/eddyb-abandoned/wiREd/blob/main/cparse/cparse.js) on top of someone else's yacc grammar, so the quirks of the syntax, like order independence of specifiers and the inside-out mode when the "declaree" is parenthesized, are "burned-in", making custom C parsers more appealing than they should be - my only hope is Khronos having been sensible enough, really)</sub>

eddyb:
Looks like custom parsing was 100% viable. By far most of the effort was spent making it powerful enough to handle the 11 function pointer `typedef`s with basically no special-casing (if `vk.xml` ever gets wild enough to have the C equivalent of `fn() -> fn() -> fn() -> fn()`, maybe with bonus extra pointers and arrays as well, my parser should be ready for it).

Outside of that, the only 2D array is in `VkTransformMatrixKHR`, and everything else is `T*`, `const T*`, `T[N]`, and the occassional spicy `const T[N]` (thanks to function parameter decay, which I explicitly model in `CType`).

- my PR (against this PR's branch): https://github.com/Friz64/ash/pull/1
- example dump output (see PR's second commit): https://gist.github.com/eddyb/5ae0247468892ea7dd2a136ed3afa869

eddyb:
Finally dug enough into the old generator to see how the C parsing works there: https://github.com/ash-rs/ash/blob/92084df65f52aa15b704279fb6d8d26a3ee71809/generator/src/lib.rs#L86-L272

I would agree with anyone who would want to replace that, and here's my first impressions:
- using `nom` is comparable to my code on the parsing side, but feels clunkier for lexing
- it doesn't separate lexing and parsing which probably makes it more annoying than it needs to be
- it special-cases a lot of declarator patterns (various combinations of maybe-`const` pointers and/or arrays) like I first theorized, but then I moved away from that very quickly due to combinatorial explosion
- I am increasingly confused by `vk-parse`: it seems to _also_ contain some ad-hoc C parsing (_[oh dear](https://github.com/krolli/vk-parse/blob/11f1a1f44e9462cb4c610d9a01b6e2e9793d8c11/vk-parse/src/c.rs#L175-L189)_), but I guess it's not enough for the old `ash` generator?
  - should my more robust `cdecl` parser be upstreamed to `vk-parse`?
  - is this `ash` `analysis` crate work meant to be a replacement for `vk-parse`?

With some of these aspects I feel a bit out of my depth, especially when it comes to "what other efforts are out there and am I at risk of duplicating work?", but nothing else I've seen so far seems principled enough (at most there's a lot of reliance on the limited set of possible patterns which is somewhat reasonable I suppose).

MarijnS95:
> Finally dug enough into the old generator to see how the C parsing works there:
> [..]
>     * I am increasingly confused by `vk-parse`: it seems to _also_ contain some ad-hoc C parsing (_[oh dear](https://github.com/krolli/vk-parse/blob/11f1a1f44e9462cb4c610d9a01b6e2e9793d8c11/vk-parse/src/c.rs#L175-L189)_), but I guess it's not enough for the old `ash` generator?

We use both. `vk-parse` creates `vkxml` constructs that represent parsed C in some "unmanageable" way, and a bit of `nom`-fu was added to parse out things (like simple C `const` expressions) that we could not get through `vkxml`. As a general code, a lot of this has been hacked together out of dire need, with a focus on generated code, not on generator internals. That's why things are so messy, and why a new generator needs to be written that takes all our ideas into account from the get-go.

With this approach from @Friz64 we should no longer need to rely on `vk-parse` (even though it is still maintained. The 7-year-old unmaintained `vkxml` is entirely out of the question, of course).

---

A little warning: this PR sat stale since June 2023, and right as activity is picking back up now, I'll be away from computers all of February. I hope no-one is affected by my absence and feel free to keep working/discussing on a generator rewrite, but it'll have to do without my input for a while 👍

Friz64:
> It doesn't try to process the XML in any way at all, even for simple things such as parsing a number.

This is not really true anymore with https://github.com/Friz64/ash/pull/1. Other than that, I've come around to saying simple parsing doesn't hurt.

analysis/src/xml.rs
Friz64: Switched to `tracing` now, which is more modern, and has a lower MSRV than `env_logger`.

Friz64:
I've added support for the `tracing` crate. Here's a snip from the output when running via `RUST_LOG=trace`:
```
...
2024-03-13T21:00:30.028385Z TRACE xml{path=generator/Vulkan-Headers/registry/vk.xml}:type{node="<type name='VkColorSpaceKHR' category='enum'>"}: analysis::xml: encountered node
2024-03-13T21:00:30.028388Z TRACE xml{path=generator/Vulkan-Headers/registry/vk.xml}:type{node="<type name='VkCompositeAlphaFlagBitsKHR' category='enum'>"}: analysis::xml: encountered node
2024-03-13T21:00:30.028392Z TRACE xml{path=generator/Vulkan-Headers/registry/vk.xml}:type{node="<type name='VkDisplayPlaneAlphaFlagBitsKHR' category='enum'>"}: analysis::xml: encountered node
2024-03-13T21:00:30.028395Z TRACE xml{path=generator/Vulkan-Headers/registry/vk.xml}:type{node="<type name='VkPresentModeKHR' category='enum'>"}: analysis::xml: encountered node
2024-03-13T21:00:30.028399Z TRACE xml{path=generator/Vulkan-Headers/registry/vk.xml}:type{node="<type name='VkSurfaceTransformFlagBitsKHR' category='enum'>"}: analysis::xml: encountered node
2024-03-13T21:00:30.028403Z TRACE xml{path=generator/Vulkan-Headers/registry/vk.xml}:type{node="<type name='VkDebugReportFlagBitsEXT' category='enum'>"}: analysis::xml: encountered node
2024-03-13T21:00:30.028406Z TRACE xml{path=generator/Vulkan-Headers/registry/vk.xml}:type{node="<type name='VkDebugReportObjectTypeEXT' category='enum'>"}: analysis::xml: encountered node
...
```

Jasper-Bekkers:
I tried playing around with this and it's pretty good already, much nicer to use then vk-parse / vkxml. I'm using it for a small codgen task that would live outside of `ash`.

Two small bits of feedback

1. Extremely easy to use!
1. On `Analysis`, the `xml` field should be `pub`
2. This work would be quite a bit more powerful if it would handle also all the "C" parsing from the vk.xml files, right now I'm just moving / copying code over into my project from the old ash generator (things like converting uint32_t to u32, converting pointers and arrays etc). It's nice that the `c_declaration` exists, but it would be even nicer if a `rust_declarations` field also existed.

Friz64:
Hey, thanks for taking a look! It's good to have feedback from real world use cases. 

> On `Analysis`, the `xml` field should be `pub`

This PR didn't really try to expose a proper public API (although maybe it should have?). Either way, I want to point you to #867, which actually does so. I've gone ahead and rebased it for you.

> This work would be quite a bit more powerful if it would handle also all the "C" parsing from the vk.xml files, right now I'm just moving / copying code over into my project from the old ash generator (things like converting uint32_t to u32, converting pointers and arrays etc). It's nice that the `c_declaration` exists, but it would be even nicer if a `rust_declarations` field also existed.

C parsing is what https://github.com/Friz64/ash/pull/1 aims to accomplish. I recall that it's in a usable state already, just not 100% finished.

This PR is currently marked as draft because the intention is that it gets merged into here first.

Friz64:
I wanna try working on the new generator again :3
I rebased this branch and updated it to make it work with and read all the new additions from the latest `vk.xml`. Because the `Vulkan-Headers` submodule in the repo is over a year old, I added a separate one in the `generator-rewrite` folder. Also eddyb's C parsing is back in.


> This PR doesn't really try to expose a public API for the `analysis` crate (although maybe it should?).

It does now.

PR #746, opened 2023-05-01
Switch to `CStr::from_bytes_until_nul` on sized char arrays in structs

MarijnS95:
Certain structs contain sized character arrays that are converted to `CStr` for debugging using unsafe `CStr::from_ptr(...as_ptr())`.  There is no need to round-trip to a pointer and possibly read out of bounds if the NULL character index (string length) is instead scanned for manually on the slice or via the newly stabilized `CStr::from_bytes_until_nul()` since Rust 1.69 (which panics if no NULL char is found before the end of the slice).

Unfortunately `unsafe` is still needed to cast the array from a `c_char` (`i8` on most platforms) to `u8`, which is what `from_bytes_until_nul()` accepts.

---

Draft because this is a rather eager MSRV bump that we should probably give more time to be adopted by the community, and because the unsafe slice casts don't look all that pretty.


Ralith:
I think this would look a lot better wrapped up in a helper function.

MarijnS95:
@Ralith indeed, because of the unfortunate slice cast. I wish there were `from_chararray_until_nul` or similar functions, taking a `c_char`. The cast from `u8` to `c_char` is anyway safe according to `CStr::from_ptr()`.

ash/src/vk/prelude.rs
```diff
@@ -59,3 +59,13 @@ impl From<vk::Extent2D> for vk::Rect2D {
...
+
+#[cfg(feature = "debug")]
+pub(crate) fn wrap_cstr_slice_until_nul(
+    bytes: &[::std::os::raw::c_char],
+) -> Result<&::std::ffi::CStr, core::ffi::FromBytesUntilNulError> {
+    ::std::ffi::CStr::from_bytes_until_nul(
+        // SAFETY: The cast from c_char to u8 is ok because a c_char is always one byte.
+        unsafe { ::std::slice::from_raw_parts(bytes.as_ptr().cast(), bytes.len()) },
+    )
+}
```
MarijnS95, line 71: Dropped the `unwrap()` here so that `Ok("...")` shows up in `Debug`, but at least we won't panic if the struct isn't formattable because the string is invalid.

MarijnS95: The only kind of error `from_bytes_until_nul()` can return is `FromBytesUntilNulError`, which means it would have gone out of bounds to find a trailing nul (i.e. the given slice doesn't contain a nul character).

Replacement characters are only emitted for `to_string_lossy()` (and `to_str()` returns an error). `Debug`-printing a `CStr`, which is what happens here in `ash`, simply prints the byte-string via `self.to_bytes().escape_ascii()`, which via https://doc.rust-lang.org/std/ascii/fn.escape_default.html prints the ASCII letter or an escape character (e.g. `\x3c`, `\n`, etc).

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=afb5100e527fb9710de21d44cdce3d07

---

So we already get the desired behaviour, just wondering what would be the best way to represent such an invalid string. Since quotation marks are already included I'd fancy if we could either see:

```
device_name: "My GPU",
device_name: FromBytesUntilNulError,
```

Without being wrapped in (multiline) `Err`/`Ok`.

Ralith: Ah, right, `CStr`s aren't UTF-8 to begin with. The proposed formatting seems fine to me.

ash/src/vk/prelude.rs
```diff
@@ -59,3 +59,13 @@ impl From<vk::Extent2D> for vk::Rect2D {
 pub unsafe trait TaggedStructure {
     const STRUCTURE_TYPE: vk::StructureType;
 }
+
+#[cfg(feature = "debug")]
+pub(crate) fn wrap_cstr_slice_until_nul(
+    bytes: &[::std::os::raw::c_char],
+) -> Result<&::std::ffi::CStr, core::ffi::FromBytesUntilNulError> {
```
MarijnS95, line 66: Note that this error type was accidentally forgotten to be reexported from `std` and is only available there since 1.73 (took me a sec to figure out why it was listed in https://blog.rust-lang.org/2023/10/05/Rust-1.73.0.html#stabilized-apis), hence this imports from `core`. Perhaps I should import the rest from `core` as well and start the migration process to `no-std` gracefully.

Ralith: Preferring `core`/`alloc` where applicable in new code seems like a fine start!

MarijnS95: Done that here now!

Ralith (approved):

ash/src/vk/prelude.rs
```diff
@@ -59,3 +59,13 @@ impl From<vk::Extent2D> for vk::Rect2D {
...
     const STRUCTURE_TYPE: vk::StructureType;
 }
+
+#[cfg(feature = "debug")]
+pub(crate) fn wrap_cstr_slice_until_nul(
+    bytes: &[::core::ffi::c_char],
+) -> Result<&::core::ffi::CStr, ::core::ffi::FromBytesUntilNulError> {
+    ::core::ffi::CStr::from_bytes_until_nul(
+        // SAFETY: The cast from c_char to u8 is ok because a c_char is always one byte.
+        unsafe { ::core::slice::from_raw_parts(bytes.as_ptr().cast(), bytes.len()) },
```
Ralith, line 69: nit: the leading `::` probably isn't needed anywhere above; this isn't a macro.

MarijnS95: Fair point, I think that got adopted from the generated code (though that doesn't live in a macro and doesn't need the `::` prefix either), and it wasn't even applied consistently to e.g. `FromBytesUntilNulError` initially.

MarijnS95:
@Ralith taking this PR one step further, what is your thought on extending these structs with a `fn <field_name>_as_cstr()` for convenience to users?

That would imply the same MSRV bump unless we decide to implement it with the current `from_ptr(...as_ptr())` code already used in the `Debug` formatters.

Ralith:
That sounds nice to me! Any trivial operation that ~every user will have to do to make use of a field is a good candidate for wrapping.

MarijnS95:
Done in #831 without breaking MSRV, and rebased this PR to turn the `unsafe` function into a `Result` with appropriate error type.

(We could also make the current function be safe and return a `Result` by reimplementing that function, but I'd rather bump MSRV at that point...)

MarijnS95:
> That sounds nice to me! Any trivial operation that ~every user will have to do to make use of a field is a good candidate for wrapping.

On this note, do we also have cases where we might want to wrap a pointer+count in a slice?

In most cases it's probably not useful as the user already has access to their slice:

```rust
let mut my_data = Vec![...];
let mut vk_struct = vk::Something::builder().data(&mut my_data);
device.get_something(&mut vk_struct);
dbg!(my_data);
```

But it might be worth it for writing layers in Rust with `ash` (though not something that we're close to supporting yet). And it could also be used to work around https://github.com/ash-rs/ash/pull/815#discussion_r1392679017 (though I'd still rather fix the builder fn).

MarijnS95:
@Ralith if you're okay with this, Rust 1.69 is now 7 months old and helps us fix the existing soundness issues discussed in https://github.com/ash-rs/ash/pull/831#discussion_r1406631636 in a trivial way. I'd say we merge this and bump the MSRV, and also unblock the `no_std` PR.

Ralith:
SGTM. I don't think there's much benefit to a conservative MSRV here, especially if you don't want one.

Issue #635, opened 2022-06-05
Split out `*Fn` generated structures with dependencies on multiple extensions


MarijnS95:
Sparked from the discussion in #629:

`vk.xml` currently models which types, enums and commands (functions) a particular extension pulls in (provides?) by a `<require>` element inside `<extension>`. It also has support for nested extension and Vulkan version dependencies through `<require extension="XXX">` / `<require feature="XXX">`.

Currently our generator globs them all under the name of the top-level `<extension>` into a single loadable `struct`, whereas we'd like to preserve a strict feature/extension -> `struct` mapping. With #629/#631 the helpers are now also clobbered with functions that are only available when two extensions are available (mixed in with functions that are available when the "root" extension is enabled.

One of the options is to glue these names together, resulting in something potentially odd to read like `KhrSwapchainPlusVulkan11Fn`, but at least makes the dependency clear. This becomes more complicated for the functions added in #629 which have multiple mixed (disjoint) dependencies on for example `VK_KHR_swapchain + Vulkan 1.1`, but also `VK_KHR_device_group + VK_KHR_surface` or `VK_KHR_device_group + VK_KHR_swapchain`. Having two loader `structs` containing the same doesn't seem too outrageous here.

---

A list of all `<require extension="XXX">` blocks containing one or more `command` elements, generated on `v1.3.211` using:

```console
xmlstarlet sel -t -c "//extension/require[(@extension or @feature) and command]" generator/Vulkan-Headers/registry/vk.xml
```
```xml
<require feature="VK_VERSION_1_1">
    <comment>This duplicates definitions in VK_KHR_device_group below</comment>
    <enum extends="VkStructureType" extnumber="61" offset="7" name="VK_STRUCTURE_TYPE_DEVICE_GROUP_PRESENT_CAPABILITIES_KHR"/>
    <enum extends="VkStructureType" extnumber="61" offset="8" name="VK_STRUCTURE_TYPE_IMAGE_SWAPCHAIN_CREATE_INFO_KHR"/>
    <enum extends="VkStructureType" extnumber="61" offset="9" name="VK_STRUCTURE_TYPE_BIND_IMAGE_MEMORY_SWAPCHAIN_INFO_KHR"/>
    <enum extends="VkStructureType" extnumber="61" offset="10" name="VK_STRUCTURE_TYPE_ACQUIRE_NEXT_IMAGE_INFO_KHR"/>
[...]

PR #748, opened 2023-05-02
Allow building `Entry`/`Instance`/`Device` from handle+fns

pac85:
Adds a constructor to build a device from a handle + functions. 
Helps with interoperability with other vulkan wrappers where the user might have loaded the functions already.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_EXT_shader_object` device extension (#732)
 - Added missing `Device::get_device_queue2()` wrapper (#736)
 - Exposed `FramebufferCreateInfo::attachment_count()` builder for `vk::FramebufferCreateFlags::IMAGELESS` (#747)
+- Allow building `Entry`/`Instance`/`Device` from handle+fns (#748)
 
 ### Changed
 
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -23,13 +23,30 @@ impl Device {
             mem::transmute((instance_fn.get_device_proc_addr)(device, name.as_ptr()))
         };
 
-        Self {
-            handle: device,
+        Self::from_parts_1_3(
+            device,
+            vk::DeviceFnV1_0::load(load_fn),
+            vk::DeviceFnV1_1::load(load_fn),
+            vk::DeviceFnV1_2::load(load_fn),
+            vk::DeviceFnV1_3::load(load_fn),
+        )
+    }
 
-            device_fn_1_0: vk::DeviceFnV1_0::load(load_fn),
-            device_fn_1_1: vk::DeviceFnV1_1::load(load_fn),
-            device_fn_1_2: vk::DeviceFnV1_2::load(load_fn),
-            device_fn_1_3: vk::DeviceFnV1_3::load(load_fn),
+    #[inline]
+    pub fn from_parts_1_3(
+        handle: vk::Device,
+        device_fn_1_0: vk::DeviceFnV1_0,
+        device_fn_1_1: vk::DeviceFnV1_1,
+        device_fn_1_2: vk::DeviceFnV1_2,
+        device_fn_1_3: vk::DeviceFnV1_3,
+    ) -> Self {
+        Self {
+            handle,
+
+            device_fn_1_0,
+            device_fn_1_1,
+            device_fn_1_2,
+            device_fn_1_3,
         }
     }
 
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -140,15 +140,26 @@ impl Entry {
     /// `static_fn` must contain valid function pointers that comply with the semantics specified by
     /// Vulkan 1.0, which must remain valid for at least the lifetime of the returned [`Entry`].
     pub unsafe fn from_static_fn(static_fn: vk::StaticFn) -> Self {
-        let load_fn = |name: &std::ffi::CStr| {
+        let load_fn = move |name: &std::ffi::CStr| {
             mem::transmute((static_fn.get_instance_proc_addr)(
                 vk::Instance::null(),
                 name.as_ptr(),
             ))
         };
-        let entry_fn_1_0 = vk::EntryFnV1_0::load(load_fn);
-        let entry_fn_1_1 = vk::EntryFnV1_1::load(load_fn);
 
+        Self::from_parts_1_1(
+            static_fn,
+            vk::EntryFnV1_0::load(load_fn),
+            vk::EntryFnV1_1::load(load_fn),
+        )
+    }
+
+    #[inline]
+    pub fn from_parts_1_1(
+        static_fn: vk::StaticFn,
+        entry_fn_1_0: vk::EntryFnV1_0,
+        entry_fn_1_1: vk::EntryFnV1_1,
+    ) -> Self {
         Self {
             static_fn,
             entry_fn_1_0,
--- a/ash/src/instance.rs
+++ b/ash/src/instance.rs
@@ -22,12 +22,27 @@ impl Instance {
             mem::transmute((static_fn.get_instance_proc_addr)(instance, name.as_ptr()))
         };
 
-        Self {
-            handle: instance,
+        Self::from_parts_1_3(
+            instance,
+            vk::InstanceFnV1_0::load(load_fn),
+            vk::InstanceFnV1_1::load(load_fn),
+            vk::InstanceFnV1_3::load(load_fn),
+        )
+    }
 
-            instance_fn_1_0: vk::InstanceFnV1_0::load(load_fn),
-            instance_fn_1_1: vk::InstanceFnV1_1::load(load_fn),
-            instance_fn_1_3: vk::InstanceFnV1_3::load(load_fn),
+    #[inline]
+    pub fn from_parts_1_3(
+        handle: vk::Instance,
+        instance_fn_1_0: vk::InstanceFnV1_0,
+        instance_fn_1_1: vk::InstanceFnV1_1,
+        instance_fn_1_3: vk::InstanceFnV1_3,
+    ) -> Self {
+        Self {
+            handle,
+
+            instance_fn_1_0,
+            instance_fn_1_1,
+            instance_fn_1_3,
         }
     }
 


Ralith:
Another wrapper will have its own function pointer table structs, and converting those to ash's would be incredibly tedious and fragile. Why not just let ash re-load the function pointers as usual?

ash/src/device.rs
```diff
@@ -33,6 +33,23 @@ impl Device {
         }
     }
 
+    pub unsafe fn from_parts(
```
MarijnS95, line 36: In the event that we do end up going this way, does this function strictly need to be `unsafe`? Calling any function on `Device` is already `unsafe`, and we don't implement `Drop` semantics or other safe functionality that requires `handle: vk::Device` to be valid.

pac85:
> Another wrapper will have its own function pointer table structs, and converting those to ash's would be incredibly tedious and fragile. Why not just let ash re-load the function pointers as usual?

Vulkano uses ash under the hood and therefore the same function pointer table structs. I don't know if there are other examples. Loading twice would obviously work but doesn't seem like it should be the only choice to me.

MarijnS95:
Then Vulkano has `ash::Device` already, or should start using it and take out the function pointers via `fp_v1_x` functions on `ash::Device`? Otherwise, is there an upstream issue for Vulkano why they are owning the function-pointer tables themselves, and why this change on the `ash` side is needed to accommodate for something?

pac85:
> Then Vulkano has `ash::Device` already, or should start using it and take out the function pointers via `fp_v1_x` functions on `ash::Device`? Otherwise, is there an upstream issue for Vulkano why they are owning the function-pointer tables themselves, and why this change on the `ash` side is needed to accommodate for something?

They don't want to use ash::Device because they don't like that it takes slices from what I understand.
I don't think that's something that is going to change.

pac85:
> Then Vulkano has `ash::Device` already, or should start using it and take out the function pointers via `fp_v1_x` functions on `ash::Device`? Otherwise, is there an upstream issue for Vulkano why they are owning the function-pointer tables themselves, and why this change on the `ash` side is needed to accommodate for something?

Also the request for this functionality is not coming from them, it's coming from me, I need to use both crates because vulkano is removing some unsafe wrappers which I use. 

Thinking this is simething one might generally be interested in doing and also something that doesn't add maintenance burden I opened this PR.

pac85:
To put it another way, it seems like it makes sense that loading the driver shouldn't be the only way of using the wrappers.

Ralith:
> They don't want to use ash::Device because they don't like that it takes slices from what I understand.

Not liking the high level wrappers does not preclude using `Device` to load and represent the function pointer tables.

> Loading twice would obviously work but doesn't seem like it should be the only choice to me.

It has significant upsides. For example, it does not force you to use the same version of ash as Vulkano.

That said, so long as we expose `DeviceFn*::load`, it makes sense to allow someone who uses those interfaces to put them together into a `Device`, so I'm okay with adding this API.

Ralith:
Thoughts:
- [`Entry::from_static_fn`](https://docs.rs/ash/0.37.2+1.3.238/ash/struct.Entry.html#method.from_static_fn) already exists and is spiritually similar
- If we're offering this for `Device`, we should also add it for `Instance`.

pac85:
> Thoughts:
> 
>     * [`Entry::from_static_fn`](https://docs.rs/ash/0.37.2+1.3.238/ash/struct.Entry.html#method.from_static_fn) already exists and is spiritually similar
> 
>     * If we're offering this for `Device`, we should also add it for `Instance`.

As suggested I added an analogous constructor to `Instance`

Ralith (approved):

pac85:
> Hmm, one other consideration: these might force semver breakage when we support new minor versions of Vulkan. Do we otherwise necessarily break on such updates? If not, that makes this more costly. We could mitigate that by having a separate constructor for every minor version, in which case this should probably be renamed, though that feels rather ugly.

That's a good point, so would it be sufficient if I appended `_1_3` to the current function? Or maybe, even though there is no real need to do so, I should add previous versions as well? (I suppose for consistency?)

Ralith:
> would it be sufficient if I appended _1_3 to the current function?

I think that solves the problem. I'm interested in @MarijnS95's thoughts on this too.

> Or maybe, even though there is no real need to do so, should I add previous versions as well?

Let's not; we can always add them in the future. It could confuse users targeting earlier Vulkan versions, but it's an obscure enough interface that I'm not too bothered.

MarijnS95:
Vulkano doesn't use `ash` for external interoperability but (at least initially) to not have to manually generate and keep up with low-level sys-like bindings and the function tables.

I was under the impression that Vulkano might not expose their internals nor `ash` types to more strictly guarantee the extra safety that they aim to provide around Vulkan by utilizing Rust semantics. But it seems to give you access to these ([all function pointers including extensions](https://docs.rs/vulkano/latest/vulkano/device/struct.DeviceFunctions.html)) after all.

Perhaps we could ask @Rua to replace `ash::vk::Device` with `ash::Device` after all if this is intended.

---

Sure, we could provide this helper to be more lenient in how ash high-level wrappers are put together and used. After all, a user could pass a custom loader function with a massive `match` table that forwards already-loaded pointers to achieve the same, which is only cumbersome.

If you want to be consistent and provide it for `Device` _and `Instance`_, we should also provide it for `Entry`. `Entry::from_static_fn()` likewise only takes a loader callback which is used to load a few enumeration functions in `vk::EntryFnV1_0` and `vk::EntryFnV1_0`.

---

Then the semver-breakage: this is tricky. We cannot easily version the function as all fields _must_ be initialized (they are not `Option<>`al nor have a `Default`). The only thing we can do is:

1. Add newer (versioned) `from_parts` constructors (or version them from the get-go), marking any non-latest `fn` as `#[deprecated]` which is not breaking;
2. Initialize every newer version that isn't passed as argument (i.e. `from_parts_v1_3` for a hypothetical new `vk::DeviceFnV1_4`) with a no-op loader function that returns `NULL` for every function.

Alternatively we can preemptively make this function return an `Option`/`Result` and utilize that, which would be an API break on newer versions but not an ABI break. The former is more likable though.

---

@Ralith what's your opinion on the `from_parts` naming?

Ralith:
I think `from_parts` is an idiomatic name for this.

> The only thing we can do is ... version them from the get-go

That's what I had in mind, though I don't know if it's useful to deprecate older ones, since they'll continue to be reasonable to use when targeting older versions. Deprecating does let us remove them when we break for other reasons, which might be nice to limit growth, though I'm not sure if Vulkan is going to get *that* many minor versions, and we can always deprecate in the future.

pac85:
Added a `1_3` postfix

MarijnS95:
> That's what I had in mind, though I don't know if it's useful to deprecate older ones, since they'll continue to be reasonable to use when targeting older versions. Deprecating does let us remove them when we break for other reasons, which might be nice to limit growth, though I'm not sure if Vulkan is going to get _that_ many minor versions, and we can always deprecate in the future.

I'd `#[deprecate]` them to point out that certain functions are never going to be initialized and will always panic, even if it is now clear from the function name and arguments. It's a discussion point we have globally anyway for the current loader functions being infallible, even if it might be helpful to make them infallible to have a clearer overview of what is available (on a side-note: this is problematic when functions _within the same extension_ have different requirements: https://github.com/ash-rs/ash/issues/635).

MarijnS95:
Thanks @pac85. Looks like a few things remain before this is ready:

- [ ] Add the function to `Entry` as well (suggested above);
- [ ] List these additions in the changelog (copy-paste the title, with `(#748)` suffix at the bottom of the `### Added` list;
- [ ] Mark these functions as `#[inline]`.

pac85:
> Thanks @pac85. Looks like a few things remain before this is ready:
> 
>     * [ ]  Add the function to `Entry` as well (suggested above);
> 
>     * [ ]  List these additions in the changelog (copy-paste the title, with `(#748)` suffix) at the bottom of the `### Added` list;
> 
>     * [ ]  Mark these functions as `#[inline]`.

Those things should now be done and I've also rebased. 
Let me know if there is something else I need to do or if you don't like the commit messages.

Changelog.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_EXT_shader_object` device extension (#732)
 - Added missing `Device::get_device_queue2()` wrapper (#736)
 - Exposed `FramebufferCreateInfo::attachment_count()` builder for `vk::FramebufferCreateFlags::IMAGELESS` (#747)
+- Allow building Entry/Instance/Device from handle+fns (#748)
```
MarijnS95, line 17:
```suggestion
- Allow building `Entry`/`Instance`/`Device` from handle+fns (#748)
```

MarijnS95 (approved):
Looks good, thanks! I've been wondering if it makes sense to let e.g. `Entry::from_static_fn()` use the new function, but other than that I'll merge and backport this shortly.

MarijnS95:
> Let me know if there is something else I need to do or if you don't like the commit messages.

I generally try to (but not always) stick to a single commit, especially if all the messages would be the same, since it gets squash-merged in the end. Having a separate commit per entry/instance/device doesn't really add value here. But don't worry about squashing them, I can fix that up very easily while applying.

Sorry for the conflicts, we just removed a bunch of fields (with empty structs) in #752 :)

pac85:
> Looks good, thanks! I've been wondering if it makes sense to let e.g. `Entry::from_static_fn()` use the new function, but other than that I'll merge and backport this shortly.

Mmm so do I need to change it to use it?

MarijnS95:
I think that would look clean, yes.

pac85:
> I think that would look clean, yes.

And I suppose we should do the same with `Device` and `Instance`

pac85:
Now constructors use the `from_parts` frunctions and I've solved the conflicts

ash/src/entry.rs
```diff
@@ -149,6 +149,23 @@ impl Entry {
         let entry_fn_1_0 = vk::EntryFnV1_0::load(load_fn);
         let entry_fn_1_1 = vk::EntryFnV1_1::load(load_fn);
 
+        Self::from_parts_1_3(
+            static_fn,
+            entry_fn_1_0,
+            entry_fn_1_1,
+            entry_fn_1_2,
+            entry_fn_1_3,
```
MarijnS95, line 157: Perhaps compile-test?

ash/src/entry.rs
```diff
@@ -149,6 +149,23 @@ impl Entry {
         let entry_fn_1_0 = vk::EntryFnV1_0::load(load_fn);
         let entry_fn_1_1 = vk::EntryFnV1_1::load(load_fn);
```
MarijnS95, line 150: Inline these like in `Device`?

ash/src/entry.rs
```diff
@@ -149,6 +149,23 @@ impl Entry {
...
+        )
+    }
+
+    #[inline]
+    pub unsafe fn from_parts_1_3(
+        static_fn: vk::StaticFn,
+        entry_fn_1_0: vk::EntryFnV1_0,
+        entry_fn_1_1: vk::EntryFnV1_1,
+        entry_fn_1_2: vk::EntryFnV1_2,
+        entry_fn_1_3: vk::EntryFnV1_3,
```
MarijnS95, line 167: These are empty and unused arguments.

MarijnS95 (changes requested):

MarijnS95:
Thanks, looks like it still wouldn't compile though, but let me allow the CI workflow to be sure.

ash/src/entry.rs
Ralith: Hmm, should we name it `_1_1`, or rename it in the future even when it doesn't change? I lean towards the former.

ash/src/entry.rs
```diff
@@ -149,6 +149,23 @@ impl Entry {
...
+            static_fn,
+            entry_fn_1_0,
+            entry_fn_1_1,
+            entry_fn_1_2,
+            entry_fn_1_3,
+        )
+    }
+
+    #[inline]
+    pub unsafe fn from_parts_1_3(
```
Ralith, line 162: Stray `unsafe`?

ash/src/entry.rs
pac85: uh sorry I forgot after the rebase

ash/src/entry.rs
pac85: rust complains about lifetimes if I do

MarijnS95: We'd have to `Copy` `let get_instance_proc_addr = static_fn.get_instance_proc_addr;` in that case.

pac85:
> Thanks, looks like it still wouldn't compile though, but let me allow the CI workflow to be sure.

compilation fixed now

ash/src/entry.rs
pac85: remaed to `1_1`

ash/src/entry.rs
pac85: It was trivial to fix by making the closure `move`

Ralith (approved):
Thanks!

ash/src/entry.rs
```diff
@@ -140,15 +140,26 @@ impl Entry {
...
 
+        Self::from_parts_1_1(
+            static_fn,
+            vk::EntryFnV1_0::load(load_fn),
+            vk::EntryFnV1_1::load(load_fn),
+        )
+    }
+
+    #[inline]
+    pub unsafe fn from_parts_1_1(
```
MarijnS95, line 158:
```suggestion
    pub fn from_parts_1_1(
```

pac85: Mmm, `from_static_fn` is unsafe and this lets you do at least as much so why would it be safe?

MarijnS95: The same could be said for all other `from_parts_*` where we removed `unsafe`: these functions don't call anything unsafely, they just construct `Self`. Any other function on `Self` that invokes any of these functions is already marked `unsafe` for that reason.

pac85: Ah I see now.

MarijnS95 (approved):
Thanks, looks like we're all set!

pac85:
Fantastic! Thanks everyone!

ash/src/entry.rs
Ralith: > Any other function on Self that invokes any of these functions is already marked unsafe for that reason.

In `Entry`'s case that isn't actually true, which is why `from_static_fn` is unsafe. This should probably also be unsafe.

MarijnS95: Oh no! We also have `fn linked()` being safe, for obvious reasons, but then the following functions should have been marked `unsafe`:

- `try_enumerate_instance_version()`;
- `enumerate_instance_layer_properties()` (calls `entry_fn_1_0`);
- `enumerate_instance_extension_properties()` (^);
- `enumerate_instance_version()`.

I thought `from_static_fn()` would be `unsafe` because it calls a C function, not necessarily because it might "unsafely" initialize `Entry` hence not needing `unsafe` anymore on the functions listed above?

`create_instance()` and `get_instance_proc_addr()` are already `unsafe fn` too.

MarijnS95: With that I wouldn't trust the ICD/driver enough to not be `unsafe`, or is that why you made `fn linked()` safe?

So, all in all, what is your suggestion for the above? Make `Entry::from_parts_1_1()` `unsafe` again or make the methods on `Entry` unsafe?

Ralith: > With that I wouldn't trust the ICD/driver enough to not be `unsafe`

I'm not sure what "to not be `unsafe`" means here exactly. We must assume that they conform to the specifications, which means certain patterns of use are sound. When those patterns can be easily enforced statically, we can mark them as safe.

> is that why you made fn linked() safe?

`linked()` is safe because if you're replacing well known libraries with ones that behave incorrectly, you could just as well replace `libc` and break anything at all. The safety model doesn't extend arbitrarily ill-behaved environments.

> So, all in all, what is your suggestion for the above? Make Entry::from_parts_1_1() unsafe again or make the methods on Entry unsafe?

I'd be okay with either of these. I think I lean slightly towards the former, since it's a smaller change, the safe methods are pretty obviously safe, and you'd have to be doing something really dumb (supplying nonsensical function pointers) to make calling them unsafe. On the other hand, there's a case to be made that just making every wrapper unsafe is more consistent.

MarijnS95: > > With that I wouldn't trust the ICD/driver enough to not be `unsafe`
> 
> I'm not sure what "to not be `unsafe`" means here exactly. We must assume that they conform to the specifications, which means certain patterns of use are sound. When those patterns can be easily enforced statically, we can mark them as safe.

I hadn't expected Ash to validate that Vulkan's spec statements for these functions conform to Rust's notion of a "safe" function, and have thus far always been slapping `unsafe` on functions that are implemented outside of Rust and/or are nontrivial to validate the environment/arguments for.

> > is that why you made fn linked() safe?
> 
> `linked()` is safe because if you're replacing well known libraries with ones that behave incorrectly, you could just as well replace `libc` and break anything at all. The safety model doesn't extend to arbitrarily ill-behaved environments.

Okay.

> > So, all in all, what is your suggestion for the above? Make Entry::from_parts_1_1() unsafe again or make the methods on Entry unsafe?
> 
> I'd be okay with either of these. I think I lean slightly towards the former, since it's a smaller change, the safe methods are pretty obviously safe, and you'd have to be doing something really dumb (supplying nonsensical function pointers) to make calling them unsafe. On the other hand, there's a case to be made that just making every wrapper unsafe is more consistent.

Exactly, it feels odd to make the safe functions (in terms of implementation) `unsafe`, even though they allow other safe functions to behave `unsafe`ly when arbitrary pointers are stuffed into the function structs.

For consistency I'd also rather see the safe functions changed to `unsafe`. If we change `Entry::from_parts_1_1()` to be `unsafe` instead, we should do the same for `Instance` and `Device` in the event a safe function (that is influenced by such a constructor) is ever added.

Ralith: > and/or are nontrivial to validate the environment/arguments for.

Some functions are very straightforward! `unsafe` is the right default when there's any doubt, of course.

> it feels odd to make the safe functions (in terms of implementation) unsafe, even though they allow other safe functions to behave unsafely when arbitrary pointers are stuffed into the function structs.

This is a conventional use of `unsafe`. Safety contracts are easier to reason about when they don't span multiple items, but we pay that price either way here: either an API function is unsafe because maybe someone passed garbage to an `Entry` constructor in arbitrarily distant code, or the constructor is unsafe because maybe arbitrarily distant code will call a method after you pass garbage.

> For consistency I'd also rather see the safe functions changed to unsafe

Fine with me; consistency is important.

MarijnS95: > > and/or are nontrivial to validate the environment/arguments for.
> 
> Some functions are very straightforward! `unsafe` is the right default when there's any doubt, of course.

I don't think we can validate how straightforwardly they are implemented in a Vulkan driver? Maybe in the ICD though?

> This is a conventional use of `unsafe`. Safety contracts are easier to reason about when they don't span multiple items, but we pay that price either way here: either an API function is unsafe because maybe someone passed garbage to an `Entry` constructor in arbitrarily distant code, or the constructor is unsafe because maybe arbitrarily distant code will call a method after you pass garbage.

Exactly. But in this case I am concerned about the `unsafe`ness of calling that function even if it came from the driver/ICD. Otherwise we might have marked a whole bunch of functions (where we can trivially validate/restrict the argument range) safe.

> > For consistency I'd also rather see the safe functions changed to unsafe
> 
> Fine with me; consistency is important.

On the other hand it's `Entry` "inconsistently" marking some `extern "C"` (no matter how trivial) function wrappers as safe.

Ralith: > I don't think we can validate how straightforwardly they are implemented in a Vulkan driver?

The behavior dictated by the spec is the relevant concern. Many (most?) Vulkan functions specify complex soundness invariants, but some (like enumerating the instance version) are simple. Non-conformant drivers are intrinsically unsound just like a non-conformant libc would be.

> On the other hand it's Entry "inconsistently" marking some extern "C" (no matter how trivial) function wrappers as safe.

I'm not sure what you mean. I agree that marking methods on `Entry` as safe is appealing from a consistency standpoint, since ash as a matter of general policy does not attempt to reason about safety for you, though this doesn't have much to do with the presence of a FFI boundary per se.

MarijnS95: > I'm not sure what you mean. I agree that marking methods on `Entry` as unsafe is appealing from a consistency standpoint, since ash as a matter of general policy does not attempt to reason about safety for you, though this doesn't have much to do with the presence of a FFI boundary per se.

When observed from an FFI boundary, but you just said that that's irrelevant for `unsafe`ness when the soundness invariants of that function can be trivially upheld (and the function pointer is valid of course).

That would have otherwise been my reason to prefer keeping `from_parts` safe and marking the `"C"` functions unsafe, but I think you're right in that we should just mark `Entry::from_parts_1_1()` `unsafe` (explain why) and get this over with.

MarijnS95: Done now in https://github.com/ash-rs/ash/pull/756!

PR #752, opened 2023-05-06
Drop `load()` functions from empty `Fn` `struct`s

### LINKED ISSUE #442: Don't generate function structs for extensions that have no functions
I noticed that Ash generates quite a few structs for extensions without any functions, for example [`KhrSurfaceProtectedCapabilitiesFn`](https://docs.rs/ash/0.32.1/ash/vk/struct.KhrSurfaceProtectedCapabilitiesFn.html). These structs end up empty and are rather useless. Would it be better if the generator filtered out such cases?

MarijnS95:
Fixes #442, cc @Rua

These functions don't contribute anything and should be removed to clean up the `features` and `extensions` files, while now also not showing in the documentation anymore.  The structs remain in place for associated constants but are replaced with true unit-like structs (no curly brackets anymore), and unneeded `unsafe impl Send/Sync` are removed as well.

As these `load()` functions have been removed from the empty feature-levels on `Entry` and `Device` as well, rather than instantiating the unit structs and returning those the fields and `fp_vX_X()` getters have been removed entirely.


MarijnS95:
> As these `load()` functions have been removed from the empty feature-levels on `Entry` and `Device` as well, rather than instantiating the unit structs and returning those the fields and `fp_vX_X()` getters have been removed entirely.

We could also leave the empty/unit types in place for consistency, but I don't think it adds any value to keep them around.

Ralith (approved):

PR #753, opened 2023-05-06
extensions/khr/device_group_creation: Take borrow of `Entry` in `new()`

MarijnS95:
Looks like #630 typo'd the argument for `DeviceGroupCreation::new()` by unnecessarily requiring a move (`Clone`) of `Entry` just to call `get_instance_proc_addr()` on it.  Replace this with a borrow to match all other extensions.


Ralith (approved):
Good catch!

PR #754, opened 2023-05-06
[0.37-stable] extensions: Provide `new_from_instance()` fallback for `Instance` functions

MarijnS95:
This is a minimal, semver-compatible backport of #734 to the `0.37-stable` branch, warning Ash users of the problem outlined below while the issue is properly being solved in the next breaking Ash release (by separating `Instance` and `Device` functions in the generator to avert this problem entirely while also always providing optimal `Device`-specific functions for extension wrappers that are currently already loading _everything_ via `Instance` to forgo the problem).

As discovered and detailed in #727 a few extension wrappers were loading and calling `Instance` functions via `Device` and `get_device_proc_addr()` which is [defined] to only return non-`NULL` function pointers for `Device` functions.  Those wrapper functions will always call into Ash's panicking NULL-stub functions as the desired `Instance` function could not be loaded.

Deprecate the `new()` functions for extension wrappers that were doing this, while pointing the reader to `new_from_instance()` and explaining in the docs what function will always `panic!()` when the struct was loaded using `new()` instead.

This function always takes a raw `vk::Device` directly to fill `handle` (rather than `ash::Device` to retrieve `handle()` from), allowing users to pass `vk::Device::null()` when they do intend to load this extension wrapper just for calling the `Instance` function.

[defined]: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetDeviceProcAddr.html#_description

### DIFF
--- a/ash/src/extensions/khr/device_group_creation.rs
+++ b/ash/src/extensions/khr/device_group_creation.rs
@@ -59,8 +59,14 @@ impl DeviceGroupCreation {
         &self.fp
     }
 
+    #[deprecated = "typo: this function is called `device()`, but returns an `Instance`."]
     #[inline]
     pub fn device(&self) -> vk::Instance {
+        self.handle
+    }
+
+    #[inline]
+    pub fn instance(&self) -> vk::Instance {
         self.handle
     }
 }
--- a/ash/src/extensions/khr/swapchain.rs
+++ b/ash/src/extensions/khr/swapchain.rs
@@ -3,10 +3,11 @@ use super::DeviceGroup;
 use crate::prelude::*;
 use crate::vk;
 use crate::RawPtr;
-use crate::{Device, Instance};
+use crate::{Device, Entry, Instance};
 use std::ffi::CStr;
 use std::mem;
 
+/// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_KHR_swapchain.html>
 #[derive(Clone)]
 pub struct Swapchain {
     handle: vk::Device,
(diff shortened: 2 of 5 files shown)


ash/src/extensions/khr/swapchain.rs
```diff
@@ -3,17 +3,25 @@ use super::DeviceGroup;
...
 }
 
 impl Swapchain {
+    /// [`Instance`] functions cannot be loaded from a [`Device`] and will always panic when called:
+    /// - [`Self::get_physical_device_present_rectangles()`]
+    ///
+    /// Load this struct using an [`Instance`] instead via [`Self::new_from_instance()`], or
+    /// suppress the deprecation warning if the above [`Instance`] function is not called.
+    /// This will be solved in the next breaking `ash` release.
+    #[deprecated = "`Instance` functions loaded via this function will always panic. Load via `new_from_instance()` instead."]
```
MarijnS95, line 24: I'm afraid that this deprecation warning comes over too aggressively on downstream consumers when we release `0.37.4`, pretty much everyone using Vulkan will be using a `swapchain` and have their project fail linting.

On the other hand I really want to make users aware of this deficiency, though they'll likely run into the panic anyway whenever testing the codepath of the instance function (if anyone calls into it at all - we received _zero_ reports about these) and hopefully find this elaborate explanation to understand the issue (or we can guide them in case an issue gets created). It'll anyway be solved properly in `0.38`, which I hope to release "soon™" after processing all the known breaking changes and actually finalizing what the proper solution (#734) is going to look like!

Ralith: If this is, as the diff suggests, only an issue for `get_physical_device_present_rectangles`, maybe it's not that important to call out? That function seems *very* niche.

Ralith (approved):

ash/src/extensions/khr/device_group_creation.rs
```diff
@@ -13,7 +13,13 @@ pub struct DeviceGroupCreation {
 }
 
 impl DeviceGroupCreation {
+    /// Call [`Self::new_()`] with a borrow of [`Entry`] instead.
+    #[deprecated = "Unnecessarily requires a move of `Entry`. Call `new_()` with a borrow instead."]
     pub fn new(entry: Entry, instance: &Instance) -> Self {
+        Self::new_(&entry, instance)
+    }
+
+    pub fn new_(entry: &Entry, instance: &Instance) -> Self {
```
Ralith, line 22: Is this necessary? Cloning the `Entry` is dumb, but pretty harmless.

MarijnS95: We'll fix it in 0.38 (#753) so yes, maybe it is annoying to force people to switch to `new_()` and then back again later. I'll drop this change, `Entry` isn't all that large though and this extension likely isn't very popular.

ash/src/extensions/khr/swapchain.rs
MarijnS95: `get_device_group_surface_present_modes2` seems very niche as well, so I'll drop `#[deprecated]` on all these.

Ralith (approved):

PR #756, opened 2023-05-26
entry: Mark `Entry::from_parts_1_1()` as `unsafe`

MarijnS95:
We don't mark any of the extern calls to Vulkan function-pointers safe except for a few `Entry` functions.  Their safety contract was easy to validate, but now that we exposed a constructor function to let the user provide function pointers in the form of `Entry::from_parts_1_1()` it is no longer safe to assume that we are calling the function that adheres to the contract specified in the Vulkan reference.

Because we don't rigorously do this validation and safe-marking anywhere else either, mark these function calls as `unsafe`.

Related discussion chain: https://github.com/ash-rs/ash/pull/748#discussion_r1186794284

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_EXT_shader_object` device extension (#732)
 - Added missing `Device::get_device_queue2()` wrapper (#736)
 - Exposed `FramebufferCreateInfo::attachment_count()` builder for `vk::FramebufferCreateFlags::IMAGELESS` (#747)
-- Allow building `Entry`/`Instance`/`Device` from handle+fns (#748)
+- Allow building `Entry`/`Instance`/`Device` from handle+fns (see their `from_parts_1_x()` associated functions) (#748)
 
 ### Changed
 
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -184,7 +184,7 @@ impl Entry {
     /// # use ash::{Entry, vk};
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
     /// let entry = Entry::linked();
-    /// match entry.try_enumerate_instance_version()? {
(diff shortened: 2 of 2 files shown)


Ralith (approved):
Did you want to also mark the individual methods as unsafe for consistency?

MarijnS95:
> Did you want to also mark the individual methods as unsafe for consistency?

We were not going to do that because the safety contract could be upheld on the Vulkan side, right? The only problem was that we can no longer validate this when the _function pointer itself_ originates from an unknown source (e.g. not `vkGetInstanceProcAddr`).

Otherwise I would have gone for marking the functions as `unsafe` instead (to match all other Vulkan functions) and leave this function as safe (to match `from_parts` on `Instance` and `Device`) but we discussed the opposite in https://github.com/ash-rs/ash/pull/748#discussion_r1186794284?

Ralith:
Both approaches are equally valid from a correctness standpoint. I understood from that conversation that you strongly favored the approach of marking the wrapper functions as unsafe for consistency?

MarijnS95:
@Ralith yes they are, but since you put so much emphasis on (extern) C functions being not necessarily/inherently `unsafe` to call, I figured the preference was to mark `from_parts` unsafe?

I haven't checked any of the safety contracts for these C functions, so we might as well mark them `unsafe` and keep the `from_parts` function safe (both matching how `ash` exposes the API everywhere else) then?

Ralith:
> you put so much emphasis on (extern) C functions being not necessarily/inherently unsafe to call

This was a point of fact, but is not necessarily the deciding factor. Apologies for the confusion!

> I haven't checked any of the safety contracts for these C functions, so we might as well mark them unsafe and keep the from_parts function safe (both matching how ash exposes the API everywhere else) then?

I find this appealing from a consistency standpoint. Even though we can easily deem these safe, it might be confusing to do so when we in general don't.

MarijnS95:
Ah, that makes sense then! Sure, I'll mark the C function wrappers `unsafe` instead.

Ralith (approved):

PR #762, opened 2023-06-21
Add MSFT vendor tag for enum variants

jenatali:
Otherwise enums with the MSFT suffix fail to parse.


jenatali:
See also: https://github.com/KhronosGroup/Vulkan-Docs/pull/2129

MarijnS95 (approved):

MarijnS95:
Thanks!

PR #765, opened 2023-07-05
extensions/google: Add VK_GOOGLE_display_timing (#765)

spencercw:


ash/src/extensions/google/display_timing.rs
```diff
@@ -0,0 +1,51 @@
...
+            .result_with_success(properties)
+    }
+
+    pub const NAME: &'static CStr = vk::GoogleDisplayTimingFn::NAME;
+
+    #[inline]
+    pub fn fp(&self) -> &vk::GoogleDisplayTimingFn {
+        &self.fp
+    }
+}
```
MarijnS95, line 51: Whenever there's a handle stored inside an extension, we add a getter to take it out:

```suggestion
    }
    
    #[inline]
    pub fn device(&self) -> vk::Device {
        self.handle
    }
}
```

MarijnS95 (approved):
Thanks! One small nit :)

ash/src/extensions/google/display_timing.rs
spencercw: Ah I missed that. Change applied.

PR #774, opened 2023-07-21
Added ash_window::get_present_support

Rob2309:
This PR introduces `ash_window::get_present_support`, which is a higher-level wrapper around the various platform-specific functions for querying whether a physical device queue family supports presenting to surfaces (see [the vulkan spec](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_android)).

This allows choosing a physical device before creating a surface.

### DIFF
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
 
     Ok(extensions)
 }
+
+/// Query whether a `queue_family` of the given `physical_device` supports presenting to any surface that might be created.
+/// This function can be used to find a suitable [`vk::PhysicalDevice`] and queue family
+/// for rendering before a single surface is created.
+///
+/// This function can be a more useful alternative for
+/// [`VkGetPhysicalDeviceSurfaceSupportKHR`](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#vkGetPhysicalDeviceSurfaceSupportKHR),
+/// which requires having an actual surface available before choosing a physical device.
+///
+/// For more information see [the vulkan spec on WSI integration](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#_querying_for_wsi_support).
+pub fn get_present_support(
+    entry: &Entry,
+    instance: &Instance,
+    physical_device: vk::PhysicalDevice,
+    queue_family_index: u32,
+    display_handle: RawDisplayHandle,
+) -> VkResult<bool> {
+    match display_handle {
+        RawDisplayHandle::Android(_) | RawDisplayHandle::UiKit(_) | RawDisplayHandle::AppKit(_) => {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_android
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_ios
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_macos
+            // On Android, iOS and macOS, every queue family supports presenting to any surface
+            Ok(true)
+        }
+        RawDisplayHandle::Wayland(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_walyand
+            let ext = khr::WaylandSurface::new(entry, instance);
+            Ok(ext.get_physical_device_wayland_presentation_support(
+                physical_device,
+                queue_family_index,
+                &mut *h.display.cast(),
+            ))
+        },
+        RawDisplayHandle::Windows(_) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_win32
+            let ext = khr::Win32Surface::new(entry, instance);
+            Ok(ext.get_physical_device_win32_presentation_support(
+                physical_device,
+                queue_family_index,
+            ))
+        },
+        RawDisplayHandle::Xcb(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_xcb
+            let ext = khr::XcbSurface::new(entry, instance);
+            Ok(ext.get_physical_device_xcb_presentation_support(
+                physical_device,
+                queue_family_index,
+                &mut *h.connection,
+                h.screen as _,
+            ))
+        },
+        RawDisplayHandle::Xlib(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_xlib
+            let ext = khr::XlibSurface::new(entry, instance);
+            Ok(ext.get_physical_device_xlib_presentation_support(
+                physical_device,
+                queue_family_index,
+                h.display,
+                h.screen as _,
+            ))
+        },
+        // All other platforms mentioned in the vulkan spec are not supported by ash.
+        _ => Err(vk::Result::ERROR_EXTENSION_NOT_PRESENT),
+    }
+}


ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
...
+            ))
+        },
+        RawDisplayHandle::Xcb(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_xcb
+            let ext = khr::XcbSurface::new(entry, instance);
+            Ok(ext.get_physical_device_xcb_presentation_support(
+                physical_device,
+                queue_family_index,
+                &mut *h.connection,
+                h.screen as _,
```
Ralith, line 223: This is supposed to be a visual, not a screen.

ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
...
+            ))
+        },
+        RawDisplayHandle::Xlib(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_xlib
+            let ext = khr::XlibSurface::new(entry, instance);
+            Ok(ext.get_physical_device_xlib_presentation_support(
+                physical_device,
+                queue_family_index,
+                h.display,
+                h.screen as _,
```
Ralith, line 233: As above.

Ralith (changes requested):
Thanks, this looks really useful!

ash-window/src/lib.rs
Rob2309: You're right, I think a visualID can be somehow derived from the screen, but I think this would require linking to some X library. As an alternative, the get_present_support function could require a window handle, but this would defeat the main use case.

Ralith: Yeah. We could pull in `x11-dl` and use it to make the necessary call (`XDefaultVisualOfScreen` I guess?). Bonus points for adding a cargo feature to use `x11` instead, as dynamic linking is a bit tidier than runtime loading on Linux, but it can be nice to avoid a buildtime dep on a foreign lib.

Alternatively, maybe a more convenient approach would be adding a visual_id member to  to `raw_window_handle::XlibDisplayHandle` upstream (and making sure winit populates it). This is the same motivation behind its inclusion in the window handle struct, though I'm not enough of an X11 expert to judge how weird that might be.

ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
 
     Ok(extensions)
 }
+
+/// Query whether a `queue_family` of the given `physical_device` supports presenting to any surface that might be created.
+/// This function can be used to find a suitable [`vk::PhysicalDevice`] and queue family
+/// for rendering before a single surface is created.
```
MarijnS95, line 177: Nit: Can you reflow these sentences to both be approx the same length (line length is 100 characters): https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#max_width

ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
...
+///
+/// This function can be a more useful alternative for
+/// [`VkGetPhysicalDeviceSurfaceSupportKHR`](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#vkGetPhysicalDeviceSurfaceSupportKHR),
```
MarijnS95, line 180: 1. This is a function, and starts with a _lowercase_ `v`;
2. It is preferred to have (long) links at the bottom, so that the source text isn't too crowded and oddly flowed (``[`vkGetPhysicalDeviceSurfaceSupportKHR`]``) automatically crates a label that you can assign a link to via ``/// [`vkGetPhysicalDeviceSurfaceSupportKHR`]: https://...``
   (More important below);
3. Since Ash is a dependency (and in scope via `ash::extensions::khr`), prefer linking to the Rust implementation of this function via ``[`khr::Surface::get_physical_device_surface_support()`]``.

(And separately link the docs page _if_ there's more useful information to describe explicitly)

ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
...
+/// which requires having an actual surface available before choosing a physical device.
+///
+/// For more information see [the vulkan spec on WSI integration](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#_querying_for_wsi_support).
```
MarijnS95, line 183:
```suggestion
/// For more information see [the Vulkan spec on WSI integration][_querying_for_wsi_support].
///
/// [_querying_for_wsi_support]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#_querying_for_wsi_support
```

ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
...
+            // On Android, iOS and macOS, every queue family supports presenting to any surface
+            Ok(true)
+        }
+        RawDisplayHandle::Wayland(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_walyand
+            let ext = khr::WaylandSurface::new(entry, instance);
+            Ok(ext.get_physical_device_wayland_presentation_support(
+                physical_device,
+                queue_family_index,
+                &mut *h.display.cast(),
```
MarijnS95, line 205: This doesn't look correct, this function takes `&mut vk::wl_display` (which is a type alias to `c_void`) but I think it should take `*mut vk::wl_display`; should we make a breaking change to improve that?

ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
...
+                queue_family_index,
+            ))
+        },
+        RawDisplayHandle::Xcb(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_xcb
+            let ext = khr::XcbSurface::new(entry, instance);
+            Ok(ext.get_physical_device_xcb_presentation_support(
+                physical_device,
+                queue_family_index,
+                &mut *h.connection,
```
MarijnS95, line 222: Same here, I think we should take `*mut vk::xcb_connection_t`.

ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
...
+                h.screen as _,
+            ))
+        },
+        RawDisplayHandle::Xlib(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_xlib
+            let ext = khr::XlibSurface::new(entry, instance);
+            Ok(ext.get_physical_device_xlib_presentation_support(
+                physical_device,
+                queue_family_index,
+                h.display,
```
MarijnS95, line 232: I.e. this is a `*mut vk::Display`, not a `&mut `vk::Display` :grimacing:

ash-window/src/lib.rs
```diff
@@ -171,3 +171,69 @@ pub fn enumerate_required_extensions(
...
+                h.screen as _,
+            ))
+        },
+        // All other platforms mentioned in the vulkan spec are not supported by ash.
```
MarijnS95, line 236:
```suggestion
        // All other platforms mentioned in the Vulkan spec are not supported by ash-window.
```

Because `ash` maps the entire Vulkan API (== supports everything), even though we might not have written a slightly-higher-level `Fn` wrapper for it yet the lower-level API is still available.

Though I would reword that to `... don't currently have an implementation in ash-window`, as we can always expand.

MarijnS95 (changes requested):

MarijnS95:
While reviewing and refactoring the PR message, should we also mention (in the PR/commit message _and the doc-comment_) that it checks for support on the given `Display` connection? I'm not sure how much of a difference it makes beyond allowing the Vulkan driver to connect to the WSI though (perhaps for Xorg `Display` handles that could be networked?).

Rob2309:
I think I addressed everything above in this new commit.
I added x11 and x11-dl as optional dependencies, x11-dl being enabled by default. There does not seem to be a xcb-dl lib so I added xcb as a mandatory dependency for now.
The xlib code path is working on my wsl2 installation, as for xcb I am not sure, winit seems to not actually support xcb as far as I can tell.
I also changed the parameters of `get_physical_device_xcb_presentation_support` and `get_physical_device_wayland_presentation_support` to be pointers instead of references.

Rob2309:
I just noticed that `--no-default-features` seems to be ignored for running the winit example, I am not sure if I made a mistake or this is a cargo quirk.

ash/src/extensions/khr/wayland_surface.rs
```diff
@@ -43,7 +43,7 @@ impl WaylandSurface {
         &self,
         physical_device: vk::PhysicalDevice,
         queue_family_index: u32,
-        wl_display: &mut vk::wl_display,
+        wl_display: *mut vk::wl_display,
```
MarijnS95, line 46: We should write these down as breaking changes in the `CHANGELOG` (maybe via a separate PR so that it shows up properly in the commit history), though if someone passes exactly a `&mut vk::wl_display` it'll automatically coerce to the raw pointer.

ash-window/src/lib.rs
```diff
@@ -171,3 +171,102 @@ pub fn enumerate_required_extensions(
 
     Ok(extensions)
 }
+
+/// Query whether a `queue_family` of the given `physical_device` supports presenting to any
+/// surface that might be created. This function can be used to find a suitable
+/// [`vk::PhysicalDevice`] and queue family for rendering before a single surface is created.
+///
+/// This function can be a more useful alternative for [`vkGetPhysicalDeviceSurfaceSupportKHR`],
```
MarijnS95, line 179: I meant to just have:
```suggestion
/// This function can be a more useful alternative for [`khr::Surface::get_physical_device_surface_support()`],
```

We don't generally use Vulkan symbol naming in Ash.

ash-window/Cargo.toml
```diff
@@ -18,13 +18,21 @@ rust-version = "1.64.0"
...
+
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 raw-window-metal = "0.3"
 
 [dev-dependencies]
 winit = "0.28.0"
 ash = { path = "../ash", version = "0.37", default-features = false, features = ["linked"] }
 
+[features]
+default = [ "x11-dl" ]
```
MarijnS95, line 34: Let's stick with the same formatting that the rest of this file uses:


```suggestion
default = ["x11-dl"]
```

ash-window/Cargo.toml
```diff
@@ -18,13 +18,21 @@ rust-version = "1.64.0"
 ash = { path = "../ash", version = "0.37", default-features = false }
 raw-window-handle = "0.5"
 
+[target.'cfg(unix)'.dependencies]
+x11-dl = { version="2.21.0", optional=true }
+x11 = { version="2.21.0", features=["xlib"], optional=true }
```
MarijnS95, line 23: Let's stick with the same formatting that the rest of this file uses:
```suggestion
x11-dl = { version = "2.21.0", optional = true }
x11 = { version = "2.21.0", features = ["xlib"], optional = true }
```

ash-window/src/lib.rs
```diff
@@ -171,3 +171,102 @@ pub fn enumerate_required_extensions(
...
+        #[cfg(unix)]
+        RawDisplayHandle::Xcb(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_xcb
+            let ext = khr::XcbSurface::new(entry, instance);
+
+            let xcb = xcb::Connection::from_raw_conn(h.connection.cast());
+            let setup = xcb.get_setup();
+            let screen = setup.roots().nth(h.screen as usize).unwrap();
+            let visual = screen.root_visual();
+            xcb.into_raw_conn();
```
MarijnS95, line 227: For cleanliness, can we write `let connection = ...` here and pass that into `get_physical_device_xcb_presentation_support()` instead?

ash-window/src/lib.rs
```diff
@@ -171,3 +171,102 @@ pub fn enumerate_required_extensions(
...
+        },
+        RawDisplayHandle::Windows(_) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_win32
+            let ext = khr::Win32Surface::new(entry, instance);
+            Ok(ext.get_physical_device_win32_presentation_support(
+                physical_device,
+                queue_family_index,
+            ))
+        },
+        #[cfg(unix)]
```
MarijnS95, line 218:
```suggestion
        #[cfg(feature = "xcb")]
```

ash-window/src/lib.rs
```diff
@@ -171,3 +171,102 @@ pub fn enumerate_required_extensions(
...
+            let res = ext.get_physical_device_xcb_presentation_support(
+                physical_device,
+                queue_family_index,
+                h.connection,
+                visual,
+            );
+
+            Ok(res)
+        },
+        #[cfg(unix)]
```
MarijnS95, line 238:
```suggestion
        #[cfg(any(feature = "x11-dl", feature = "x11"))]
```

ash-window/src/lib.rs
```diff
@@ -171,3 +171,102 @@ pub fn enumerate_required_extensions(
...
+        RawDisplayHandle::Xlib(h) => unsafe {
+            // https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap34.html#platformQuerySupport_xlib
+            let ext = khr::XlibSurface::new(entry, instance);
+
+            #[cfg(not(any(feature = "x11-dl", feature = "x11")))]
+            compile_error!("Exactly one of x11-dl or x11 must be enabled on unix");
+            #[cfg(all(feature = "x11-dl", feature = "x11"))]
+            compile_error!("Exactly one of x11-dl or x11 must be enabled on unix");
```
MarijnS95, line 246: Features should be additive, so if we can just pick one of the two crates instead of forcing them to be mutually exclusive?

ash-window/src/lib.rs
MarijnS95: > Bonus points for adding a cargo feature to use `x11` instead, as dynamic linking is a bit tidier than runtime loading on Linux, but it can be nice to avoid a buildtime dep on a foreign lib.

Quickly scrolling through the changes, it seems that this feature is now forcibly requiring either `x11` or `x11-dl`, or a compiler error will occur on _all of `unix`_. Same for `xcb` now being forced on all consumers of `ash-window`.

The `cfg` guards on those `RawDisplayHandle`s should be refactored to contain the optional libraries as features, otherwise `get_present_support()` won't provide support for x11/xcb but the other functions are not affected.

Then, the module documentation and function documentation should be extended to explain these features.

ash-window/src/lib.rs
```diff
@@ -171,3 +171,102 @@ pub fn enumerate_required_extensions(
...
 }
+
+/// Query whether a `queue_family` of the given `physical_device` supports presenting to any
+/// surface that might be created. This function can be used to find a suitable
+/// [`vk::PhysicalDevice`] and queue family for rendering before a single surface is created.
+///
+/// This function can be a more useful alternative for [`vkGetPhysicalDeviceSurfaceSupportKHR`],
+/// which requires having an actual surface available before choosing a physical device.
+///
+/// For more information see [the vulkan spec on WSI integration][_querying_for_wsi_support].
```
MarijnS95, line 182: Once again:


```suggestion
/// For more information see [the Vulkan spec on WSI integration][_querying_for_wsi_support].
```

ash-window/src/lib.rs
```diff
@@ -171,3 +171,102 @@ pub fn enumerate_required_extensions(
...
+            }
+
+            Ok(ext.get_physical_device_xlib_presentation_support(
+                physical_device,
+                queue_family_index,
+                h.display,
+                visual_id as _,
+            ))
+        },
+        // All other platforms mentioned in the vulkan spec don't
```
MarijnS95, line 268: https://github.com/ash-rs/ash/pull/774#discussion_r1377297481

```suggestion
        // All other platforms mentioned in the Vulkan spec don't
```

MarijnS95:
> I just noticed that `--no-default-features` seems to be ignored for running the winit example, I am not sure if I made a mistake or this is a cargo quirk.

How so? If I run this on your branch, which disables both `x11` and `x11-dl` (which I do not care about when using Wayland), the creates are disables as expected:

```
$ cargo r -p ash-window --example winit --no-default-features
...
error: Exactly one of x11-dl or x11 must be enabled on unix
   --> ash-window/src/lib.rs:244:13
    |
244 |             compile_error!("Exactly one of x11-dl or x11 must be enabled on unix");
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Though as said above, I don't think we should make this a compile error.

Rob2309:
I now made every feature optional, but made x11-dl and xcb the default, so that every platform just works without enabling any features manually. A possible problem I see is that when x11/x11-dl is not enabled, the function will just return a generic EXTENSION_NOT_PRESENT error, which might be confusing. The same goes for xcb.

Rob2309:
> > I just noticed that `--no-default-features` seems to be ignored for running the winit example, I am not sure if I made a mistake or this is a cargo quirk.
> 
> How so? If I run this on your branch, which disables both `x11` and `x11-dl` (which I do not care about when using Wayland), the creates are disables as expected:
> 
> ```
> $ cargo r -p ash-window --example winit --no-default-features
> ...
> error: Exactly one of x11-dl or x11 must be enabled on unix
>    --> ash-window/src/lib.rs:244:13
>     |
> 244 |             compile_error!("Exactly one of x11-dl or x11 must be enabled on unix");
>     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> Though as said above, I don't think we should make this a compile error.

I just found out that cargo seems to ignore `--no-default-features` when not explicitly specifying a package. So `cargo run --example winit --no-default-features` still uses default features, but `cargo run -p ash-window --example winit --no-default-features` does not... Weird cargo quirk I guess

MarijnS95:
> I just found out that cargo seems to ignore `--no-default-features` when not explicitly specifying a package. So `cargo run --example winit --no-default-features` still uses default features, but `cargo run -p ash-window --example winit --no-default-features` does not... Weird cargo quirk I guess

I didn't expect `cargo run --example` to collect all examples from all members in the workspace automatically. TIL :)

MarijnS95:
Cool stuff! Can we extend the example to showcase this API?

I hacked in this chunk to see the _type_ of window handle and print the result, but we should probably come up with something a bit nicer.

```rust
        dbg!(window.raw_display_handle());

        let &pdev = instance.enumerate_physical_devices()?.first().unwrap();

        let support = ash_window::get_present_support(
            &entry,
            &instance,
            pdev,
            0,
            window.raw_display_handle(),
        );
        dbg!(support);
```

And on that note, Rust naming says to not use `get_` prefixes; can we come up with a better name? `has_present_support(for_display_handle)()` perhaps?

Rob2309:
> And on that note, Rust naming says to not use `get_` prefixes; can we come up with a better name? `has_present_support(for_display_handle)()` perhaps?

I would probably prefer keeping the get_prefix, as all the platform specific functions used also start with a get prefix. Ash itself also leaves the get_ prefix in place for many functions. But that is just my opinion

MarijnS95:
> winit seems to not actually support xcb as far as I can tell.

I just ran into https://github.com/rust-windowing/winit/issues/3198

Rob2309:
I added a simple loop to the example that prints info about every physical device.

Rob2309:
> > winit seems to not actually support xcb as far as I can tell.
> 
> I just ran into [rust-windowing/winit#3198](https://github.com/rust-windowing/winit/issues/3198)

Since I can't get winit to use xcb, I might test the xcb code by creating a window with the xcb crate directly. Might take some time to get it right though...

Rob2309:
I now added a basic xcb example, the xcb code seems to be working just fine. Not sure if the example should be kept though...

MarijnS95:
> Since I can't get winit to use xcb, I might test the xcb code by creating a window with the xcb crate directly. Might take some time to get it right though...

No need to, it's supposed to a simple `winit` example. We can wait for them to implement `xcb`.

ash-window/examples/xcb.rs
```diff
@@ -0,0 +1,92 @@
...
+            for i in 0..queue_families.len() {
+                let present_support = ash_window::get_present_support(
+                    &entry,
+                    &instance,
+                    dev,
+                    i as _,
+                    display_handle,
+                )?;
+                println!(
+                    "{dev_name}, queue {i} {} presenting to the surface",
```
MarijnS95, line 43: Nit: "a" surface? There's no "the surface" to refer to here yet?

Rob2309: Whoops it seems like I forgot that the entire purpose is to check before a surface exists :D

Rob2309:
> > Since I can't get winit to use xcb, I might test the xcb code by creating a window with the xcb crate directly. Might take some time to get it right though...
> 
> No need to, it's supposed to a simple `winit` example. We can wait for them to implement `xcb`.

In that case, should I remove the example again? I used it to verify that the xcb code works, but I don't think it adds any value that the winit example does not.

MarijnS95:
Hmm you have already written it, but it also adds some extra maintenance burden to keep it up-to-date and functional.

ash-window/Cargo.toml
```diff
@@ -18,13 +18,26 @@ rust-version = "1.64.0"
...
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 raw-window-metal = "0.3"
 
 [dev-dependencies]
 winit = "0.28.0"
 ash = { path = "../ash", version = "0.37", default-features = false, features = ["linked"] }
+xcb = { version = "1.2.2" }
+
+[features]
+default = ["x11-dl", "xcb"]
```
Ralith, line 35: Requiring `xcb` by default means we need libxcb present at buildtime, right? While probably most environments will have this, disabling it will make it easier for downstream users to get CI working in minimal VMs and such.

Ralith (approved):
Implementation LGTM overall, thanks!

MarijnS95 (approved):
Looks like I forgot to submit this outstanding review. Don't forget to address https://github.com/ash-rs/ash/pull/774/files#r1378560481 :)

ash-window/Cargo.toml
Rob2309: The reason for defaulting this is to have the least surprising behavior by default. I dont think the default should be to not support a very broadly used platform.

ash-window/Cargo.toml
```diff
@@ -18,13 +18,26 @@ rust-version = "1.64.0"
...
 
 [[example]]
 name = "winit"
 required-features = ["ash/linked"]
+
+[[example]]
+name = "xcb"
+required-features = ["ash/linked"]
```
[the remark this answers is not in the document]
Rob2309: I am not quite sure whether normal dependencies are available in examples. Will test this and change if possible.

Rob2309:
> Looks like I forgot to submit this outstanding review. Don't forget to address https://github.com/ash-rs/ash/pull/774/files#r1378560481 :)

I can add this to the changelog, I was just hesitant because of the separate PR point.

ash-window/Cargo.toml
Ralith: We know winit doesn't use xcb. Where in the raw-window-handle ecosystem is xcb broadly used?

Rob2309:
The forced dev-dependency is now removed, it seems to just work with the regular optional dependency.
I also added entries to both the ash-window and ash changelog reflecting the changes here. Is there any special marking that the breaking changes should have?

ash-window/Cargo.toml
Rob2309: Good point, however if winit at some point decides to switch to xcb, it should be defaulted again...

Ralith: I think winit will want to avoid adding foreign build-time dependencies for the same reason.

Rob2309:
Is there anything left to do here? I must admit I am not quite sure about the status of this PR anymore...

Ralith:
Triage: This is still desirable, and is nearly ready. I've gone through and resolved threads which have been addressed.

Remaining concerns:
- enabling `xcb` by default on Unix platforms may be a poor experience, especially on e.g. macos. Also possibly not desired in Wayland-only environments.
- Needs a rebase.

PR #775, opened 2023-07-26
Add ash_window::get_present_support to 0.37-stable

Rob2309:
This fulfills the same purpose as #774, but for the 0.37-stable branch, which is probably the more important one...

### DIFF
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -175,3 +175,69 @@ pub fn enumerate_required_extensions(
 
     Ok(extensions)
 }
+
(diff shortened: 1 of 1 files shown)


MarijnS95:
This is not a branch that is used as target for PRs. Things get merged to `master` and non-breaking changes get cherry-picked to the backport branch when relevant.

I'm no longer planning to make an incremental non-breaking `0.37.x` release in favour of focusing and finishing the last remaining breaking change and get `0.38.0` out.

Note also that `ash-window` has quite a few PRs with contentious API additions open, meaning throughput on that crate has mostly stalled.

PR #779, opened 2023-07-31
extensions/ext: Add VK_EXT_host_image_copy

MarijnS95:


ash/src/extensions/ext/image_compression_control.rs
```diff
@@ -20,6 +22,11 @@ impl ImageCompressionControl {
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetImageSubresourceLayout2EXT.html>
+    ///
+    /// Also available as [`HostImageCopy::get_image_subresource_layout2()`]
+    /// when [`VK_EXT_host_image_copy`] is enabled.
+    ///
+    /// [`VK_EXT_host_image_copy`]: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_EXT_host_image_copy.html
```
MarijnS95, line 29: Still unsure if this is an OR or AND... Maybe an AND in the case that `vk::Image` lives on the host?

MarijnS95: I figured it's an OR. In `vk.xml` both extensions have this command inside a `<require>` block, _without_ extra `depends="the_other_extension"` attribute.

PR #783, opened 2023-08-04
Update Vulkan-Headers to 1.3.269

MarijnS95:
Looks like a bunch of `maintenance5` bitflags had their extension suffix missing, which is added back in the `.261` release.


Ralith (approved):

Ralith (approved):

repi:
would be awesome to get this in, anything in particular that is blocking it?

MarijnS95:
@repi no particular reason, I'm just batching these up until:

- There is a conflict or dependency with another PR;
- A release is happening soon.

But for the time being I am not planning on another patch release (unless required / heavily requested) and would rather sort out the remaining _breaking_ PRs and get `0.38` out the door.

repi:
wow thanks for fast response :) 

and think that makes sense, is this rebased / up-to-date with latest `main` in here? that would be nice in the mean time. the version number seems to be different at 0.37.0 and did get some conflicts when doing a quick test to use this as a fork in an experiment.

MarijnS95:
> wow thanks for fast response :)
> 
> and think that makes sense, is this rebased / up-to-date with latest `main` in here? that would be nice in the mean time.

It is. Any specific extension tat you need?

> the version number seems to be different at 0.37.0 and did get some conflicts when doing a quick test to use this as a fork in an experiment.

Yes, I have yet to decide what the best approach is to updating `main` when patch-compatible releases have been made from a backport branch.

(And on that note, roughly 95% of the PRs/contributions were non-breaking and have been backported, so for the future I might as well stack up open breaking PRs instead, and not have a tedious backports-branch to manage)

repi:
thanks, it is specifically [VK_NV_low_latency2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_NV_low_latency2.html) we are after

MarijnS95:
Right, that also adds some new commands that could conveniently use an `Fn` wrapper.

PR #789, opened 2023-08-30
Adding several missing extensions

Hossein-Noroozpour:
- dedicated allocation
 - pipeline library
 - ray query
 - shader clock


MarijnS95:
These extensions contain no functions, hence they are not manually re-wrapped because this wrapper provides no extra functionality/convenience.

PR #792, opened 2023-09-05
nv/ray_tracing: Initialize tagged struct with `default()`

MarijnS95:
Old and probably unused/deprecated extension now that `VK_KHR_ray_tracing` exists and is stabilized for quite some time, but it's still wrong to zero-initialize `sType`.

Additionally checking all other `mem::zeroed()` calls made me stumble across another PR where `mem::MaybeUninit::zeroed()` was used and `assume_init()` after a function returned successfully.  Perhaps we should switch to that on all occasions (and possibly replace `zeroed()` with `uninit()`)?

### DIFF
--- a/ash/src/extensions/nv/ray_tracing.rs
+++ b/ash/src/extensions/nv/ray_tracing.rs
@@ -57,7 +57,7 @@ impl RayTracing {
         &self,
         info: &vk::AccelerationStructureMemoryRequirementsInfoNV,
     ) -> vk::MemoryRequirements2KHR {
-        let mut requirements = mem::zeroed();
+        let mut requirements = Default::default();
         (self.fp.get_acceleration_structure_memory_requirements_nv)(
             self.handle,
             info,


Ralith (approved):

Ralith:
> Perhaps we should switch to that on all occasions (and possibly replace zeroed() with uninit())?

I like this idea, but:
- it's a pretty marginal difference
- it complicates initializing `sType`, which must now be done with careful use of `addr_of_mut`.

In short, I wouldn't prioritize it.

In my OpenXR bindings, I addressed the latter problem by generating `out` helper functions for everything, which are like `Default` but produce a `MaybeUninit<T>` with the type field populated instead.

MarijnS95:
Perhaps I hadn't been very clear but when I mentioned _"checking all other `mem::zeroed()` calls"_ I was pointing specifically at all _non-`sType`_ datastructures and PODs. All `sType`/`pNext` structures _must_ properly use `default()` (unless it's worth doing a partial `MaybeUninit` where only those two fields are initialized)?

Ralith:
Ah, sure, that's straightforward.

> unless it's worth doing a partial MaybeUninit where only those two fields are initialized

Probably not worth it.

MarijnS95:
Additionally thinking about switching this pattern for simple handles to `let thing = vk::HandleType::null();` instead of any `zeroed()`/`uninit()` implementation, so that we don't keep checking these every once in a while.

Ralith:
That does seem easier to audit at a glance.

PR #795, opened 2023-10-10
Bump `raw-window-handle` to `0.6.0`

Pengiie:
Updated `raw-window-handle` to `0.6.0` and updated `Changelog.md`. 
Waiting for https://github.com/rust-windowing/winit/pull/3126 to update example using raw-window-handle

### DIFF
--- a/ash-window/Cargo.toml
+++ b/ash-window/Cargo.toml
@@ -16,7 +16,7 @@ rust-version = "1.64.0"
 
 [dependencies]
 ash = { path = "../ash", version = "0.37", default-features = false }
-raw-window-handle = "0.5"
+raw-window-handle = "0.6"
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 raw-window-metal = "0.3"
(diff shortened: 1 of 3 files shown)


Pengiie:
Should I modify to follow the rwh strategy in https://github.com/rust-windowing/winit/pull/3126?

Pengiie:
I don't need this now-now, just noticed the `raw-window-handle` update and didn't see a PR here for it yet, I'll close this and wait for your implementation with the backwards compatibility features

notgull:
> I'm already working on this after implementing the producer end in the `ndk`. It's a tad unfortunate that backwards compatibility has to be implemented on both the producer and consumer end, which used to be provided by `raw-window-handle` around the `0.3`/`0.4` era (cc @notgull).

I wasn't aware that backwards compatibility used to be provided in-crate. Although translating a `WindowHandle` from 0.6 is probably harder than translating one from 0.4.

MarijnS95:
> I wasn't aware that backwards compatibility used to be provided in-crate. Although translating a `WindowHandle` from 0.6 is probably harder than translating one from 0.4.

Only in a `0.3` patch release for `0.4` iirc, I don't think the same was done for `0.4 <-> 0.5`? And back then it was only one way around, from producer to consumer iirc.

PR #797, opened 2023-10-11
platform_types: Convert Windows `HANDLE` types to `isize`

MarijnS95:
The `windows` crate treats these as `isize` rather than raw void pointers:
https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Foundation/struct.HWND.html

And `raw-window-handle 0.6` recently started to do the same:
https://github.com/rust-windowing/raw-window-handle/pull/136

However, the win32 documentation still states that these should be `PVOID`:
https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types

Hence I'm not super confident what the right step forward is: `isize` is a breaking change (we'll do a breaking release soon™™™ anyway) but plays more nicely with interop with other crates. @madsmtm was your motivation for `raw-window-handle` also the type representation in windows-rs?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Define `Display` as `c_void` instead of `*mut c_void` to match Xlib (#751)
 - `VK_KHR_device_group_creation`: Take borrow of `Entry` in `fn new()` (#753)
 - `VK_KHR_device_group_creation`: Rename `vk::Instance`-returning function from `device()` to `instance()` (#759)
+- Windows `HANDLE` types (`HWND`, `HINSTANCE`, `HMONITOR`) are now defined as `isize` instead of `*const c_void` (#797)
 
 ### Removed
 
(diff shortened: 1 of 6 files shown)


Ralith (approved):
This is ABI-compatible, so I see no harm in conforming to the de facto standard.

madsmtm:
> @madsmtm was your motivation for `raw-window-handle` also the type representation in windows-rs?

Yup, see also discussion in https://github.com/microsoft/windows-rs/issues/1643.

Apologies for not making that change explicitly, I intended to do so but it must have gotten it mixed in with the linked PR.

MarijnS95:
Thanks @madsmtm, that issue seems rather long and confusing. The TL;DR being that even if a handle may (sometimes?) be a pointer, the user shouldn't be aware of this (and always use appropriate APIs) so an `isize` to hide that (ignoring the discussion about pointer aliasing rules for the compiler) is appropriate.

I'll go ahead and merge this :+1:

MarijnS95:
As noted in the linked `windows-rs` by me some time ago, these types changed back to raw pointers once again:

> For anyone subscribed to this thread, it looks like the types are changed back to pointers recently:
>
> https://github.com/microsoft/win32metadata/issues/1924
> https://github.com/microsoft/win32metadata/commit/b4dfd2f25c0114da8b27d0d390bca5c1555fc55c
>
> Looks like this on a regen: https://github.com/microsoft/windows-rs/commit/96b9a27a6293fe08e4a96b64400239594941ff20

And related discussion https://github.com/rust-windowing/raw-window-handle/issues/171. It doesn't seem like we should blindly change it back though.

PR #798, opened 2023-10-11
Convert `mem::zeroed()` / `0` to `MaybeUninit::uninit()`

MarijnS95:
As noted in #792 changes like this might help us more strictly identify and validate that the argument in question is only relevant as an output argument (i.e. structs with `sType` are read by the function call, even if the caller strictly expects return values there, so that it can fill in multiple structures in a `pNext` chain, and must hence be `Default`- initialized).

It does embed rather nicely with the `assume_init_on_success()` API, though.

---

Creating this as draft just to feel out where we stand with this; it might not net us too much, and doesn't seem like it would help us audit for accidental `sType` structs that are zero/uninit.  There was also the idea to initialize handles with `::null()` explicitly (instead of `mem::zeroed()` or the `MaybeUninit::uninit()` added here), but that also doesn't sound like much of an improvement.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -70,14 +70,14 @@ impl Device {
         create_info: &vk::PrivateDataSlotCreateInfo<'_>,
         allocation_callbacks: Option<&vk::AllocationCallbacks<'_>>,
     ) -> VkResult<vk::PrivateDataSlot> {
-        let mut private_data_slot = mem::zeroed();
+        let mut private_data_slot = mem::MaybeUninit::uninit();
         (self.device_fn_1_3.create_private_data_slot)(
             self.handle,
             create_info,
             allocation_callbacks.as_raw_ptr(),
-            &mut private_data_slot,
+            private_data_slot.as_mut_ptr(),
         )
-        .result_with_success(private_data_slot)
+        .assume_init_on_success(private_data_slot)
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkDestroyPrivateDataSlot.html>
@@ -119,15 +119,15 @@ impl Device {
         object: T,
         private_data_slot: vk::PrivateDataSlot,
     ) -> u64 {
-        let mut data = mem::zeroed();
+        let mut data = mem::MaybeUninit::uninit();
         (self.device_fn_1_3.get_private_data)(
             self.handle,
             T::TYPE,
             object.as_raw(),
             private_data_slot,
-            &mut data,
+            data.as_mut_ptr(),
         );
-        data
+        data.assume_init()
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCmdPipelineBarrier2.html>
@@ -511,16 +511,16 @@ impl Device {
         &self,
         memory_requirements: &vk::DeviceImageMemoryRequirements<'_>,
     ) -> usize {
-        let mut count = 0;
+        let mut count = mem::MaybeUninit::uninit();
         (self
             .device_fn_1_3
             .get_device_image_sparse_memory_requirements)(
             self.handle,
             memory_requirements,
-            &mut count,
+            count.as_mut_ptr(),
             ptr::null_mut(),
         );
-        count as usize
+        count.assume_init() as usize
     }
 
     /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetDeviceImageSparseMemoryRequirements.html>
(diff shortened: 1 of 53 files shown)


ash/src/device.rs
```diff
@@ -828,14 +832,15 @@ impl Device {
         &self,
         info: &vk::ImageSparseMemoryRequirementsInfo2,
     ) -> usize {
-        let mut count = 0;
+        // TODO: when pSparseMemoryRequirements is NULL, is it okay/safe for count to be uninit?
+        let mut count = mem::MaybeUninit::uninit();
```
Ralith, line 836: I think this is sound as rwitten. The spec seems to be missing the relevant text, but consider this analogous section:

> **If pPhysicalDevices is NULL**, then the number of physical devices available is returned in pPhysicalDeviceCount. Otherwise, pPhysicalDeviceCount must point to a variable set by the user to the number of elements in the pPhysicalDevices array [...]

Ralith (approved):
I agree that this helps clarify whether an intention to pass in a specific value exists or not. Perhaps a marginal change, but certainly doesn't hurt.

ash/src/device.rs
MarijnS95: Oh, I thought I had deleted all TODOs from this PR again :sweat_smile:

MarijnS95:
Sure thing! I'll get going on finding the remaining assignments for `0` / `mem::zeroed()`...

MarijnS95:
@Ralith would you mind taking another look? Besides tackling the remaining obvious `= 0;` and `mem::zeroed()`, I've also converted the `let fd = -1;` initializations to this (assuming Vulkan always initializes them when it retrurns `OK`), and replaced a few remaining zero-initialized `Vec`s with `Vec::with_capacity()` + `set_len()`.

ash/src/extensions/amdx/shader_enqueue.rs
```diff
@@ -38,7 +38,9 @@ impl ShaderEnqueue {
             allocation_callbacks.as_raw_ptr(),
             pipelines.as_mut_ptr(),
         )
-        .result_with_success(pipelines)
+        .result()?;
+        pipelines.set_len(create_infos.len());
```
Ralith, line 42: Should we have a `set_len_on_success<T>(self, v: Vec<T>, len: usize) -> Result<Vec<T>>` helper?

Ralith (approved):
LGTM. Didn't double-check the spec for every single command, but each case got at least a quick look.

ash/src/device.rs
MarijnS95: @Ralith I'm marking this as unresolved because, at least for `vkGetLatencyTimingsNV` (https://github.com/ash-rs/ash/issues/814, https://github.com/KhronosGroup/Vulkan-Docs/issues/2269#issuecomment-1813339455), `pTimingCount` will be set to the available number of timings _if it was `0`_, even if `pTimings` **was non-`NULL`**. Meaning that there's probably an `if (!pTimings || !*pTimingCount)` in the driver which could read uninitialized values.

@chansen-nv is adding that mention to the spec, but I don't think this behaviour is consistent/expected wrt other Vulkan functions?

---

However, https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetDeviceImageSparseMemoryRequirements.html only says:

> `pSparseMemoryRequirementCount` is a pointer to an integer related to the number of sparse memory requirements available or queried, as described below.

(There is no description below)

And the VUID doesn't rule out a similar implementation to `vkGetLatencyTimingsNV` either:

> If the value referenced by `pSparseMemoryRequirementCount` is not `0`, and `pSparseMemoryRequirements` is not `NULL`, `pSparseMemoryRequirements` must be a valid pointer to an array of `pSparseMemoryRequirementCount` [VkSparseImageMemoryRequirements2](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSparseImageMemoryRequirements2.html) structures

Perhaps @oddhack can help us clarify whether or not the count pointer/value should be written when it's zero even if the array-pointer is `NULL` (in the generic case). If we cannot get a guarantee on this for some/most functions, we should back this change out and keep a `0` init (`let mut count = 0;`).

ash/src/extensions/amdx/shader_enqueue.rs
MarijnS95: Sounds like a good idea, maybe we can share some code with other vector initializers, I'll have a look at that. Even cooler would be something utilizing `Vec<MaybeUninit<T>>` but `with_capacity()` serves the same purpose for now.

Ralith: > Even cooler would be something utilizing Vec<MaybeUninit<T>>

I think that's specifically redundant to `Vec`'s notion of capacity, yeah. We could do `Box<[MaybeUninit<T>]>` and convert that to `Box<[T]>` but that seems needlessly esoteric.

MarijnS95: Fair point, there are even helper functions to get the uninitialized part of a `Vec` as `&mut [MaybeUninit<T>]` but that's not really of use to us.

Note that a similar call for pipelines in `device.rs` was returning this `set_len`'d `Vec` even when `create_graphics_pipelines()`/`create_compute_pipelines()` returns `Err`. This is because some of the pipelines could have been created, but I don't remember if the driver sets the rest to `VK_NULL_HANDLE` or if we should initialize it to that.

Ralith (changes requested):

Ralith (approved):

PR #799, opened 2023-10-14
ash-window: Upgrade to raw-window-handle 0.6.0

MarijnS95:
This is the successor to #795.  While it doesn't tackle things like backwards-compatibility nor use of the new lifetimed handles (if we _can_ even come up with a safe abstraction for them - it's very likely out of scope!), it includes the following improvements:

- Upgrade `raw-window-metal` via [#8] (Mac/iOS wasn't tested in our CI);
- Fix Windows platform types (in the `ash` crate) to be `isize` instead of `*const c_void`, matching the usptream definition;
- Update example code (impossible until `winit` with `raw-window-handle 0.6` becomes available).

[#8]: https://github.com/norse-rs/raw-window-metal/pull/8

### DIFF
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -14,7 +14,7 @@ use ash::{
 };
 use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
 
-/// Create a surface from a raw surface handle.
+/// Create a surface from a raw display and window handle.
 ///
 /// `instance` must have created with platform specific surface extensions enabled, acquired
 /// through [`enumerate_required_extensions()`].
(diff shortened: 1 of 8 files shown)


Ralith:
winit 0.29 w/ rwh 0.6 is now out.

ilyas-taouaou:
yep winit now uses 0.6

MarijnS95:
I'm subscribed to the releases after being involved with the `ndk`/Android portion of the release process :wink:

SkillerRaptor:
Any updates on this PR?

dvtkrlbs:
I am open to testing for macos and ios tomorrow.

HugoPeters1024:
For what it's worth. This works on Ubuntu 22.04 on my vulkan backend for bevy after updating to bevy 0.13 (which in turn uses the new winit that depends on rwh-0.6).

Skareeg:
I must admit my agreeance with the comment from #864 that RWH 0.6 might be overengineering a little, but I will admit a focus on safety is fine. I am using the feature to downgrade to RWH 0.5 on a personal project but would like to +1 this pull request. If there is anything I can do to assist, let me know.

ash-window/src/lib.rs
```diff
@@ -33,6 +33,47 @@ use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
...
+/// [`Instance`].
+///
+/// The window represented by `window_handle` must be associated with the display connection
+/// in `display_handle`.
+///
+/// `window_handle` and `display_handle` must be associated with a valid window and display
+/// connection, which must not be destroyed for the lifetime of the returned [`vk::SurfaceKHR`].
+///
+/// [parent/child relation]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#fundamentals-objectmodel-lifetime
+#[deprecated = "Direct use of raw window handles is deprecated"]
```
Ralith, line 75: Is there a use case for keeping this at all? Users can already go from raw -> non-raw in unsafe code if an escape hatch is needed.

ash-window/src/lib.rs
```diff
@@ -119,6 +173,21 @@ pub unsafe fn create_surface(
...
+}
+
+/// Query the required instance extensions for creating a surface from a display handle.
+///
+/// This [`RawDisplayHandle`] can typically be acquired from a window, but is usually also
+/// accessible earlier through an "event loop" concept to allow querying required instance
+/// extensions and creation of a compatible Vulkan instance prior to creating a window.
+///
+/// The returned extensions will include all extension dependencies.
+#[deprecated = "Direct use of raw display handles is deprecated"]
```
Ralith, line 189: As above.

Ralith (approved):
LGTM aside from one nit.

ash-window/src/lib.rs
MarijnS95: Sure, there's the easy-to-find `borrow_raw()` helper function.

Inversely, since the returned `ash::SurfaceKHR` type has no notion of an implicit lifetime on this type, we could exclusively keep the raw function, and rely on the user to call the safe `as_raw()` function?

Ralith: True. I think semantically that's slightly cleaner, since we can't do anything with the lifetime information. On the other hand, windowing crates will presumably tend to emit non-raw handles (since `HasRaw...` are deprecated upstream), which makes that form slightly more concise to use.

I lean towards keeping the raw one only, since it's harder to miss the safety invariants that way. So long as we're only exposing one form I'm happy regardless, though.

MarijnS95: Let's stick with the `Raw` interface for those reasons, until upstream starts pestering that code-path with `#[deprecated]` :)

MarijnS95: > On the other hand, windowing crates will presumably tend to emit non-raw handles (since `HasRaw...` are deprecated upstream)

Fortunately we're not taking any of the `Has*` trait variants, so this should be less of an issue besides the user calling `event_loop.display_handle().as_raw()`/`window.window_handle().as_raw()` now.

Ralith: Yeah, I think that aligns well with upstream's intent.

PR #803, opened 2023-10-25
Resolve lint warnings for `deprecated_in_future`, `rust_2018_idioms` and `unused_qualifications`

MarijnS95:
These are 3 non-default lints that cause find a lot of violations in this project that are sensible to resolve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,8 @@ The Vulkan Video bindings are experimental and still seeing breaking changes in 
 ```rust
 // function signature
 pub fn create_instance(&self,
-                       create_info: &vk::InstanceCreateInfo,
-                       allocation_callbacks: Option<&vk::AllocationCallbacks>)
+                       create_info: &vk::InstanceCreateInfo<'_>,
+                       allocation_callbacks: Option<&vk::AllocationCallbacks<'_>>)
                        -> Result<Instance, InstanceError> { .. }
 let instance = entry.create_instance(&create_info, None)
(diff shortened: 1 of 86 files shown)


filnet:
@MarijnS95 you are a source of inspiration and I try to replicate these kind of changes in my project :)

Out of curiosity, why did you move some lints from `deny` to `warn` ?
For example `clippy::use_self` was moved: https://github.com/ash-rs/ash/pull/803/files#diff-3adc4b835bf1dfdf0a10d32290602a8d3b51711aca3a8fc61516b746ebd38370.

MarijnS95:
I don't remember why it was `deny` in the first place and it doesn't make sense to be. Running `cargo clippy` (or via the LSP) it'll bail on the first crate with `use-self` violations rather than printing issues for all of them. And on the CI we turn all warnings into denials (`-Dwarnings`) anyway, so that things won't slip through.

Regarding useful lints themselves, you can find an older list at https://github.com/EmbarkStudios/rust-ecosystem/blob/main/lints.rs. Internally we've modified that a fair bit, and using `ash = { path = "../ash/ash" }` with that for testing causes all these warnings of which most are now solved. Maybe I'll do another pass later.

There should be an issue open about extending that list with more useful lints.

filnet:
Thanks for the detailed answer. Makes sense.

I remember you mentioning the longer list of lints in the past. Will take a closer look.

PR #804, opened 2023-10-25
extensions/ext: Add VK_EXT_hdr_metadata extension

MarijnS95:


ash/src/extensions/ext/mesh_shader.rs
Ralith, line 1: Unrelated change?

Ralith (approved):

ash/src/extensions/ext/mesh_shader.rs
MarijnS95: It's intentional: I used this mod as a base and noticed that `;` was hampering our ability to ensure that the return types are identical.

PR #807, opened 2023-10-26
Add missing ImageAspectFlags

AdrianEddy:
This PR adds remaining ImageAspectFlags in order to support multi-planar image views


MarijnS95:
Hopefully running the CI will tell you:

1. That this file is autogenerated;
2. That these already exist but are part of the Vulkan 1.1 spec and generated elsewhere.

filnet:
@AdrianEddy welcome to the club. Did the exact same thing in my first PR.

PR #809, opened 2023-10-30
Enable `descriptor_count()` setter on `ash::vk::WriteDescriptorSet`

### LINKED ISSUE #806: Descriptor count for INLINE_UNIFORM_BLOCK WriteDescriptorSet
I might be missing something but there's a small paper cut in `WriteDescriptorSet`'s builder API:
```rust
let data: &[u8] = ...;
let mut inline_uniform_block = vk::WriteDescriptorSetInlineUniformBlock::default().data(data);
let mut descriptor_write = vk::WriteDescriptorSet::default()
  .descriptor_type(vk::DescriptorType::INLINE_UNIFORM_BLOCK)
  .dst_set(set)
  .dst_binding(binding)
  .push_next(&mut inline_uniform_block);
descriptor_write.descriptor_count = data.len().try_into().unwrap();
```

The spec for [`VkWriteDescriptorSet`](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkWriteDescriptorSet.html) says:
> descriptorCount is the number of descriptors to update. If the descriptor binding identified by dstSet and dstBinding has a descriptor type of VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK, then descriptorCount specifies the number of bytes to update. Otherwise, descriptorCount is one of
> - the number of elements in pImageInfo
> - the number of elements in pBufferInfo
> [...]

Currently, `descriptorCount` is set by the builder API depending [on the descriptor type](https://github.com/ash-rs/ash/blob/a6f8450edf4721536bc37a056838c54b36905a41/ash/src/vk/definitions.rs#L1843-L1859) so there's no `descriptor_count()` method but it's not special cased for `INLINE_UNIFORM_BLOCK` hence the assignment to`descriptor_count` afterwards.

I guess there's a similar issue for `VkWriteDescriptorSetAccelerationStructureKHR`'s `accelerationStructureCount`.

MarijnS95:
Fixes #806

This parameter is not only used for the length of `pImageInfo`, `pBufferInfo` or `pTexelBufferView`, but also matching the value of `dataSize` when `VkWriteDescriptorSetInlineUniformBlock` is appended in `pNext`, or the value of `accelerationStructureCount` when `VkWriteDescriptorSetAccelerationStructureKHR` is in `pNext`. Having the count setter directly avaialble makes builder code more natural, instead of having to use a `mut` variable and manually assign `.descriptor_count = xx.len();` afterwards.

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkWriteDescriptorSet.html

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_NV_low_latency2` device extension (#802)
 - Added `VK_EXT_hdr_metadata` device extension (#804)
 - Added `VK_NV_cuda_kernel_launch` device extension (#805)
+- Added `descriptor_count()` setter on `ash::vk::WriteDescriptorSet` (#809)
 
 ### Changed
 
--- a/ash/src/vk/definitions.rs
+++ b/ash/src/vk/definitions.rs
@@ -1835,6 +1835,11 @@ impl<'a> WriteDescriptorSet<'a> {
         self
     }
     #[inline]
+    pub fn descriptor_count(mut self, descriptor_count: u32) -> Self {
+        self.descriptor_count = descriptor_count;
+        self
+    }
+    #[inline]
     pub fn descriptor_type(mut self, descriptor_type: DescriptorType) -> Self {
         self.descriptor_type = descriptor_type;
         self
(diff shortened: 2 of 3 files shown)


Ralith (approved):

Ralith:
There's always FRU syntax, but yeah.

MarijnS95:
Sure, but it feels as hidden away as the `mut` variant, while in this case there are multiple obvious uses where this member needs to be used in isolation.

Combined FRU syntax isn't too nice to read in my opinion:

```rust
let write_descriptor_set = vk::WriteDescriptorSet {
    descriptor_count: as.acceleration_structure_count,
    ..Default::default()
}
.dst_set(...)
.dst_binding(0)
.descriptor_type(vk::DescriptorType::ACCELERATION_STRUCTURE_KHR)
.push_next(&mut as);
```

Unless one moves all member assignments (PODs) into FRU, except `.push_next()` which requires casting and should update the lifetime.

PR #815, opened 2023-11-09
extensions/nv/low_latency2: Use `count` parameter for `p_timings` "array" member

### LINKED ISSUE #814: Wrapper `get_latency_timings` of `VK_NV_low_latency2` is incorrect
The `get_latency_timings` for this extension (added in #802) is a bit odd. it takes a single `VkGetLatencyMarkerInfoNV` as a pointer. and that struct has a number of timings. the `count` variable returns the size of `ptimings` NOT the amount of `VkGetLatencyMarkerInfoNV`'s to pass in.

currently in ash the function signature is this:

```rust
    pub unsafe fn get_latency_timings(
        &self,
        swapchain: vk::SwapchainKHR,
        latency_marker_info: &mut [vk::GetLatencyMarkerInfoNV<'_>],
    )
```
But it shouldn't take a slice of `latency_marker_info` but instead a slice of `VkLatencyTimingsFrameReportNV` that is put inside the ptimings variable of `GetLatencyMarkerInfoNV`

This is my manual impl i wrote for testing:

```rust
            // get ptimings count.
            let mut marker = vk::GetLatencyMarkerInfoNV::default();
            let mut count = 0;
            (self
                .device
                .extensions()
                .low_latency2
                .fp()
                .get_latency_timings_nv)(
                self.device.extensions().low_latency2.device(),
                self.swapchain,
                &mut count,
                (&mut marker) as *mut _,
            );

            // alloc ptimings array initialized to default.
            let mut timings = vec![vk::LatencyTimingsFrameReportNV::default(); count as usize];
            marker.p_timings = timings.as_mut_ptr();

            // actually get the timings.
            if count > 0 {
                (self
                    .device
                    .extensions()
                    .low_latency2
                    .fp()
                    .get_latency_timings_nv)(
                    self.device.extensions().low_latency2.device(),
                    self.swapchain,
                    &mut count,
                    (&mut marker) as *mut _,
                );
            }
```

MarijnS95:
Fixes #814, CC @repi @vzout

As it turns out this extension has a rather suboptimal layout where the `vk::GetLatencyMarkerInfoNV` struct is not aware that its own `p_timings` field is an array, because it does not have a count field. Instead the length of this array is passed around via the `pTimingCount` pointer parameter of `get_latency_timings()`.

A couple things are needed to set this straight:
1. The struct builder must be adjusted to accept a slice (without storing the length anywhere...);
2. The `get_latency_timings()` wrapper function should no longer accept a mutable slice of `vk::GetLatencyMarkerInfoNV`, but instead accept a single object that contains a pointer to the array (and `p_next`...);
3. This function must have a separate parameter to pass the length of the `p_timings` slice mentioned at 1., as we cannot reconstruct it from `vk::GetLatencyMarkerInfoNV` nor via its `p_timings` pointer;
4. The mutable struct and initialized `p_timings` array must still be passed fully by the user because both `vk::GetLatencyMarkerInfoNV` and `vk::LatencyTimingsFrameReportNV` have a `p_next` field that the user could initialize to point to some extension structure that they desire to be filled in.  This info might also be relevant in the `_len()` query;
5. Asserts have been put in place to ensure correct use of this API with regards to `p_timings` being (not) NULL for querying the count or actual values, respectively.

### DIFF
--- a/ash/src/vk/definitions.rs
+++ b/ash/src/vk/definitions.rs
@@ -4319,8 +4319,8 @@ impl<'a> PipelineMultisampleStateCreateInfo<'a> {
         self.min_sample_shading = min_sample_shading;
         self
     }
-    #[doc = r" Sets `p_sample_mask` to `null` if the slice is empty. The mask will"]
-    #[doc = r" be treated as if it has all bits set to `1`."]
+    #[doc = r" Sets [`Self::p_sample_mask`] to [`std::ptr::null()`] if the slice is empty. The"]
+    #[doc = r" mask will be treated as if it has all bits set to `1`."]
     #[doc = r""]
     #[doc = r" See <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkPipelineMultisampleStateCreateInfo.html#_description>"]
     #[doc = r" for more details."]
@@ -51890,9 +51890,22 @@ unsafe impl<'a> TaggedStructure for GetLatencyMarkerInfoNV<'a> {
     const STRUCTURE_TYPE: StructureType = StructureType::GET_LATENCY_MARKER_INFO_NV;
 }
 impl<'a> GetLatencyMarkerInfoNV<'a> {
+    #[doc = r" This only sets the slice pointer. Users must manually pass the length of"]
+    #[doc = r" `timings` to [`crate::extensions::nv::LowLatency2::get_latency_timings()`]."]
+    #[doc = r""]
+    #[doc = r" [`Self::p_timings`] will be set to to [`std::ptr::null_mut()`] if the"]
+    #[doc = r" slice is empty, which is necessary to query the length of the array via"]
+    #[doc = r" [`crate::extensions::nv::LowLatency2::get_latency_timings_len()`]."]
+    #[doc = r""]
+    #[doc = r" See <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetLatencyTimingsNV.html#_description>"]
+    #[doc = r" for more details."]
     #[inline]
-    pub fn timings(mut self, timings: &'a mut LatencyTimingsFrameReportNV<'a>) -> Self {
-        self.p_timings = timings;
+    pub fn timings(mut self, timings: &'a mut [LatencyTimingsFrameReportNV<'a>]) -> Self {
+        self.p_timings = if timings.is_empty() {
+            std::ptr::null_mut()
+        } else {
+            timings.as_mut_ptr()
+        };
         self
     }
 }
--- a/ash/src/extensions/nv/low_latency2.rs
+++ b/ash/src/extensions/nv/low_latency2.rs
@@ -4,7 +4,6 @@ use crate::RawPtr;
 use crate::{Device, Instance};
 use std::ffi::CStr;
 use std::mem;
-use std::ptr;
 
 /// <https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VK_NV_low_latency2.html>
 #[derive(Clone)]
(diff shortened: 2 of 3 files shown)


MarijnS95:
If it wasn't for `p_next` in `vk::GetLatencyMarkerInfoNV` we could have hidden away the construction of that type. And if it weren't for `p_next` in `vk::LatencyTimingsFrameReportNV` we could have done the allocation in a helper function and hid all this away and just return a `Vec<vk::LatencyTimingsFrameReportNV<'_>>`...

ash/src/vk/definitions.rs
```diff
@@ -51890,9 +51890,22 @@ unsafe impl<'a> TaggedStructure for GetLatencyMarkerInfoNV<'a> {
...
+    #[doc = r" [`Self::p_timings`] will be set to to [`std::ptr::null_mut()`] if the"]
+    #[doc = r" slice is empty, which is necessary to query the length of the array via"]
+    #[doc = r" [`crate::extensions::nv::LowLatency2::get_latency_timings_len()`]."]
+    #[doc = r""]
+    #[doc = r" See <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetLatencyTimingsNV.html#_description>"]
+    #[doc = r" for more details."]
     #[inline]
-    pub fn timings(mut self, timings: &'a mut LatencyTimingsFrameReportNV<'a>) -> Self {
-        self.p_timings = timings;
+    pub fn timings(mut self, timings: &'a mut [LatencyTimingsFrameReportNV<'a>]) -> Self {
```
MarijnS95, line 51903: It seems all lifetime-fu has escaped me; I can no longer use the slice passed as `timings` here even after `self` is dropped. Notably this is the first and only mutable slice that contains objects with a lifetime parameter.

```rust
let mut info = vk::GetLatencyMarkerInfoNV::default();
let latency_timings = base
    .low_latency2_loader
    .get_latency_timings_len(base.swapchain, &mut info);
dbg!(latency_timings);
if latency_timings != 0 {
    let mut timings = vec![vk::LatencyTimingsFrameReportNV::default(); latency_timings];
    {
        // Creating a new object just for show
        let mut info = vk::GetLatencyMarkerInfoNV::default().timings(&mut timings);
        base.low_latency2_loader.get_latency_timings(
            base.swapchain,
            latency_timings,
            &mut info,
        );
        // `info` should be dropped here, removing the mutable borrow on `timings`?
    }
    dbg!(&timings); // Error: `timings` is still borrowed in `.timings(&mut timings);` above?
}
```

```
error[E0502]: cannot borrow `timings` as immutable because it is also borrowed as mutable
   --> examples/src/bin/texture.rs:800:22
    |
793 |                     let mut info = vk::GetLatencyMarkerInfoNV::default().timings(&mut timings);
    |                                                                                  ------------ mutable borrow occurs here
...
800 |                 dbg!(&timings);
    |                      ^^^^^^^^
    |                      |
    |                      immutable borrow occurs here
    |                      mutable borrow later used here
```

Ralith: I think by constructing an `&'a mut [LatencyTimingsFrameReportNV<'a>]`, you're saying that the slice is uniquely borrowed for (`&'a mut`) a lifetime that outlives it (`T<'a>`). Does it work better if you do `for<'b: 'a> &'a mut [LatencyTimingsFrameReportNV<'b>]`?

MarijnS95: @Ralith that is indeed the first thing I tried, making sense that `'b` is a lifetime that is bigger / lives longer than `'a` (because `LatencyTimingsFrameReportNV` and the things that it borrows can live longer than the slice it is eventually stored within).

Unfortunately the compiler is not having it:

```rust
pub fn timings<'b: 'a>(mut self, timings: &'a mut [LatencyTimingsFrameReportNV<'b>]) -> Self {
    self.p_timings = if timings.is_empty() {
        std::ptr::null_mut()
    } else {
        timings.as_mut_ptr()
    };
    self
}
```

Results in:

```
error: lifetime may not live long enough
     --> ash/src/vk/definitions.rs:51910:13
      |
51892 | impl<'a> GetLatencyMarkerInfoNV<'a> {
      |      -- lifetime `'a` defined here
...
51906 |     pub fn timings<'b: 'a>(mut self, timings: &'a mut [LatencyTimingsFrameReportNV<'b>]) -> Self {
      |                    -- lifetime `'b` defined here
...
51910 |             timings.as_mut_ptr()
      |             ^^^^^^^^^^^^^^^^^^^^ argument requires that `'a` must outlive `'b`
      |
      = help: consider adding the following bound: `'a: 'b`
      = note: requirement occurs because of a mutable reference to `[vk::definitions::LatencyTimingsFrameReportNV<'_>]`
      = note: mutable references are invariant over their type parameter
      = help: see <https://doc.rust-lang.org/nomicon/subtyping.html> for more information about variance
```

Writing the bound the other way around results in the same error as before. And it doesn't make much sense for `'a` to outlive `'b` when `'a` is referring to a thing of shorter 
ifetime `'b`?

MarijnS95:
Let's hold off merging these changes for a little bit until upstream fixes land in `vk.xml`: https://github.com/KhronosGroup/Vulkan-Docs/issues/2269#issuecomment-1813339455

MarijnS95:
The extension API definition has been fixed upstream (`pCount` is moved from being a function argument to a field on `VkGetLatencyMarkerInfoNV`), and our generator now "correctly" (see above regarding the lifetime...) generates a slice-setter. The changes are pulled in via a regular Vulkan spec-update PR: #816

Ralith:
Awesome work!

PR #818, opened 2023-11-14
Bump `libloading` to `0.8`

AngelicosPhosphoros:
Main motivation is that in my projects, many dependencies use windows-sys to interface Windows API so and only ash keeps using `winapi` crate because of libloading. Libloading 0.8 switched to `windows-sys` so problem solved.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_EXT_hdr_metadata` device extension (#804)
 - Added `VK_NV_cuda_kernel_launch` device extension (#805)
 - Added `descriptor_count()` setter on `ash::vk::WriteDescriptorSet` (#809)
+- Bumped `libloading` crate dependency to 0.8
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_EXT_hdr_metadata` device extension (#804)
 - Added `VK_NV_cuda_kernel_launch` device extension (#805)
 - Added `descriptor_count()` setter on `ash::vk::WriteDescriptorSet` (#809)
+- Bumped `libloading` crate dependency to 0.8
```
MarijnS95, line 32: Can you move this under `# Changed` and add the PR number?

MarijnS95 (approved):
Yeah glad to see winapi gone. How long has this been out for? Looks like our dependabot setup which was created to tell us about this isn't working yet.

Ralith:
Note that you can also avoid this dependency by disabling the "loaded" feature, and using "linked" instead. This does require that the Vulkan SDK be installed in the build environment, though.

MarijnS95:
I recalled this being in the changelog already, and it turns out that's in #739 where we also bump `libloading` to `0.8` so I'll close this one to not have a dupe open.

PR #820, opened 2023-11-15
Rename `examples` to `ash-examples`

MarijnS95:
Depends on #819

[Dependabot complains] that:

     the binary target name `examples` is forbidden, it conflicts with with cargo's build directory names

And fails to provide dependency upgrades for Rust code.  Fix that by renaming the folder and crate to `ash-examples`.

[Dependabot complains]: https://github.com/ash-rs/ash/network/updates/748770724


Ralith (approved):
Weird, I don't think we actually have a *binary target* by that name, specifically? Probably a dependabot bug, but oh well.

MarijnS95:
> Weird, I don't think we actually have a _binary target_ by that name, specifically? Probably a dependabot bug, but oh well.

Yeah, I found this suspicious as well but didn't read too much into it because folks like to call shared objects "binary" as well. That said it should only emit an `rlib`?

PR #822, opened 2023-11-18
Move lints configuration to manifest

filnet:
This is possible since [Rust 1.74.0](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html).

This PR shows what it looks like.


ash/Cargo.toml
```diff
@@ -39,3 +39,16 @@ no-dev-version = true
...
+[lints.rust]
+deprecated_in_future = "warn"
+rust_2018_idioms = "warn"
+trivial_casts = "warn"
+trivial_numeric_casts = "warn"
+unused_qualifications = "warn"
+
+[lints.clippy]
+use_self = "warn"
+too_many_arguments = "allow"
```
filnet, line 52: Unfortunately it is not possible to override lints from the root.
So here the common lints are duplicated to be able to add the three "allows"

MarijnS95 (changes requested):
I am intentionally not pushing this to the `ash` repo because:

1. Lints are explicitly inherited from the workspace, not implicitly forced down;
2. Unnecessary MSRV bump;
3. Strange duplication in overrides.

Can I close this as "Not planned"?

filnet:
> Can I close this as "Not planned"?

No problems...

filnet:
I also found it strange that workspace lints can't be overridden at the workspace member level.
This behavior is not documented.

filnet:
I dropped the MSRV bump.

What about the lint fixes and the "resolve" warning fix ? Should we keep them ?

filnet:
I moved the `ash` package specific lints back to the `lib.rs` file and enabled the workspace lints.
This works as "expected".

PS: I don't expect this to be merged. Just showcasing.

MarijnS95:
> PS: I don't expect this to be merged. Just showcasing.

Note that everyone gets notifications when you're debugging by means of (force-)pushing.

filnet:
> Note that everyone gets notifications when you're debugging by means of (force-)pushing.

Sorry about that. I'll be more careful next time.

filnet:
Current status / limitations:
* all common lints are now configured at the workspace level.
* `ash-window` now uses the common lints (was using a subset).
* as mentioned earlier, it is not possible to override or tweak the lints at the workspace member level (but it is still possible to do it from  `main.rs`/`lib.rs`.

PR #828, opened 2023-11-22
extensions: Always return `pipeline`/`shaders`, even on error

MarijnS95:
In `ash::Device` `create_compute_pipeline()` and `create_graphics_pipeline()` already return the list of pipelines regardless of the error code, as [documented in the Multiple Pipeline Creation chapter].  Callers are expected to scan the returned array for valid pipeline handles and either use or destroy them (when handling an error).  Furthermore, the caller is guaranteed that all handles will be NULL after the first returned NULL handle when `VK_PIPELINE_CREATE_EARLY_RETURN_ON_FAILURE_BIT` is set.

An upstream Khronos/Vulkan request which has since been merged to the documentation makes all pipeline creation functions across the core and extensions point to this `Multiple Pipeline Creation` chapter which has also been improved, clarifying that they too return impartial results. This documentation reference has been embedded in `ash`.

`VK_EXT_shader_object` is similar in nature, but has its own documentation to account for the difference in naming.

[documented in the Multiple Pipeline Creation chapter]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap10.html#pipelines-multiple

### DIFF
--- a/ash/src/extensions/ext/shader_object.rs
+++ b/ash/src/extensions/ext/shader_object.rs
@@ -8,21 +8,34 @@ use core::ptr;
 
 impl crate::ext::shader_object::Device {
     /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCreateShadersEXT.html>
+    ///
+    /// When this function returns, whether or not it succeeds, it is guaranteed that every returned
+    /// element is either [`vk::ShaderEXT::null()`] or a valid [`vk::ShaderEXT`] handle.
+    ///
+    /// This means that whenever shader creation fails, the application can determine which shader
+    /// the returned error pertains to by locating the first [`vk::Handle::is_null()`] element
+    /// in the returned [`Vec`]. It also means that an application can reliably clean up from a
+    /// failed call by iterating over the returned [`Vec`] and destroying every element that is not
+    /// [`vk::Handle::is_null()`].
     #[inline]
     pub unsafe fn create_shaders(
         &self,
         create_infos: &[vk::ShaderCreateInfoEXT<'_>],
         allocator: Option<&vk::AllocationCallbacks<'_>>,
-    ) -> VkResult<Vec<vk::ShaderEXT>> {
+    ) -> Result<Vec<vk::ShaderEXT>, (Vec<vk::ShaderEXT>, vk::Result)> {
         let mut shaders = Vec::with_capacity(create_infos.len());
-        (self.fp.create_shaders_ext)(
+        let err_code = (self.fp.create_shaders_ext)(
             self.handle,
             create_infos.len() as u32,
             create_infos.as_ptr(),
             allocator.as_raw_ptr(),
             shaders.as_mut_ptr(),
-        )
-        .set_vec_len_on_success(shaders, create_infos.len())
+        );
+        shaders.set_len(create_infos.len());
+        match err_code {
+            vk::Result::SUCCESS => Ok(shaders),
+            _ => Err((shaders, err_code)),
+        }
     }
 
     /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkDestroyShaderEXT.html>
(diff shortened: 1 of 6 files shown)


Ralith (approved):
It's tempting to push for `(Vec<T>, Result<()>)` here, but it does make it a lot easier to forget to check the error condition.

MarijnS95:
> It's tempting to push for `(Vec<T>, Result<()>)` here, but it does make it a lot easier to forget to check the error condition.

Yeah it's unfortunate that there's no lint to catch an unused `Result` when it is returned as part of a tuple.

MarijnS95:
https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap10.html#pipelines-multiple

Turns out there's actually documentation for this, and the references+language are being updated upstream.

I'd still like to think of a better way to represent the validity of the return values (and necessity to clean up non-NULL handles) before merging this in.

MarijnS95:
I've been contemplating that but it sets the false precedent that every pipeline individually has an error code. It seems better than the status-quo, though!

Ralith:
`(Vec<Option<T>>, VkResult)`?

MarijnS95:
Also considered, but that might be ambiguous with `None` vs `Some(vk::Handle::null())` :sweat_smile:

MarijnS95:
We're gonna write high-level bindings at this point, but how about an `OwnedPipeline` that has an `fn take(self) -> vk::Pipeline` drops itself if it wasn't taken (via `std::mem::take()`)?

It wouldn't have access to the `device` to do so though if the type is `#[repr(transparent)]`.

MarijnS95:
> Anything like RAII ownership of Vulkan objects starts feeling out of scope for ash.

It'd also start implying that similar mechanisms are available or supposed to be available on other types, which wouldn't be the case by for `ash` in any shape or form.

Ralith (approved):

MarijnS95:
Upstream documentation improvements have been merged, command pages now contain this:

> Pipelines are created and returned as described for [Multiple Pipeline Creation](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap10.html#pipelines-multiple).

I'll adopt and update the relevant portions of that doc here in `ash` on _all_ functions, before merging this. While the function interface indeed shows that users have to do some work on error, it's better to explain and/or link _what_ explicitly.

MarijnS95:
Perhaps that doc is a bit excessive. I pruned it down somewhat, but we could also boil it down to just the link, or perhaps at the very least (because of the return signature) state that valid references may always be returned - corresponding to `create_infos[i]` - and should either be used or desroyed.

Ralith:
Yeah, I think something a lot more terse will suffice. No need to duplicate upstream docs at length. Maybe something like:

> If multiple pipelines are being created in one call, a subset of pipelines might be created even in the event of an error. Care should be taken not to leak these.

ash/src/extensions/ext/shader_object.rs
```diff
@@ -8,21 +8,34 @@ use core::ptr;
 
 impl crate::ext::shader_object::Device {
     /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCreateShadersEXT.html>
+    ///
+    /// When this function returns, whether or not it succeeds, it is guaranteed that every returned
+    /// element is either [`vk::ShaderEXT::null()`] or a valid [`vk::ShaderEXT`] handle.
```
Ralith, line 13: is it intentional that this one case has different docs?

Ralith (approved):

ash/src/extensions/ext/shader_object.rs
MarijnS95: Yeah, I mentioned that in the updated commit-message/PR-description: upstream doesn't even reference the multi pipeline creation chapter. Probably because of the subject conflicts (pipeline != shader object even though the functions behave identically in these cases).

The upstream docs are terse enough that it makes sense to include them here, I think.

Ralith: SGTM.

PR #829, opened 2023-11-23
extensions/khr/ray_tracing_pipeline: Pass indirect SBT regions as single item

MarijnS95:
In an identical fashion to commit 84624fd ("ray_tracing_pipeline: Pass SBT regions as reference instead of slice (#350)") `cmd_trace_rays_indirect()` also only needs a pointer to a single `StridedDeviceAddressRegionKHR` structure.  After all no length is ever passed to the API anywhere, and this could also lead to users passing empty slices, or passing too many elements that are never used.

Clear up the confusion by replacing the slice argument with a direct borrow of the struct.

### DIFF
--- a/ash/src/extensions/khr/ray_tracing_pipeline.rs
+++ b/ash/src/extensions/khr/ray_tracing_pipeline.rs
@@ -122,18 +122,18 @@ impl RayTracingPipeline {
     pub unsafe fn cmd_trace_rays_indirect(
         &self,
         command_buffer: vk::CommandBuffer,
-        raygen_shader_binding_table: &[vk::StridedDeviceAddressRegionKHR],
-        miss_shader_binding_table: &[vk::StridedDeviceAddressRegionKHR],
-        hit_shader_binding_table: &[vk::StridedDeviceAddressRegionKHR],
-        callable_shader_binding_table: &[vk::StridedDeviceAddressRegionKHR],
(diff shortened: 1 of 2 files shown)


ash/src/extensions/khr/ray_tracing_pipeline.rs
```diff
@@ -51,20 +51,24 @@ impl RayTracingPipeline {
         &self,
         deferred_operation: vk::DeferredOperationKHR,
         pipeline_cache: vk::PipelineCache,
-        create_info: &[vk::RayTracingPipelineCreateInfoKHR<'_>],
+        create_infos: &[vk::RayTracingPipelineCreateInfoKHR<'_>],
         allocation_callbacks: Option<&vk::AllocationCallbacks<'_>>,
-    ) -> VkResult<Vec<vk::Pipeline>> {
-        let mut pipelines = vec![mem::zeroed(); create_info.len()];
-        (self.fp.create_ray_tracing_pipelines_khr)(
+    ) -> Result<Vec<vk::Pipeline>, (Vec<vk::Pipeline>, vk::Result)> {
```
Ralith, line 56: This seems to be unrelated to the change described by the commit.

MarijnS95: Looks like I did end up making a mess after all when rebasing these changes off of that PR, and failing to properly reload the file without those changes in my editor. Fixed now!

Ralith (approved):
Nice catch!

Issue #830, opened 2023-11-24
non UTF8 encoded string in enumerate_instance_extension_properties function


GummyGun:
There is a bug in the method enumerate_instance_extension_properties() where sometimes the last character in the field ExtensionProperties.extension_name is not properly encoded and may hold non-valid utf8 char.

GummyGun:
this bug to my findings is only present in the extension VK_EXT_direct_mode_display

MarijnS95:
Can you share the `u8` data that this array contains, i.e. the way that you ran into this issue?

MarijnS95:
After checking the ICD on my ArchLinux machine with the following snippet:

```rust
            for p in entry.enumerate_instance_extension_properties(None).unwrap() {
                let n = ::std::ffi::CStr::from_ptr(p.extension_name.as_ptr());

                if n.to_string_lossy() == "VK_EXT_direct_mode_display" {
                    dbg!(p.extension_name);
                }
            }
```

I get a sensible result out of the array:

```
[ash-examples/src/lib.rs:229] p.extension_name = [
    86,
    75,
    95,
    69,
    88,
    84,
    95,
    100,
    105,
    114,
    101,
    99,
    116,
    95,
    109,
    111,
    100,
    101,
    95,
    100,
    105,
    115,
    112,
    108,
    97,
    121,
    0,
    ...,
    0,
]
```

Note that Ash initially allocates uninitialized memory for these structs, and it's up to the ICD to optionally zero out the whole string buffer after the first `NUL` character. However, that might even be optional given the description at https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkExtensionProperties.html:

> `extensionName` is an array of `VK_MAX_EXTENSION_NAME_SIZE` char containing a null-terminated UTF-8 string which is the name of the extension.

Typically (as far as I'm aware) implementations should stop reading data after the first `NUL` byte/character.

Are you sure that your test case isn't accidentally reading _beyond the `NUL` character_?

GummyGun:
I should use the `CStr` since it saves me a unsafe block and over all is a better practice and I'm pretty sure `CStr` would solve the problem. Either way I think it is worth fixing the this awkward behavior.

```
let extension_list = entry.enumerate_instance_extension_properties(None).unwrap();
 
for extension in &extension_list {
    println!("{:?}", extension);
    println!("{:?}", extension.extension_name);
    if extension.extension_name[255] != 0 {
        panic!("found nonValid utf8");
    }
[...]

PR #831, opened 2023-11-25
Provide `CStr` getters and setters for `c_char` pointers and arrays

MarijnS95:
It is a common operation to read and write NUL-terminated C string in the Vulkan API, yet the only helpers for that were thus far open-coded in the `Debug` printing implementations.

Move them to separate functions that are exposed to the user, in hopes of helping them no longer misunderstand NUL-terminated strings (see e.g. #830).

Important to note is that the array-copy for a static-sized `c_char` array has also been replaced with a `CStr` wrapper: this forces the user and our implementation to have a NUL-terminated string, but may now also panic if the incoming `CStr` with NUL-terminator is too large for the target `c_char` array.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_EXT_hdr_metadata` device extension (#804)
 - Added `VK_NV_cuda_kernel_launch` device extension (#805)
 - Added `descriptor_count()` setter on `ash::vk::WriteDescriptorSet` (#809)
+- Added `*_as_c_str()` getters for `c_char` pointers and `c_char` arrays (#831)
 
 ### Changed
 
@@ -48,7 +49,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `VK_KHR_device_group_creation`: Take borrow of `Entry` in `fn new()` (#753)
 - `VK_KHR_device_group_creation`: Rename `vk::Instance`-returning function from `device()` to `instance()` (#759)
 - Windows `HANDLE` types (`HWND`, `HINSTANCE`, `HMONITOR`) are now defined as `isize` instead of `*const c_void` (#797)
-- extensions/ext/ray_tracing_pipeline: Pass indirect SBT regions as single item reference. (#829)
+- extensions/ext/ray_tracing_pipeline: Pass indirect SBT regions as single item reference (#829)
+- Replaced `c_char` array setters with `CStr` setters (#831)
 
 ### Removed
 
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1822,25 +1822,16 @@ fn derive_debug(
         let param_ident = field.param_ident();
         let param_str = param_ident.to_string();
         let debug_value = if is_static_array(field) && field.basetype == "char" {
-            quote! {
-                &unsafe {
-                    ::std::ffi::CStr::from_ptr(self.#param_ident.as_ptr())
-                }
-            }
+            let param_ident = format_ident!("{}_as_c_str", param_ident);
+            quote!(&unsafe { self.#param_ident() })
         } else if param_str.contains("pfn") {
-            quote! {
-                &(self.#param_ident.map(|x| x as *const ()))
-            }
+            quote!(&(self.#param_ident.map(|x| x as *const ())))
         } else if union_types.contains(field.basetype.as_str()) {
             quote!(&"union")
         } else {
-            quote! {
-                &self.#param_ident
-            }
+            quote!(&self.#param_ident)
         };
-        quote! {
-            .field(#param_str, #debug_value)
-        }
+        quote!(.field(#param_str, #debug_value))
     });
     let name_str = name.to_string();
     let lifetime = has_lifetime.then(|| quote!(<'_>));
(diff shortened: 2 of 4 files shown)


MarijnS95:
> but may now also panic if the incoming `CStr` with NUL-terminator is too large for the target `c_char` array.

Pushed a second commit that returns a `Result` here based on `get_mut(..bytes.len())` returning `None`, but I'm not too happy about it yet...

Ralith:
IMO the `Result` is fine, so this LGTM overall. One note: [std precedent](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.as_c_str) is `as_c_str`, not `as_cstr`. This is more consistent with the capitalization of the type, besides.

MarijnS95:
Ah fair enough, I hadn't seen that yet but I agree; corrected that.

Bit unfortunate that the standard library didn't follow this convention with `into_cstring` (same capitalization on `CString`) while they did for `into_boxed_c_str`).

Ralith (approved):

Changelog.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_EXT_hdr_metadata` device extension (#804)
 - Added `VK_NV_cuda_kernel_launch` device extension (#805)
 - Added `descriptor_count()` setter on `ash::vk::WriteDescriptorSet` (#809)
+- Added `*_as_c_str()` getters for `c_char` pointers and `c_char` arrays (#831)
```
MarijnS95, line 32: Nit in hindsight: we also added _setters_ for `c_char` pointers... And only replaced the `c_char` **array** setters with a `CStr` interface.

generator/src/lib.rs
```diff
@@ -1822,25 +1822,16 @@ fn derive_debug(
         let param_ident = field.param_ident();
         let param_str = param_ident.to_string();
         let debug_value = if is_static_array(field) && field.basetype == "char" {
-            quote! {
-                &unsafe {
-                    ::std::ffi::CStr::from_ptr(self.#param_ident.as_ptr())
-                }
-            }
+            let param_ident = format_ident!("{}_as_c_str", param_ident);
+            quote!(&unsafe { self.#param_ident() })
```
Ralith, line 1826: I think this is actually unsound: the user can safely construct a structure with non-null-terminated data in a string field. We need a checked version for `Debug`. Not sure it's worth the trouble to preserve an unchecked version along side.

Ralith (changes requested):

generator/src/lib.rs
MarijnS95: It has always been unsound as written. This PR is not changing those semantics, #746 is taking care of making that safe.

We can also make a separate contribution to address those soundness issues by means of cloning the implementation of `CStr::from_bytes_until_nul()` for a slice, but (as written in https://github.com/ash-rs/ash/pull/746#issuecomment-1826270501) I'd rather bump the MSRV instead.

Ralith (approved):

generator/src/lib.rs
```diff
@@ -1989,83 +1980,103 @@ fn derive_setters(
...
                     #deprecated
-                    pub fn #param_ident_short(mut self, #param_ident_short: &'a ::std::ffi::CStr) -> Self {
+                    pub fn #param_ident_short(mut self, #param_ident_short: &'a std::ffi::CStr) -> Self {
                         self.#param_ident = #param_ident_short.as_ptr();
                         self
                     }
+                    #[inline]
+                    #deprecated
+                    pub unsafe fn #param_ident_as_c_str(&self) -> &std::ffi::CStr {
+                        std::ffi::CStr::from_ptr(self.#param_ident)
```
MarijnS95, line 1998: @Ralith I don't remember if we discussed this but these pointers can trivially be `NULL`. Fortunately the function is `unsafe` and this call is _not_ used in `Debug`, so I'm curious if it was deliberate.

However, we might want to switch this to return `Option` to represent that (and to safely use this API in `Debug`.

Ralith: Oops. I agree we should return `Option` here, and further, we should probably accept `Option` in the setters.

MarijnS95: > we should probably accept `Option` in the setters

Same for the pointer slice setters.

> I don't think we can safely use this API in `Debug` regardless because the pointer might be garbage.

Exactly, that's also what I wrote in https://github.com/ash-rs/ash/pull/858#issue-2075037358: we can only print (subslices of) static arrays that are embedded in known-valid structs.

MarijnS95: > Same for the pointer slice setters.

And arbitrary struct pointers (if Vulkan denotes them optional). Perhaps we should ask ourselves if the builder pattern isn't intended to (re)set fields to `NULL`?

Ralith: Good point. Let's not. The asymmetry is weird, but treating slices specially is weirder, and putting `Option` in every setter would be a significant downgrade in ergonomics.

MarijnS95: > The asymmetry is weird

On that note, how about providing similar `as_slice()` helpers for the pointer-slice builders?

Ralith: Seems reasonable, though of limited use outside implementing layers or ICDs.

MarijnS95: In hindsight it makes total sense, as the caller already allocates and passes those mutable arrays/sizes. That's not the case for the static-arrays in #858, so they're the only ones with a special case.

Ralith: I'm confused -- are you saying it makes sense to not have getters for pointer-slice fields? Why would the storage not being inline make that any less useful?

MarijnS95: As you said, it's of limited use outside of layers and ICDs. Might be a useful addition if we wish to target those.

Just thinking out loud why I've never needed these slices helpers in client code :wink:

Ralith: Ah, yeah. I know there's been occasional interest. We could have a feature gate for implementer-side APIs, even. Still, far from urgent.

PR #841, opened 2023-12-02
generator: Work around invariance for assigning mutable pointer of lifetimed slice

### LINKED ISSUE #837: NV_low_latency2 Impossible to use `timings` array after mutably borrowing in `GetLatencyMarkerInfoNV`
CC @repi @VZout, in case you run into this while incorporating the updated extension bindings in your code. This happens after https://github.com/ash-rs/ash/pull/816 was merged.

The `timings()` builder function on `GetLatencyMarkerInfoNV` mutably borrows a slice of `LatencyTimingsFrameReportNV`, which itself contains pointers and a lifetime. Rust seems to assume

```rust
#[test]
fn lifetime() {
    let mut timings = vec![vk::LatencyTimingsFrameReportNV::default()];
    let info = vk::GetLatencyMarkerInfoNV::default().timings(&mut timings);
    drop(info); // Lifetime wouldn't extend beyond this point if info is no longer used
    dbg!(timings);
}
```

Results in:

```
error[E0505]: cannot move out of `timings` because it is borrowed
  --> ash/src/extensions/nv/low_latency2.rs:93:5
   |
90 |     let mut timings = vec![vk::LatencyTimingsFrameReportNV::default()];
   |         ----------- binding `timings` declared here
91 |     let info = vk::GetLatencyMarkerInfoNV::default().timings(&mut timings);
   |                                                              ------------ borrow of `timings` occurs here
92 |     drop(info); // Lifetime wouldn't extend beyond this point if info is no longer used
93 |     dbg!(timings);
   |     ^^^^^^^^^^^^^
   |     |
   |     move out of `timings` occurs here
   |     borrow later used here
   |
   = note: this error originates in the macro `dbg` (in Nightly builds, run with -Z macro-backtrace for more info)
```

(EDIT: Instead of `drop(info);`, we could also wrap `let info` in a new lifetime scope `{}`)

More context: https://github.com/ash-rs/ash/pull/815#discussion_r1392679017

_Originally posted by @MarijnS95 in https://github.com/ash-rs/ash/pull/816#discussion_r1405804943_

MarijnS95:
Fixes #837

@Ralith I'm in quite a hurry but managed to piece together a rather minimal generator change to solve the issue, but probably did a **horrible** job of understanding, summarizing and re-explaining the issues with variance.  Feel free to suggest something better, otherwise I'll revisit it when I return.

---

In essence this builder function needs to adhere to two rules:
1. No ref-after-free: the slice must outlive (uses of) the builder object;
2. No aliasing: the slice cannot be (im)mutably used while it is mutably borrowed within a live builder object.

These two rules have been tested and are satisfied by the given builder implementation.  Without this change `timings` seems to be borrowing itself, hence is not allowed to be used after it has been temporarily mutably borrowed inside the builder, even after that builder was dropped.  Thus defeating the purpose of this "getter" API via a struct.

Without the `.cast()`, because mutable raw pointers are invariant (i.e. there is no subtyping relationship) the compiler complains about requiring `self` to outlive `timings` instead, which does not satisfy the two rules above.

### DIFF
--- a/ash/src/vk/definitions.rs
+++ b/ash/src/vk/definitions.rs
@@ -52045,9 +52045,9 @@ unsafe impl<'a> TaggedStructure for GetLatencyMarkerInfoNV<'a> {
 }
 impl<'a> GetLatencyMarkerInfoNV<'a> {
     #[inline]
-    pub fn timings(mut self, timings: &'a mut [LatencyTimingsFrameReportNV<'a>]) -> Self {
+    pub fn timings(mut self, timings: &'a mut [LatencyTimingsFrameReportNV<'_>]) -> Self {
         self.timing_count = timings.len() as _;
-        self.p_timings = timings.as_mut_ptr();
+        self.p_timings = timings.as_mut_ptr().cast();
         self
     }
 }
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1929,7 +1929,7 @@ fn derive_setters(
 
         let deprecated = member.deprecated.as_ref().map(|d| quote!(#d #[allow(deprecated)]));
         let param_ident = field.param_ident();
-        let type_lifetime = has_lifetimes
+        let mut type_lifetime = has_lifetimes
             .contains(&name_to_tokens(&field.basetype))
             .then(|| quote!(<'a>));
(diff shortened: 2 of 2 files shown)


Ralith (approved):
This looks great!

PR #843, opened 2023-12-02
README: Modernize "install guide" for Linux systems

MarijnS95:
Nowadays there's not much to do: get the provided SDK (ICD and VVL) from your package manager, and make sure there's a compatible Vulkan driver for the hardware installed in your machine (or an emulator like LavaPipe).  There is no longer a need to mess around with environment variables pointed out by a 7-year-old stackoverflow post, which typically leaves the Vulkan environment in a broken state instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,7 +204,12 @@ Make sure that you have a Vulkan ready driver and install the [LunarG Vulkan SDK
 
 #### Linux
 
-Make sure that you have a Vulkan ready driver and install the [LunarG Vulkan SDK](https://lunarg.com/vulkan-sdk/). You also have to add the library and layers to your path. Have a look at my [post](http://askubuntu.com/a/803110/77183) if you are unsure how to do that.
+Install a Vulkan driver for your graphics hardware of choice, and (optionally) the [Validation Layers](https://github.com/KhronosGroup/Vulkan-ValidationLayers) via your package manager:
+
+- Arch Linux: https://wiki.archlinux.org/title/Vulkan.
+- Gentoo: https://wiki.gentoo.org/wiki/Vulkan.
+- Ubuntu/Debian: Besides installing a compatible graphics driver, install [`vulkan-validationlayers`](https://packages.ubuntu.com/vulkan-validationlayers) ([Debian](https://packages.debian.org/search?keywords=vulkan-validationlayers)) for the Validation Layers.
+- Other distros: consult your distro documentation and/or package repository for the preferred method to install and use Vulkan.
 
 #### macOS
 


README.md
```diff
@@ -204,7 +204,10 @@ Make sure that you have a Vulkan ready driver and install the [LunarG Vulkan SDK
 
 #### Linux
 
-Make sure that you have a Vulkan ready driver and install the [LunarG Vulkan SDK](https://lunarg.com/vulkan-sdk/). You also have to add the library and layers to your path. Have a look at my [post](http://askubuntu.com/a/803110/77183) if you are unsure how to do that.
+Install a Vulkan driver for your graphics hardware of choice, and the [LunarG Vulkan SDK](https://lunarg.com/vulkan-sdk/) for the [ICD](https://github.com/KhronosGroup/Vulkan-Loader) and [Validation Layers](https://github.com/KhronosGroup/Vulkan-ValidationLayers).
+
+- Arch Linux: https://wiki.archlinux.org/title/Vulkan
+- Ubuntu: https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=vulkan-loader&searchon=sourcenames
```
MarijnS95, line 210: Any other distro that we should mention here? I could so far only find useful docs for Arch, the rest is just package manager searches...

Ralith: Meh, this seems fine. People can PR additional docs if they like. Maybe even omit the Ubuntu link; I bet the loader and most drivers are shipped by default there, and a novice user might not be sure what to do with this info.

README.md
```diff
@@ -204,7 +204,11 @@ Make sure that you have a Vulkan ready driver and install the [LunarG Vulkan SDK
 
 #### Linux
 
-Make sure that you have a Vulkan ready driver and install the [LunarG Vulkan SDK](https://lunarg.com/vulkan-sdk/). You also have to add the library and layers to your path. Have a look at my [post](http://askubuntu.com/a/803110/77183) if you are unsure how to do that.
+Install a Vulkan driver for your graphics hardware of choice, and the [LunarG Vulkan SDK](https://lunarg.com/vulkan-sdk/) for the [ICD](https://github.com/KhronosGroup/Vulkan-Loader) and [Validation Layers](https://github.com/KhronosGroup/Vulkan-ValidationLayers).
```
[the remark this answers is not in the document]
codecnotsupported: `Installing the LunarG SDK manually is not usually required.` 
`Linux users should typically not "install the SDK", especially not via the provided link; they should install specific distro-provided packages.`
I think I understand the reasoning behind this but,
I don't think all the tools from `LunarG SDK` are available as a package. (I couldn't find vkconfig on Ubuntu)
More over installing the LunarG SDK on linux is can be as simple as download, extract, run `source setup-env.sh`.
Or more to persistently, `echo 'source path_to_sdk/setup-env.sh'`
All it does it setup some environment variables which will work on most distributions that don't have very exotic configurations.

Unless I'm missing something, should that not be the first option since it'll probably work on most distributions?

README.md
i509VCB: Maybe a comment along the lines of "For other distros, check your package manager"?

README.md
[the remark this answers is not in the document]
MarijnS95: > Unless I'm missing something, should that not be the first option since it'll probably work on most distributions?

It actually won't, using a package manager is always preferable (proper dependency resolution, package tailored to distro-specifics, automatic updates, ...).

Ralith (approved):

PR #845, opened 2023-12-04
generator: Apply `must_use` attributes to all Vulkan structs

MarijnS95:
All structs are marked as `Copy`, and builders move `self` for a convenient builder pattern directly on `default()` (using `&mut` requires first keeping the instance alive in a `let` binding).  This however leads to builder functions accidentally moving a copy of `self`, mutating it, and dropping the result directly when the caller did not consume the returned value in ad-hoc field updates, in turn leaving the author confused why their code compiles without warnings while the struct was not updated.

Annotating all Vulkan structs with `#[must_use]` should at least give them an idea why, especially because of Rust allowing us to set a custom message.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1986,14 +1986,14 @@ fn derive_setters(
                 assert!(field.null_terminate);
                 assert_eq!(field.size, None);
                 return Some(quote! {
-                    #[inline]
                     #deprecated
+                    #[inline]
                     pub fn #param_ident_short(mut self, #param_ident_short: &'a core::ffi::CStr) -> Self {
                         self.#param_ident = #param_ident_short.as_ptr();
                         self
                     }
-                    #[inline]
                     #deprecated
+                    #[inline]
                     pub unsafe fn #param_ident_as_c_str(&self) -> &core::ffi::CStr {
                         core::ffi::CStr::from_ptr(self.#param_ident)
                     }
@@ -2001,13 +2001,13 @@ fn derive_setters(
             } else if is_static_array(field) {
                 assert_eq!(field.size, None);
                 return Some(quote! {
-                    #[inline]
                     #deprecated
+                    #[inline]
                     pub fn #param_ident_short(mut self, #param_ident_short: &core::ffi::CStr) -> core::result::Result<Self, CStrTooLargeForStaticArray> {
                         write_c_str_slice_with_nul(&mut self.#param_ident, #param_ident_short).map(|()| self)
                     }
-                    #[inline]
                     #deprecated
+                    #[inline]
                     pub fn #param_ident_as_c_str(&self) -> core::result::Result<&core::ffi::CStr, core::ffi::FromBytesUntilNulError> {
                         wrap_c_str_slice_until_nul(&self.#param_ident)
                     }
@@ -2075,8 +2075,8 @@ fn derive_setters(
                 let mutable = if field.is_const { quote!() } else { quote!(mut) };
 
                 return Some(quote! {
-                    #[inline]
                     #deprecated
+                    #[inline]
                     pub fn #param_ident_short(mut self, #param_ident_short: &'a #mutable #slice_param_ty_tokens) -> Self {
                         #set_size_stmt
                         self.#param_ident = #param_ident_short #ptr;
@@ -2088,8 +2088,8 @@ fn derive_setters(
 
         if field.basetype == "VkBool32" {
             return Some(quote!{
-                #[inline]
                 #deprecated
+                #[inline]
                 pub fn #param_ident_short(mut self, #param_ident_short: bool) -> Self {
                     self.#param_ident = #param_ident_short.into();
                     self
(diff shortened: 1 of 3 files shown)


Ralith:
> using &mut requires first storing the builder on the heap

I think you mean stack, but yeah.

generator/src/lib.rs
```diff
@@ -1986,28 +1988,30 @@ fn derive_setters(
...
                         core::ffi::CStr::from_ptr(self.#param_ident)
                     }
                 });
             } else if is_static_array(field) {
                 assert_eq!(field.size, None);
                 return Some(quote! {
-                    #[inline]
                     #deprecated
+                    #[inline]
                     pub fn #param_ident_short(mut self, #param_ident_short: &core::ffi::CStr) -> core::result::Result<Self, CStrTooLargeForStaticArray> {
```
Ralith, line 2010: Shouldn't this have it too?

Ralith (approved):

generator/src/lib.rs
MarijnS95: Not sure if we can annotate `Self` like that so that `.some_string(CStr...)?;` still counts as dropping `Self` by accident...

But then the same applies to a lacking "`must_use` pattern" on the previously discussed `(Vec<vk::Pipeline>, vk::?Result)` returns...

MarijnS95: I have contemplated whether such an annotation on the `struct` as a whole might appear too heavy. The standard library doesn't mark many types as such either (But in this case the only usefulness is the returned value, so it makes sense to annotate it as such).

Ralith: It seems like it'll work well for us, and it's low risk.

MarijnS95:
Reworded the commit message to mention `let` binding :)

generator/src/lib.rs
```diff
@@ -2407,6 +2407,7 @@ pub fn generate_struct(
         #dbg_str
         #[derive(Copy, Clone, #default_str #manual_derive_tokens)]
         #[doc = #khronos_link]
+        #[must_use = "Vulkan structs are moved by builders"]
```
MarijnS95, line 2410: Thoughts on the message?

generator/src/lib.rs
```diff
@@ -1986,28 +1986,28 @@ fn derive_setters(
                 assert!(field.null_terminate);
                 assert_eq!(field.size, None);
                 return Some(quote! {
-                    #[inline]
                     #deprecated
+                    #[inline]
```
MarijnS95, line 1990: Alphabetical sorting :)

generator/src/lib.rs
Ralith: I think talking about "moved" for `Copy` types is confusing. How about "Builders create new updated structs"?

Even then, referring to builders will be confusing when this inevitably comes up in other contexts. Maybe the best thing is no custom message?

MarijnS95: > Even then, referring to builders will be confusing when this inevitably comes up in other contexts.

That is why I was hesitant to move this to the `struct` definition in the first place; I haven't come across such a case yet but would love to know when it comes up and if the `must_use` makes sense there at all.

Your suggestion is much better though, I would even consider `Builders create new updated Vulkan structs`, maybe even add `... that must be consumed`. And then find different wording for `must`, since it's more of a suggestion that _not_ using the resulting object indicates an accidental no-op.

Ralith: > That is why I was hesitant to move this to the struct definition in the first place; I haven't come across such a case yet but would love to know when it comes up and if the must_use makes sense there at all.

I can think of lots of reasons why you might return a Vulkan struct from user code. For example:
- Abstracting out a common pattern
- Decoding serialized information into a Vulkan format (e.g. `VkSamplerCreateInfo`)
- Yielding memory barriers for accumulation elsewhere

 `must_use` is desirable in all of these cases, though it'll be a bit odd to see messages about builders for most of them, or discussion of "updated structs" when an API is creating the struct internally to begin with.

Incidentally, I think "builders" might be poor word choice regardless, since we've ditched the idea of specific objects that could be named as such. Maybe "helper functions"?

> I would even consider Builders create new updated Vulkan structs, maybe even add ... that must be consumed. And then find different wording for must, since it's more of a suggestion that not using the resulting object indicates an accidental no-op.

rustc already unconditionally says "... that must be used" here; I think that's sufficient.

MarijnS95: Apologies, I more so meant cases where `#[must_use]` could trigger but the user is thoughtfully ignoring the result. Maybe a function that performs an operation and returns a Vulkan struct that the user might optionally use for reference, but that seems particularly niche (e.g. `Option::get_or_insert_with()` style, where I might _not always_ want to use the returned value; but it fortunately returns a `&mut` anyway).

Will strip the message out, good points.

Ralith: Yeah, explicitly ignoring values in those corner cases makes sense.

Ralith (approved):

PR #846, opened 2023-12-05
device,instance: Provide `load_with()` constructor for get_proc_addr closure

MarijnS95:
While working on a GStreamer Vulkan interop example (where GStreamer-Vulkan opens the ICD and creates most objects for us, which need to be imported in an `ash::Instance` and `ash::Device`), it wasn't feasible to construct a `vk::EntryFnV1_0` and `vk::InstanceFnV1_0` with `extern` functions while keeping object data in some global static, especially `vk::InstanceFnV1_0` which contains many more functions that are not consumed by `Instance::load()`. GStreamer provides function loaders directly on its `GstVulkanInstance` and `GstVulkanDevice` which are desired to be used rather than attempting to open the same ICD and loading the same functions by hand.
The  original `Device::load()` and `Instance::load()` already create a closure internally, which is exactly what we need to expose to have a single callback that can hold the `&gst_vulkan::VulkanInstance/Device` state, and respond to a char-pointer name with a function pointer.

Note that this doesn't map very clearly to `Entry`, where the `load()` constructor is named `from_static_fn()` and a closure signature is equally lacking.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -2,8 +2,8 @@
 use crate::prelude::*;
 use crate::vk;
 use crate::RawPtr;
+use core::ffi;
 use std::mem;
-use std::os::raw::c_void;
 use std::ptr;
 
 /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDevice.html>
@@ -19,16 +19,24 @@ pub struct Device {
 
 impl Device {
     pub unsafe fn load(instance_fn: &vk::InstanceFnV1_0, device: vk::Device) -> Self {
-        let load_fn = |name: &std::ffi::CStr| {
-            mem::transmute((instance_fn.get_device_proc_addr)(device, name.as_ptr()))
-        };
+        Self::load_with(
+            |name: &std::ffi::CStr| {
+                mem::transmute((instance_fn.get_device_proc_addr)(device, name.as_ptr()))
+            },
+            device,
+        )
+    }
 
+    pub unsafe fn load_with(
+        mut load_fn: impl FnMut(&ffi::CStr) -> *const ffi::c_void,
+        device: vk::Device,
+    ) -> Self {
         Self::from_parts_1_3(
             device,
-            vk::DeviceFnV1_0::load(load_fn),
-            vk::DeviceFnV1_1::load(load_fn),
-            vk::DeviceFnV1_2::load(load_fn),
-            vk::DeviceFnV1_3::load(load_fn),
+            vk::DeviceFnV1_0::load(&mut load_fn),
+            vk::DeviceFnV1_1::load(&mut load_fn),
+            vk::DeviceFnV1_2::load(&mut load_fn),
+            vk::DeviceFnV1_3::load(&mut load_fn),
         )
     }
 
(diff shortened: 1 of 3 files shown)


ash/src/device.rs
```diff
@@ -18,11 +18,10 @@ pub struct Device {
 }
 
 impl Device {
-    pub unsafe fn load(instance_fn: &vk::InstanceFnV1_0, device: vk::Device) -> Self {
-        let load_fn = |name: &std::ffi::CStr| {
-            mem::transmute((instance_fn.get_device_proc_addr)(device, name.as_ptr()))
-        };
-
+    pub unsafe fn load_with(
+        load_fn: impl FnMut(&core::ffi::CStr) -> *const c_void + Copy,
```
Ralith, line 22: Let's avoid the `Copy` constraint by passing a unique reference down.

Ralith (changes requested):
Seems reasonable overall!

ash/src/device.rs
MarijnS95: By changing the signature of all the `fn load()`s? Sure, we can do that, there's unlikely to be any harm in that.

MarijnS95: Oh, I hadn't even seen that. For that reason it also wouldn't make sense for us to take `&mut impl FnMut`, as it requires the caller to use `&mut |name| { ... }`. In case they end up in a similar situation they can pass an `&mut` of _their_ closure as well.

Ralith (approved):
LGTM. It's not obvious to me why we couldn't do this for `Entry` as well.

MarijnS95:
> LGTM. It's not obvious to me why we couldn't do this for `Entry` as well.

I just updated the commit message with the reason, but didn't update the PR description: `Entry` uses or returns a `StaticFn` in various places. We either need to convert the closure to a `StaticFn` (hard because the closure may capture data) or strip out `StaticFn` altogether in favour of storing the closure (with a lifetime or `+ 'static`).

Alternatively we could use the closure - supposedly implemented as `vkGetInstanceProcAddr()` - to resolve itself. Since Vulkan 1.2 this use-case (resolving itself where `instance=NULL` should be supported):

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetInstanceProcAddr.html

> Starting with Vulkan 1.2, vkGetInstanceProcAddr can resolve itself with a NULL instance pointer.

ash/src/device.rs
```diff
@@ -19,16 +19,24 @@ pub struct Device {
...
+        Self::load_with(
+            |name: &std::ffi::CStr| {
+                mem::transmute((instance_fn.get_device_proc_addr)(device, name.as_ptr()))
+            },
+            device,
+        )
+    }
 
+    pub unsafe fn load_with(
+        mut load_fn: impl FnMut(&core::ffi::CStr) -> *const c_void,
```
Ralith, line 31: Making this `&mut dyn FnMut` could save downstream monomorphization on this cold path. Probably a marginal difference and slightly worse ergonomics, though.

Ralith: I think it'd have to be comparative benchmarking of code size and compile time.

In any case, we could construct a `&dyn mut FnMut` internally without breaking the interface introduced in this PR, thereby only having to monomorphize a trivial amount of code downstream. I'm unsure if that would actually work so long as the inner table `load`s are still generic (depends how smart rustc is), but we can stick with the nice ergonomic interface here regardless.

PR #858, opened 2024-01-10
generator: Derive slice getters/setters for runtime-bounded static arrays

MarijnS95:
Upstream Vulkan is okay with the request to annotate static arrays with a `len="structField"` annotation when the size of the static arrary is bounded by a value at runtime.  This allows us to generate more convenient builder functions that copy slices into the target "builder" struct while _also_ updating the length, rather than forcing the caller to move an array of the desired length with zeroed elements and setting the length field separately.

In addition provide a safe getter (and use it in the `Debug` implementation) to return a slice view containing only valid items per the length field.  As with strings this is only possible for static-sized arrays as we can never safely dereference a random pointer.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -804,33 +804,30 @@ impl FieldExt for vkxml::Field {
         assert!(!self.is_void());
         let ty = name_to_tokens(&self.basetype);
 
-        match self.array {
-            Some(vkxml::ArrayType::Static) => {
-                assert!(self.reference.is_none());
-                let size = self
-                    .size
-                    .as_ref()
-                    .or(self.size_enumref.as_ref())
-                    .expect("Should have size");
-                // Make sure we also rename the constant, that is
-                // used inside the static array
-                let size = convert_c_expression(size, &BTreeMap::new());
-                // arrays in c are always passed as a pointer
-                if is_ffi_param {
-                    quote!(*const [#ty; #size])
-                } else {
-                    quote!([#ty #type_lifetime; #size])
-                }
+        if is_static_array(self) {
+            assert!(self.reference.is_none());
+            let size = self
+                .size_enumref
+                .as_ref()
+                .or(self.size.as_ref())
+                .expect("Should have size");
+            // Make sure we also rename the constant, that is
+            // used inside the static array
+            let size = convert_c_expression(size, &BTreeMap::new());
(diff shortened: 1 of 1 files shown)


ash/src/vk/definitions.rs
```diff
@@ -1467,24 +1477,24 @@ impl ::std::default::Default for PhysicalDeviceMemoryProperties {
 }
 impl PhysicalDeviceMemoryProperties {
     #[inline]
-    pub fn memory_type_count(mut self, memory_type_count: u32) -> Self {
-        self.memory_type_count = memory_type_count;
+    pub fn memory_types(mut self, memory_types: &'_ [MemoryType]) -> Self {
```
MarijnS95, line 1480: I've modified `vk.xml` on this struct as an example, but I believe we can merge this without any generator output changes while making sure the upstream CI doesn't start failing when they make and merge this change :ok_hand:

Ralith: SGTM.

Ralith (commented):
Nice improvement!

ash/src/vk.rs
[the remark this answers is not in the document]
MarijnS95: https://github.com/ash-rs/ash/pull/859

Ralith, line 1: Still got a diff here.

ash/src/vk/definitions.rs
Ralith: Approving conditional on this diff being rolled back.

Ralith (approved):

ash/src/vk/definitions.rs
MarijnS95: CI would reject those diffs anyway :tada:

generator/src/lib.rs
```diff
@@ -2032,6 +2047,33 @@ fn derive_setters(
 
                 let mut slice_param_ty_tokens = field.safe_type_tokens(quote!('a), type_lifetime.clone(), None);
 
+                // vkxml wrongly annotates static arrays with a len= field is dynamic. These fields have a static length,
```
filnet, line 2050: should be "as dynamic" instead of "is dynamic" ?

MarijnS95: Neither; I'll fix this up in another PR.

PR #860, opened 2024-01-10
generator: Wrap `_as_c_str()` getter for possibly-pointers in `Option`

MarijnS95:
While this function is already marked `unsafe` to represent cases where an invalid pointer might be dereferenced, it should at least be obvious to the caller that there is a real chance for `NULL` pointers in these `CStr` getters, which will now be returned as `None`.  This function won't be used in `Debug` now as the dereference operation is still `unsafe`.

The `_as_c_str()` getters for static arrays is left untouched, as the data is read directly from the known-valid struct here.

### DIFF
--- a/ash/src/vk/definitions.rs
+++ b/ash/src/vk/definitions.rs
@@ -1062,8 +1062,12 @@ impl<'a> ApplicationInfo<'a> {
         self
     }
     #[inline]
-    pub unsafe fn application_name_as_c_str(&self) -> &core::ffi::CStr {
-        core::ffi::CStr::from_ptr(self.p_application_name)
+    pub unsafe fn application_name_as_c_str(&self) -> Option<&core::ffi::CStr> {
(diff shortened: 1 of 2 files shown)


ash/src/vk/definitions.rs
```diff
@@ -1454,6 +1461,17 @@ pub struct PhysicalDeviceMemoryProperties {
     pub memory_heap_count: u32,
     pub memory_heaps: [MemoryHeap; MAX_MEMORY_HEAPS],
 }
+#[cfg(feature = "debug")]
```
Ralith, line 1464: Unintended changes here and below?

Ralith (approved):

ash/src/vk/definitions.rs
MarijnS95: Looks like I stil had the `len=` annotation from testing #858

Ralith (approved):

PR #867, opened 2024-02-29
rewrite: Initial codegen 

Friz64:
This is based on #745 and eddyb's C parsing PR (https://github.com/Friz64/ash/pull/1). I will keep this a draft until the latter PR is merged.

I only implemented the most basic of struct outputs to have something to work with, extending that is the next step.

The directory structure used here is inspired by erupt. All items still get re-exported into `ash::vk` though.


stefnotch:
This looks very promising. 
On that note, I would be happy to assist if there's anything you need, from small testing tasks to implementing some tricky features.

Friz64:
Sure, the next step in my mind would be to _actually_ implement struct/union generation. For that we need lifetime propagation.

I think it would be convenient to be able to discuss this in some kind of group chat for ash development. I'd usually use Discord for this, but I'm open to suggestions as I'm not adamant on it.

MarijnS95:
@Friz64 there is a gitter instance for this repository. Link should be in the readme, it's using the old repo url because gitter couldn't migrate.

You can use various clients like Matrix to connect.

Friz64:
I've gone ahead and touched up this PR from the ground up.

The module structure is now extensions at the top level just like current ash, Vulkan feature levels at `vk1_0`, `vk1_1` etc just like erupt, and Vulkan Video at `video`. Not that it reaaally matters, cause the `vk` re-export is also still there ofc.

PR #874, opened 2024-03-13
Don't check MSRV for ash-examples

Friz64:
ash-examples has a lot of dependencies (mostly through winit), and so it's not unlikely for it to break the MSRV check. Here's the current CI error:
```
error: package `half v2.4.0` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.69.0
Either upgrade to rustc 1.70 or newer, or use
cargo update -p half@2.4.0 --precise ver
where `ver` is the latest version of `half` supporting rustc 1.69.0
```

It doesn't make much sense to have our MSRV influenced by it IMO.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,12 +11,12 @@ jobs:
       - run: cargo check --workspace --all-targets --all-features
 
   check_msrv:
-    name: Check ash and ash-window MSRV (1.69.0)
+    name: Check ash, ash-window and ash-rewrite MSRV (1.69.0)
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


MarijnS95 (approved):
https://github.com/ash-rs/ash/commit/1c9e42257752bb3b6a07e1a8d7eadf6280e99ec8 don't think there was a compelling reason back then, besides "it adheres to it now".

MarijnS95:
Turns out I intended to land this via #799 in c3d2dd1e759fc9ca3995d0b872611a49c77d6526 too :)

Issue #876, opened 2024-03-16
Merge function pointer table struct with high level wrapper struct


Friz64:
-- Using VK_KHR_swapchain as an example:

I think it's pretty confusing to have both `ash::extensions::khr::Swapchain` and `ash::vk::KhrSwapchainFn`. Is there a good justification for why the wrapper functions shouldn't be directly defined on the struct that also holds the function pointers?

I originally alluded to this in this comment: https://github.com/ash-rs/ash/pull/734#issuecomment-2001024938.

The Device handle would be a bit awkwardly grouped with the function pointers, but overall it results in an easier to grok API, especially after the changes in #734.

Ralith:
I'm weakly in favor of this, since very few users will care about the `*Fn` structs, I've seen them cause confusion/impair discoverability in the wild, and I can't think of a good reason to want a `*Fn` struct that isn't also satisfied by a unified struct.

One case that *might* call for a raw function table struct is storing wrapped functions in the implementation of a layer, in which storing the handle is (usually?) going to be redundant to the passed-in handle -- but even there, having access to the ergonomic wrappers is more valuable than saving a few bytes.

Ralith:
https://github.com/ash-rs/ash/pull/894 unifies the module hierarchies. It does not unify the structs, but since they're now clearly documented and immediately adjacent, confusion from those should be at least lessened.

Rua:
I hope the function pointer tables, and everything else in the `vk` module, remain available for those who just want to use the raw Vulkan API without any additional wrappers (e.g. Vulkano). I think in the past, someone proposed separating out the `vk` module into its own crate, containing only autogenerated code?

MarijnS95:
@Rua the proposed PR that will close this issue does exactly that: it moves the original types around to sit inside the same module, but otherwise keeps their original definition the same. You are still able to use the function pointer loaders without the `Instance`/`Device` wrappers.

PR #894, opened 2024-03-26
Unify extension module hierarchies

Ralith:
Fixes https://github.com/ash-rs/ash/issues/876.

As a bonus, this significantly reduces the amount of boilerplate required to wrap a new extension.


ash-examples/src/lib.rs
```diff
@@ -11,7 +11,7 @@ use std::{
     borrow::Cow, cell::RefCell, default::Default, error::Error, ffi, ops::Drop, os::raw::c_char,
 };
 
-use ash::extensions::{
+use ash::vk::{
```
MarijnS95, line 14: Let's merge this with the `ash` import below?

generator/src/lib.rs
```diff
@@ -2709,6 +2784,7 @@ pub fn generate_feature<'a>(
             &HashMap::new(),
             fn_cache,
             has_lifetimes,
+            "Raw Vulkan 1 static function pointers",
```
MarijnS95, line 2787: Do you think it makes sense to start looking into how to generate the same for `Entry`/`Instance`/`Device`? They share an even more complicated ambiguity with `ash::Device`, `ash::DeviceFn1_X` and `ash::vk::Device` (we have open issues/PRs about it). The version separation might make this even more complicated.

ash/src/lib.rs
```diff
@@ -70,8 +70,7 @@ pub mod util;
 pub mod vk;
 
 // macros of vk need to be defined beforehand
-/// Wrappers for Vulkan extensions
```
MarijnS95, line 73: We could keep an (updated) doc-comment for developer usage.

MarijnS95 (commented):
Looks like we're heading in the right direction to make Ash increasingly easy to use! Just wondering if we need to consider something similar for `Entry`/`Insteance`/`Device` now.

Don't forget a changelog entry!

generator/src/lib.rs
Ralith: I'm not sure that makes sense. `ash::vk::{Instance, Device}` are a very different kind of thing than `ash::Device`; it's not like the extensions where we had a function pointer table floating around in complete isolation.

Ralith:
Updated changelog.

generator/src/lib.rs
MarijnS95: Yeah, exactly. And that's my only gripe with the approach we're going for here: the split between `ash::` and `ash::vk::` used to be that the latter contains generated "low-level" `sys`-like bindings, and all the high-level stuff remains in `ash::{Entry,Instance,Device, extensions}`.

Changelog.md
```diff
@@ -76,6 +76,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - extensions/ext/ray_tracing_pipeline: Pass indirect SBT regions as single item reference (#829)
 - Replaced `c_char` array setters with `CStr` setters (#831)
 - `push_next()` functions now allow unsized `p_next` argument (#855)
+- Removed the redundant `ash::extensions` module hierarchy in favor of defining wrappers directly in the extension modules under `ash::vk` (#894)
```
MarijnS95, line 79: Does it make sense to write this below under `### Removed`?

ash-examples/src/lib.rs
```diff
@@ -11,11 +11,14 @@ use std::{
     borrow::Cow, cell::RefCell, default::Default, error::Error, ffi, ops::Drop, os::raw::c_char,
 };
 
-use ash::extensions::{
-    ext::debug_utils,
-    khr::{surface, swapchain},
+use ash::{
+    vk,
+    vk::{
```
MarijnS95, line 16: There's also (not sure if better):

```suggestion
    vk::{self,
```

ash/src/extensions/khr/create_render_pass2.rs
```diff
@@ -1,26 +1,11 @@
-//! <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_create_render_pass2.html>
+//! <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_create_renderpass2.html>
```
MarijnS95, line 1: Nice catch! I generated these from filenames, if you can please rename this one.

MarijnS95 (approved):
That's an amazing amount of repetitive hand-written code gone from the codebase! We should've done this much sooner, specifically these duplicated bits have been a continuous source of conflicts and compiler errors during the refactors of the past week.

---

This is also a direction I was hoping to head into with better generator support: at some point we can start autogenerating _some_ of these functions, and use both heuristics and config files to allow keeping manual implementations where a generator just doesn't understand the desired API.

ash-examples/src/lib.rs
Ralith: Eh, I kinda like that the syntax of the import matches the syntax of the use site.

Changelog.md
Ralith: I don't think so. Logically we haven't removed any functionality, we've just changed where it's exported from. Tweaked the wording to make that a bit less weird.

ash/src/extensions/khr/create_render_pass2.rs
Ralith: Done.

generator/src/lib.rs
Ralith: Specifically, while `vk::Device` exists and is important to preserve as a useful thing, extensions don't generally introduce new dispatchable handles.

I agree that this blurs the boundary a bit and I don't love that, but I think the reduction in confusion from not having duplicate extension module hierarchies is hard to ignore.

What if we go the other way? Flatten `vk::*` back down, but expose the function-pointer tables, `NAME` constants, and so forth exclusively under `ash::extensions`? I think this recovers the graceful separation while still preserving the wins here.

Changelog.md
MarijnS95: You are right, and I thought about the same just after placing this comment and leaving. Starting the sentence with `Removed` immediately makes it feel like that (or at least that portion of the) sentence belongs under `# Removed`. Thanks for clearing up the wording!

ash-examples/src/lib.rs
MarijnS95: That's also true. In our application this pattern is a common sign for inconsistencies with imports and qualifications:



```rust
use some_crate::{self}; // Does nothing, because some_crate is already in scope? Maybe to disambiguate?

use std::sync::{self, Arc}; // Probably inconsistent use of `Arc` in one place and `sync::Mutex` in another
```

generator/src/lib.rs
MarijnS95: > What if we go the other way? Flatten `vk::*` back down, but expose the function-pointer tables, `NAME` constants, and so forth exclusively under `ash::extensions`? I think this recovers the graceful separation while still preserving the wins here.

Should we merge this in, and toy around with that as a followup PR - or push it straight here? I think you're on to another good idea. `ash` in Rust "means" Vulkan bindings, so it doesn't mean much to have an other `vk` module level when the "previously made argument about a `sys` boundary" no longer really applies thanks to this change.

The only problem: how do we address the subsequent naming clash between `ash::Device` and `ash::vk::Device`? Besides renaming (to what?), any extra module hierarchy we might introduce for that (i.e. a theoretical but unlikable `mut wrappers`) introduces the same split we have for extensions before this PR.

MarijnS95 (approved):

generator/src/lib.rs
Ralith: Added a new commit that revises the module layout per our discussion this morning.

Ralith:
Revised approach: all core and extension raw type definitions and constants are exported directly under `ash::vk`. This includes e.g. extension name constants, which are named in the upstream style, less `VK_`. Extensions modules are exported under `ash::{khr, ext, ...}`, and include high-level wrappers, `*Fn` tables, and `NAME` and `SPEC_VERSION` reexports. Core `*Fn` structs are exported directly under `ash`.

This standardizes `ash::vk` as being purely unopinionated "sys" style bindings, while everything else is exposed under `ash`. We accept a small risk of a name collision between a future vendor tag and any other items we want to expose at the crate level.

generator/src/lib.rs
```diff
@@ -991,21 +997,14 @@ fn generate_function_pointers<'a>(
...
 
-            if let Some((vendor, ident)) = module {
-                if type_in_module.is_none() {
-                    fn_cache.insert(
-                        name,
-                        (format_ident!("{}", vendor.to_lowercase()), ident.clone()),
-                    );
-                }
+            if define_pfn {
+                fn_cache.insert(name);
```
MarijnS95, line 1003: Doesn't this shorten to `let define_pfn = fn_cache.insert(name);`?

generator/src/lib.rs
```diff
@@ -991,21 +997,14 @@ fn generate_function_pointers<'a>(
...
             }
 
             Command {
-                // PFN function pointers are global and can not have duplicates.
-                // This can happen because there are aliases to commands
```
MarijnS95, line 1007: We should probably keep tracking with a comment somewhere that because of aliases some loaders may overlap and load similar functions to others.

generator/src/lib.rs
```diff
@@ -3222,10 +3328,10 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
         .map(|ty| generate_aliases_of_types(ty, &required_types, &has_lifetimes, &mut ty_cache))
         .collect();
 
-    let feature_code: Vec<_> = features
+    let (feature_fp_code, feature_table_code): (Vec<_>, Vec<_>) = features
         .iter()
         .map(|feature| generate_feature(feature, &commands, &mut fn_cache, &has_lifetimes))
-        .collect();
+        .multiunzip();
```
MarijnS95, line 3334: Sounds like a regular unzip?

MarijnS95 (approved):
Lots of changes to separate code out of `extensions.rs`, but all for the better in the end!

generator/src/lib.rs
Ralith: Oops, forgot that existed.

generator/src/lib.rs
Ralith: Restored similar wording.

generator/src/lib.rs
Ralith: Nice catch!

Changelog.md
```diff
@@ -76,6 +76,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - extensions/ext/ray_tracing_pipeline: Pass indirect SBT regions as single item reference (#829)
 - Replaced `c_char` array setters with `CStr` setters (#831)
 - `push_next()` functions now allow unsized `p_next` argument (#855)
+- Flattened `ash::extensions` into `ash`, flattened `ash::vk::*` extension modules into `ash::vk`, and moved `*Fn` function pointer table structs from `ash::vk` into `ash` or the associated extension module (#894)
```
MarijnS95, line 79: >  flattened `ash::vk::*` extension modules into `ash::vk`

That seems slightly strange to mention as extension modules didn't exist (not even in `ash::vk::*`) in the latest release (they've only been introduced and resided there temporarily).

However, looks like we totally forgot to even mention that entire `mod`-refactor for extensions in #734.

Alas, nothing too bad, as I think we should write a proper actionable summary when publishing the `0.38` release.

generator/src/lib.rs
MarijnS95: Thanks! In hindsight it's not even (just) aliases if I remember correctly, but multiple `extensions` (including the base v1.x Vulkan core spec) `<require>`ing that `PFN_` type, and us naively emitting the `type PFN_` on the first occurrence while using a `HashSet` to prevent it from being emitted again.

PR #895, opened 2024-03-28
extension/ext: Add missing length-rounding in `cmd_set_sample_mask()` assert

Macdu:
As stated in the vulkan specification for `vkCmdSetSampleMaskEXT`: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdSetSampleMaskEXT.html

`sample_mask` must be an array of size `ceil(samples/32)`. However right now the ash wrapper for this function is checking that it is of size `floor(samples/32)`, making this wrapper unusable for any sample count less than 32.

### DIFF
--- a/ash/src/extensions/ext/extended_dynamic_state3.rs
+++ b/ash/src/extensions/ext/extended_dynamic_state3.rs
@@ -69,7 +69,7 @@ impl Device {
             samples.as_raw().is_power_of_two(),
             "Only one SampleCount bit must be set"
         );
-        assert_eq!(samples.as_raw() as usize / 32, sample_mask.len());
+        assert_eq!((samples.as_raw() as usize + 31) / 32, sample_mask.len());
         (self.fp.cmd_set_sample_mask_ext)(command_buffer, samples, sample_mask.as_ptr())
     }
 
--- a/ash/src/extensions/ext/shader_object.rs
+++ b/ash/src/extensions/ext/shader_object.rs
@@ -374,7 +374,7 @@ impl Device {
             samples.as_raw().is_power_of_two(),
             "Only one SampleCount bit must be set"
         );
-        assert_eq!(samples.as_raw() as usize / 32, sample_mask.len());
+        assert_eq!((samples.as_raw() as usize + 31) / 32, sample_mask.len());
         (self.fp.cmd_set_sample_mask_ext)(command_buffer, samples, sample_mask.as_ptr())
     }
 


MarijnS95 (approved):

Ralith (approved):

Macdu:
Ah, div_ceil was only stabilized in Rust 1.73. Do you want me to rewrite this function the classic way (`(samples.as_raw() + 31) / 32 as usize`), or let it be this way as I'm seeing you are planning to increase the MSRV anyway?

MarijnS95:
That MSRV bump has already happened, to 1.69 which the CI is enforcing. We don't have a strict (written down) MSRV policy other than trying to not bump too eagerly (i.e. at least 6 months) and keeping them to semver-breaking releases.

Having `div_ceil` over hand-written rounding in two instances isn't enough to warrant an MSRV bump. On the other hand the October 5th release of Rust 1.73 is nearing that same 6-month mark, and _could_ allow us to use more features added in those 4 releases in future patch-releases that we don't yet know about.

Macdu:
I see, I changed the content of this PR so it doesn't use `div_ceil` anymore.

MarijnS95:
We can make a separate PR to preemptively bump the MSRV if deemed necessary.

This PR also reminded me of an overwritten setter for `VkPipelineMultisampleStateCreateInfo::pSampleMask` in:

https://github.com/ash-rs/ash/blob/71387e942ff769cfefe824179c70b87e537f05ba/generator/src/lib.rs#L2014-L2031

This is to treat nullability properly (where Rust actually has a non-`NULL` pointer for `&[]` with length `0`) and we might use that same override to also set `rasterization_samples: vk::SampleCountFlags` with similar validation to these commands, but not a requirement.

Also, just to make sure, unlike `VkPipelineMultisampleStateCreateInfo`, neither of these two `cmd_set_sample_mask()` functions allow `pSampleMask` to be `NULL` to "treat the mask as if all bits were set to `1`".

Macdu:
> Also, just to make sure, unlike `VkPipelineMultisampleStateCreateInfo`, neither of these two `cmd_set_sample_mask()` functions allow `pSampleMask` to be `NULL` to "treat the mask as if all bits were set to `1`".

Yes, the fact that `pSampleMask` has slightly different semantics in `VkPipelineMultisampleStateCreateInfo` compared to `vkCmdSetSampleMaskEXT` is a bit counter-intuitive (although at least the validation layer should be able to detect easily any misuse).

PR #896, opened 2024-03-30
Outline monomorphic load functions

Ralith:
Seems to save about 5MiB (out of 77MiB total pre-change) in the debug build of ash-window's winit example. No obvious impact on compile time or release build size, probably because these only have one callsite each anyway in normal use. Still, seems like an easy win, might have more impact to folks who use `*Fn` directly, and it's nice to know we're not emitting a bunch of unnecessary IR.

No changelog entry as this is not a publicly-visible change.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1109,9 +1109,7 @@ fn generate_function_pointers<'a>(
     let loader = commands.is_empty().not().then(|| {
         quote! {
             impl #ident {
-                pub fn load<F>(mut f: F) -> Self
-                    where F: FnMut(&CStr) -> *const c_void
-                {
+                pub fn load<F: FnMut(&CStr) -> *const c_void>(mut f: F) -> Self {
                     Self::load_erased(&mut f)
                 }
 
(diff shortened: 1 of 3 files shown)


generator/src/lib.rs
```diff
@@ -1109,9 +1109,13 @@ fn generate_function_pointers<'a>(
     let loader = commands.is_empty().not().then(|| {
         quote! {
             impl #ident {
-                pub fn load<F>(mut _f: F) -> Self
+                pub fn load<F>(mut f: F) -> Self
                     where F: FnMut(&CStr) -> *const c_void
                 {
```
MarijnS95, line 1114: How bad would it be to shorten these lines a bit:

```suggestion
                pub fn load<F: FnMut(&CStr) -> *const c_void>(mut f: F) -> Self {
```

Or to look even more similar to the `dyn` below, since `F` is never referenced:

```suggestion
                pub fn load(mut f: impl FnMut(&CStr) -> *const c_void) -> Self {
```

MarijnS95 (approved):
Sounds like a simple but solid improvement! Now that the expensive monomorphization around `XXXFnXXX::load()` is gone, there should be no need to apply the same trick to the new `load_with()` function introduced in #846 I presume.

generator/src/lib.rs
Ralith: Went with the former, because the latter is a breaking change. Specifying `F` explicitly probably isn't all that useful, but no harm in preserving the capability.

PR #898, opened 2024-03-31
extensions/khr/calibrated_device: Use `self.handle` instead of requiring `vk::Device` argument

MarijnS95:
When introducing this new `VK_KHR_calibrated_timestamps` extension based on `VK_EXT_calibrated_timestamps` in #890 in parallel to finalizing and merging the `Instance`/`Device` separation in #734, I seem to have missed an opportunity to use the newly available `self.handle` for this extension from `struct Device`, instead leaving an unnecessary `device: vk::Device` argument in place.


Ralith (approved):
Nice catch!

PR #909, opened 2024-04-14
Change `push_next()` to `unsafe fn extend()` and add `push()` alternative

### LINKED ISSUE #905: `push_next` being safe to call is unsound
This code triggers UB using only safe code:
```rs
use ash::vk;

let create_info = vk::DeviceCreateInfo::default();
let mut features11 = vk::PhysicalDeviceVulkan11Features {
    p_next: 1 as _,
    ..vk::PhysicalDeviceVulkan11Features::default()
};
let create_info = create_info.push_next(&mut features11);
```
Personally I would prefer having a function that doesn't dereference raw pointers at all, because at least we I don't think need this functionality and it would alleviate having the unsafe block when calling `push_next`.

MarijnS95:
Fixes #905

`push_next()` is totally `unsafe` because of dereferencing a chain of `p_next` pointers to find the end of the chain to insert, which was obfuscated by a large `unsafe` block for the `BaseOutStructure` pointer cast in commit c8c8f69a ("`next` can contain a pointer chain and we need to correct insert it.").

While this function should definitely be marked unsafe, wrapping builders in `unsafe {}` en masse in user code isn't all too desirable, especially when this soundness issue only exists to optionally walk a `p_next` chain while most users are likely inserting bare structs without pointer chains most of the time.  `push()` is introduced for this reason, remaining safe to call without any unintended raw pointer dereferences.  `push_next()` has been renamed to `extend()` to more closely align with standard naming, and force existing users to notice this change and hopefully analyze their uses to decide which are safe/correct to switch over to safe `push()`.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `push()` method to all root structs to insert a single extension-struct in the pointer chain. (#909)
 - Update Vulkan-Headers to 1.3.296 (#910)
 - Added `VK_KHR_get_display_properties2` instance extension (#932)
 - Added `VK_EXT_metal_objects` device extension (#942)
@@ -24,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `get_pipeline_executable_properties()`;
     - `get_pipeline_executable_statistics()`.
   The expected length of this array can be queried with the respective `*_len()` variant of these functions.
+- `push_next()` has been renamed to `extend()` and marked as `unsafe`. Users are encouraged to call `push()` for singular structs instead. (#909)
 
 ## [0.38.0] - 2024-04-01
 
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -2320,34 +2320,61 @@ fn derive_getters_and_setters(
     let next_function = if let Some(next_member) = root_struct_next_field {
         let next_field = &next_member.vkxml_field;
         assert_eq!(next_field.basetype, "void");
-        let mutability = if next_field.is_const {
-            quote!(const)
-        } else {
-            quote!(mut)
-        };
         quote! {
             /// Prepends the given extension struct between the root and the first pointer. This
             /// method only exists on structs that can be passed to a function directly. Only
             /// valid extension structs can be pushed into the chain.
-            /// If the chain looks like `A -> B -> C`, and you call `x.push_next(&mut D)`, then the
+            /// If the chain looks like `A -> B -> C`, and you call `A.push(&mut D)`, then the
             /// chain will look like `A -> D -> B -> C`.
-            pub fn push_next<T: #extends_name + ?Sized>(mut self, next: &'a mut T) -> Self {
-                unsafe {
-                    let next_ptr = <*#mutability T>::cast(next);
-                    // `next` here can contain a pointer chain. This means that we must correctly
-                    // attach he head to the root and the tail to the rest of the chain
-                    // For example:
-                    //
-                    // next = A -> B
-                    // Before: `Root -> C -> D -> E`
-                    // After: `Root -> A -> B -> C -> D -> E`
-                    //                 ^^^^^^
-                    //                 next chain
-                    let last_next = ptr_chain_iter(next).last().unwrap();
-                    (*last_next).p_next = self.p_next as _;
-                    self.p_next = next_ptr;
-                }
+            ///
+            /// # Panics
+            /// If `next` contains a pointer chain of its own, this function will panic.  Call
+            /// `unsafe` [`Self::extend()`] to insert this chain instead.
+            pub fn push<T: #extends_name + ?Sized>(mut self, next: &'a mut T) -> Self {
+                // SAFETY: All implementors of T are required to have the `BaseOutStructure` layout
+                let next_base = unsafe { &mut *<*mut T>::cast::<BaseOutStructure<'a>>(next) };
+                // `next` here can contain a pointer chain.  This function refuses to insert the struct,
+                // in favour of calling unsafe extend().
+                assert!(next_base.p_next.is_null(), "push() expects a struct without an existing p_next pointer chain (equal to NULL)");
+                next_base.p_next = self.p_next as _;
+                self.p_next = <*mut T>::cast(next);
                 self
+            }
+
+            /// Prepends the given extension struct between the root and the first pointer. This
+            /// method only exists on structs that can be passed to a function directly. Only
+            /// valid extension structs can be pushed into the chain.
+            /// If the chain looks like `A -> B -> C` and `D -> E`, and you call `A.extend(&mut D)`,
+            /// then the chain will look like `A -> D -> E -> B -> C`.
+            ///
+            /// # Safety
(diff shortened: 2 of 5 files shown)


generator/src/lib.rs
```diff
@@ -2323,22 +2323,46 @@ fn derive_getters_and_setters(
...
+            }
+
+            /// Prepends the given extension struct between the root and the first pointer. This
+            /// method only exists on structs that can be passed to a function directly. Only
+            /// valid extension structs can be pushed into the chain.
+            /// If the chain looks like `A -> B -> C`, and you call `x.push_next(&mut D)`, then the
+            /// chain will look like `A -> D -> B -> C`.
+            ///
+            /// # Panics
+            /// If D optionally contains a pointer chain of its own, this function will panic.  Call
```
Ralith, line 2355:
```suggestion
            /// If D contains a pointer chain of its own, this function will panic.  Call
```

generator/src/lib.rs
```diff
@@ -2323,22 +2323,46 @@ fn derive_getters_and_setters(
...
+            /// `unsafe` [`Self::push_next()`] to insert this chain instead.
+            pub fn push_next_one<T: #extends_name + ?Sized>(mut self, next: &'a mut T) -> Self {
+                // SAFETY: All implementors of T are required to have the `BaseOutStructure` layout
+                let base = unsafe { &mut *<*mut T>::cast::<BaseOutStructure<'a>>(next) };
+                // `next` here can contain a pointer chain.  This function refuses to insert the struct,
+                // in favour of calling unsafe push_next().
+                // TODO: Docs say to write what we expect instead. I.e. "push_next_one() expects the inserted struct to have a NULL p_next pointer chain"
```
Ralith, line 2362: Seems like the message is already consistent with this?

Ralith (approved):

generator/src/lib.rs
MarijnS95: Edited the message without removing the `TODO`, not particularly clever of me today :)

MarijnS95:
Rebased and marked as ready for review, since we're now accepting breaking changes in light of other required breaking changes for `ash 0.39` with Vulkan 1.4.

`push_next_one()` didn't sound right to me (as in: push "the" next one), suggestions are welcome.

Ralith (approved):
LGTM semantically. I don't love either name, TBH. `push_one` feels like a weirdly verbose synonym of bare `push`, and both feel a bit obscure. `extend` might be apt for adding an extension struct, except `Vec::extend` inserts multiple items, so that's differently confusing.

Do we have any evidence that everyone's ever used the multi-item function of `push_next` on purpose? Asserting against that might not be the end of the world, though I guess it could be a really unpleasant surprise if someone supports environments they don't often test...

MarijnS95:
> I guess I hate plain `push` the least, if we need a new name. Being more concise is a decent consolation prize for the churn.

Good one! I forgot to mention, though recall that I originally used `push_next_one()` to have the `next` word explicitly in the function name to refer to `pNext` chains, which isn't in the new `push_one()` range.

Of course, if we stick with `push()` and add `next` to it for that reason, we end up with the already-existing `push_next()` function name again (see below).

On the other hand, since `pNext` chains are an integral part of Vulkan structures, `push()` in isolation should be clear enough to indicate these chains.

One final drawback: at that point `push_next()` no longer clearly distinguishes itself from `push()` as being the "unsafely deal with an existing pointer chain" variant, perhaps again what you're alluding to below.

> Do we have any evidence that everyone's ever used the multi-item function of `push_next` on purpose? Asserting against that might not be the end of the world, though I guess it could be a really unpleasant surprise if someone supports environments they don't often test...

Are you suggesting to change `push_next()` to be the "push only one item, or panic" variant? Or to rename it altogether in light of the above?

With this PR as written, users will either notice the change in the changelog, have to wrap `push_next()` in `unsafe` or don't even notice because they put all their Vulkan code in one big `unsafe {}` block altogether.

> `extend` might be apt for adding an extension struct, except `Vec::extend` inserts multiple items, so that's differently confusing.

Summarizing all the above, perhaps it is worth renaming `push_one()` to `push()`, and `push_next()` to `extend()`? That way everyone will have to change, and everyone will have to make a case-by-case decision which variant to switch to (or find-replace, run their app, and find where the `assert!()` goes off).

Ralith:
> On the other hand, since pNext chains are an integral part of Vulkan structures, push() in isolation should be clear enough to refer to these chains.

That's my thinking, yeah. Risk of collision with future setters seems low.

> Are you suggesting to change push_next() to be the "push only one item, or panic" variant?

Right, changing the existing behavior avoids the trouble of having to come up with a new name, but at some difficult-to-quantify risk of runtime surprises downstream.

> Summarizing all the above, perhaps it is worth renaming push_one() to push(), and push_next() to extend()?

That sounds like a great outcome to me. We end up with exactly the Rust container idioms, improved concision all around, and no breakage risk. We could even keep `push_next`, plus `unsafe`, a deprecation annotation, and hidden from docs, as an alias of `extend` for easier incremental migration.

MarijnS95:
Also, given `doc(hidden)` there's no real place to justify the change, and I feel like `#[deprecated]` is too short for that. IIRC it doesn't support intradoc links to refer to the new functions either.

Pushed it regardless, LMKWYT.

Changelog.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - `get_pipeline_executable_properties()`;
     - `get_pipeline_executable_statistics()`.
   The expected length of this array can be queried with the respective `*_len()` variant of these functions.
+- `push_next()` has been renamed to `extend()` and marked as `unsafe`. Users are encouraged to call `push()` for singular structs instead. (#909)
```
MarijnS95, line 27: Perhaps this should mention/describe that `push()` is new as well?

Either way, as with `0.38`, I'll write a migration guide that highlights a few features like this at the top of the README when releasing.

(Though could also write them now and submit it as part of this PR, let me know)

generator/src/lib.rs
```diff
@@ -2320,35 +2320,58 @@ fn derive_getters_and_setters(
...
+            /// If `next` contains a pointer chain of its own, this function will panic.  Call
+            /// `unsafe` [`Self::extend()`] to insert this chain instead.
+            pub fn push<T: #extends_name + ?Sized>(mut self, next: &'a mut T) -> Self {
+                // SAFETY: All implementors of T are required to have the `BaseOutStructure` layout
+                let next_base = unsafe { &mut *<*mut T>::cast::<BaseOutStructure<'a>>(next) };
+                // `next` here can contain a pointer chain.  This function refuses to insert the struct,
+                // in favour of calling unsafe extend().
+                assert!(next_base.p_next.is_null(), "push() expects a struct without an existing p_next pointer chain (equal to NULL)");
+                next_base.p_next = self.p_next as _;
+                self.p_next = <*mut T>::cast(next);
```
MarijnS95, line 2340: These casts are somewhat annoying and misleading, with the mani problem highlighted in https://github.com/ash-rs/ash/issues/970:

We treat every `next` as a `BaseOutStructure`, even if we should "technically" only be adding `BaseOutStructure`s to `p_next: *mut c_void` and `BaseInStructure`s to `p_next: *const c_void`.  At that point the cast only exists to change the type.

MarijnS95:
@Ralith also added some extra safety docs explaining writing to the pointer chain in `p_next`.

Ralith:
LGTM.

PR #910, opened 2024-04-14
Update Vulkan-Headers to 1.3.296

MarijnS95:


Ralith (approved):

ash/src/vk/extensions.rs
```diff
@@ -755,7 +755,6 @@ impl ColorSpaceKHR {
     pub const BT709_NONLINEAR_EXT: Self = Self(1_000_104_006);
     pub const BT2020_LINEAR_EXT: Self = Self(1_000_104_007);
     pub const HDR10_ST2084_EXT: Self = Self(1_000_104_008);
-    pub const DOLBYVISION_EXT: Self = Self(1_000_104_009);
```
MarijnS95, line 758: This was deprecated, but perhaps we should keep it if it isn't deprecated because of "invalid name"?

MarijnS95: I've refactored our `deprecated` handling a bit to _only_ skip `"aliased"` reasons, which is where we obviously cannot generate proper names. All the rest gets a `#[deprecated]` attribute.

MarijnS95:
Let's get this reviewed and merged. The `vk-parse` bump is required to unblock CI in upstream Khronos repos.

Ralith (approved):

PR #913, opened 2024-04-25
Add ExtensionMeta traits

Neo-Zhixing:
Supersedes #912

Create new marker traits `ExtensionMeta` as defined below:

 ```rs
pub enum PromotionStatus {
    None,
    PromotedToCore(u32),
    PromotedToExtension(&'static core::ffi::CStr),
}

pub trait ExtensionMeta {
    const NAME: &'static core::ffi::CStr;
    const SPEC_VERSION: u32;
    const PROMOTION_STATUS: PromotionStatus;

    type Device;
    fn new_device(instance: &crate::Instance, device: &crate::Device) -> Self::Device;
    type Instance;
    fn new_instance(entry: &crate::Entry, instance: &crate::Instance) -> Self::Instance;
}
```

Add marker struct `Meta` in all extension mod root. `ExtensionMeta` trait is automatically implemented for all `Meta` marker structs.

### Why is this useful?
Higher level abstractions may want to offer a way for users to enable extensions during application startup. Instead of asking for a `&'static CStr` extension name, we could instead ask for an `impl ExtensionMeta` which is significantly more idiomatic. This also gives us the promotion status of an extension, which allows the framework to behave differently depending on the core Vulkan version.

```rs
// During application startup
app.enable_extension<ash::khr::acceleration_structure::Meta>();
```

Unfortunately, we cannot implement traits for modules. That's why we need the extra `Meta` type that we added on module roots.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1264,13 +1264,18 @@ pub struct ExtensionCommands<'a> {
 }
 
 pub fn generate_extension_commands<'a>(
-    full_extension_name: &'a str,
-    items: &'a [vk_parse::ExtensionChild],
+    extension: &'a vk_parse::Extension,
     cmd_map: &CommandMap<'a>,
     cmd_aliases: &HashMap<&'a str, &'a str>,
     fn_cache: &mut HashSet<&'a str>,
     has_lifetimes: &HashSet<Ident>,
 ) -> ExtensionCommands<'a> {
+    let vk_parse::Extension {
+        name: full_extension_name,
+        children: items,
+        promotedto,
+        ..
+    } = extension;
     let byte_name_ident = Literal::byte_string(format!("{full_extension_name}\0").as_bytes());
 
     let extension_name = full_extension_name.strip_prefix("VK_").unwrap();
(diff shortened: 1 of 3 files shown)


Neo-Zhixing:
One additional thing we can do: Exposing the meta types directly on `ash::khr::*`. So instead of `ash::khr::acceleration_structure::Meta` you can write `ash::khr::AccelerationStructure`. And as long as the `ExtensionMeta` trait is in scope, you can also write `ash::khr::AccelerationStructure::Device`

generator/src/lib.rs
```diff
@@ -1264,13 +1264,18 @@ pub struct ExtensionCommands<'a> {
...
+    extension: &'a vk_parse::Extension,
     cmd_map: &CommandMap<'a>,
     cmd_aliases: &HashMap<&'a str, &'a str>,
     fn_cache: &mut HashSet<&'a str>,
     has_lifetimes: &HashSet<Ident>,
 ) -> ExtensionCommands<'a> {
+    let vk_parse::Extension {
+        name: full_extension_name,
+        children: items,
+        promotedto,
```
filnet, line 1276: `promoted_to` ?

Neo-Zhixing: This is defined in vk_parse. We can probably rename it to promoted_to in our code if that makes you happy

Neo-Zhixing:
@Ralith  @MarijnS95  any changes for this to be adopted?

PR #919, opened 2024-05-15
generator: Skip setters on `returnedonly="true"` structs

MarijnS95:
Vulkan annotates structs that are purely returned by the driver and do not have to be constructed by users (besides initializing `sType`/ `pNext`).  For these we can skip all builder functions (except `push_next()` and the `CStr`/slice getter helpers) and lighten our codebase somewhat.

Note that it is still possible to update the structure in a more low-level way by directly acessing the fields (or via FRU syntax). For Rust-based layer implementations this may be somewhat cumbersome, for which we could re-expose the setters behind a `cfg(feature = "returnedonly-setters")` of sorts?

### DIFF
--- a/ash/src/vk/definitions.rs
+++ b/ash/src/vk/definitions.rs
@@ -903,55 +903,8 @@ impl ::core::default::Default for PhysicalDeviceProperties {
 }
 impl PhysicalDeviceProperties {
     #[inline]
-    pub fn api_version(mut self, api_version: u32) -> Self {
-        self.api_version = api_version;
-        self
-    }
-    #[inline]
-    pub fn driver_version(mut self, driver_version: u32) -> Self {
-        self.driver_version = driver_version;
-        self
-    }
-    #[inline]
-    pub fn vendor_id(mut self, vendor_id: u32) -> Self {
-        self.vendor_id = vendor_id;
-        self
-    }
-    #[inline]
-    pub fn device_id(mut self, device_id: u32) -> Self {
-        self.device_id = device_id;
-        self
-    }
-    #[inline]
-    pub fn device_type(mut self, device_type: PhysicalDeviceType) -> Self {
-        self.device_type = device_type;
-        self
-    }
-    #[inline]
-    pub fn device_name(
-        mut self,
-        device_name: &CStr,
-    ) -> core::result::Result<Self, CStrTooLargeForStaticArray> {
-        write_c_str_slice_with_nul(&mut self.device_name, device_name).map(|()| self)
-    }
-    #[inline]
     pub fn device_name_as_c_str(&self) -> core::result::Result<&CStr, FromBytesUntilNulError> {
         wrap_c_str_slice_until_nul(&self.device_name)
-    }
-    #[inline]
-    pub fn pipeline_cache_uuid(mut self, pipeline_cache_uuid: [u8; UUID_SIZE]) -> Self {
-        self.pipeline_cache_uuid = pipeline_cache_uuid;
-        self
-    }
-    #[inline]
-    pub fn limits(mut self, limits: PhysicalDeviceLimits) -> Self {
-        self.limits = limits;
-        self
-    }
-    #[inline]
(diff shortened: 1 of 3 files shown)


MarijnS95:
Note also that `vk.xml` retroactively adds this attribute to existing structures from time to time, because it was initially forgotten. That would immediately break semver compatibility for us so we have to be careful with this if we want to bump Vulkan spec versions without breaking ash releases.

MarijnS95:
For completeness, note that upstream _does not_ set `returnedonly="true"` for structs that require the user to allocate storage, like https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSurfacePresentModeCompatibilityEXT.html / https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkLatencySurfaceCapabilitiesNV.html.

Ralith:
> For Rust-based layer implementations this may be somewhat cumbersome, for which we could re-expose the setters behind a cfg(feature = "returnedonly-setters") of sorts?

Sounds like a good idea to me. Perhaps `output-setters`?

> For completeness, note that upstream does not set returnedonly="true" for structs that require the user to allocate storage

Hmm, seems like this should probably be an attribute on individual fields rather than entire structs.

generator/src/lib.rs
```diff
@@ -2138,12 +2144,16 @@ fn derive_getters_and_setters(
...
             } else if is_static_array(field) {
                 assert_eq!(field.size, None);
-                return Some(quote! {
+                let setter = struct_.is_return.not().then(|| quote! {
                     #deprecated
                     #[inline]
                     pub fn #param_ident_short(mut self, #param_ident_short: &CStr) -> core::result::Result<Self, CStrTooLargeForStaticArray> {
                         write_c_str_slice_with_nul(&mut self.#param_ident, #param_ident_short).map(|()| self)
                     }
+                 });
```
filnet, line 2153: formatting issue ?

MarijnS95:
> Hmm, seems like this should probably be an attribute on individual fields rather than entire structs.

Perhaps that's something @oddhack can help us out with, but I can also report it upstream. A lot of `sType`/`pNext` structs could have this attribute moved to all other fields.

Otherwise a `caller-allocates` like annotation/feature might be useful separately as well. Caller-allocated arrays aren't that common in structs (typically only reside in function parameters) but are a necessity for extendability for extending functions via `pNext` chains.

generator/src/lib.rs
MarijnS95: This bunch of code is so large that `rustfmt` refuses to deal with it. I don't like fixing it up (incorrectly) manually either, we should figure out how to split it up into formattable functions.

ash/src/vk/definitions.rs
```diff
@@ -33351,36 +29671,14 @@ unsafe impl<'a> TaggedStructure for PipelineExecutableInternalRepresentationKHR<
...
-    pub fn is_text(mut self, is_text: bool) -> Self {
-        self.is_text = is_text.into();
-        self
-    }
-    #[inline]
-    pub fn data(mut self, data: &'a mut [u8]) -> Self {
-        self.data_size = data.len();
-        self.p_data = data.as_mut_ptr().cast();
-        self
-    }
```
MarijnS95, line 33383: Edge-case: this is an out-array that the caller would have to allocate, by calling [`vkGetPipelineExecutableInternalRepresentationsKHR()`](https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetPipelineExecutableInternalRepresentationsKHR.html) twice with this set to `NULL` initially, and set to a valid mutable slice on the second call.

MarijnS95: There appear to be 5 removed `&'a mut T`s in this PR: those all need to remain for this change o be convenient.

Ralith: Good catch. Can we just check for mutable ref arguments?

Also, you previously noted:

> note that upstream does not set returnedonly="true" for structs that require the user to allocate storage

Is that not true for these? Is the XML inconsistent here?

PR #923, opened 2024-06-08
Make bindings resilient to OOM errors

tomaka:
This PR proposes a relatively simple but maybe controversial change.

In the functions that return a `Vec`, instead of using `Vec::with_capacity` and similar, we use `Vec::new()` followed with `Vec::try_reserve`. If `try_reserve` fails, we return `VK_ERROR_OUT_OF_HOST_MEMORY`.

Note that there's no shorter way on stable Rust to do this. Notably, `try_with_capacity` is unstable.

If I'm not mistaken, the bindings now never panic in case of OOM.

I personally don't have a need for this change, but I thought that this was a neat thing to add.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -1547,7 +1547,10 @@ impl Device {
         &self,
         allocate_info: &vk::DescriptorSetAllocateInfo<'_>,
     ) -> VkResult<Vec<vk::DescriptorSet>> {
-        let mut desc_set = Vec::with_capacity(allocate_info.descriptor_set_count as usize);
+        let mut desc_set = Vec::new();
+        desc_set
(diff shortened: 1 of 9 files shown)


Ralith (approved):
Not sure how useful this is in practice, but it seems like a strict improvement, and not too bad to maintain.

MarijnS95:
If we end up doing this, is there a lint that enables us to catch accidental/new remaining calls to `Vec::with_capacity()` and friends, that may result in panicking allocations?

Alternatively we could list such functions in `disallowed-methods` in `clippy.toml` via [`clippy::disallowed_methods`](https://rust-lang.github.io/rust-clippy/master/#disallowed_methods).

> I personally don't have a need for this change, but I thought that this was a neat thing to add.

Might be neat, but if we do this we're better complete about it :)

PR #929, opened 2024-07-27
ash: Document `vk::Result` mapping whenever it is transformed on return

### LINKED ISSUE #928: Document the meaning of `bool` return values
The C signature of `vkGetFenceStatus` is the following (from the spec):
```c
// Provided by VK_VERSION_1_0
VkResult vkGetFenceStatus(
    VkDevice                                    device,
    VkFence                                     fence);
```

The Rust signature of `ash::Device::get_fence_status` is the following:
```rust
pub unsafe fn get_fence_status(&self, fence: Fence) -> VkResult<bool>
```

It's not obvious how the `VkResult` return value is converted into a `bool`, especially because the spec also defines the following for `VkResult`:
```c
// Provided by VK_VERSION_1_0
typedef enum VkResult {
    VK_SUCCESS = 0,
    VK_NOT_READY = 1,
...
```

Someone might expect that the VkResult is just casted into a bool, with non-zero values converted to `true`, and zero values converted to `false`. However, this is contrary to the behavior of the function, which makes `VK_SUCCESS` map to `true` and `VK_NOT_READY` map to `false`.

This behavior should be documented on Vulkan functions that return `bool`s in Rust but not in C to avoid confusion.

MarijnS95:
Fixes #928

In case one or more `VkResult` values returned by the underlying Vulkan function are mapped to a different return value inside `ash` for user convenience (i.e. whether a `swapchain` is suboptimal, or whether a `fence` or `event` was signaled), document explicitly what error codes from the upstream documentation result in what output values.

Note that "user convenience" here relates specifically to "success" scenarios that shouldn't be accidentally bubbled up via for example the `?` (`Try`) operator, but handled more clearly by providing a `bool` instead of forcing the user to match it out of `Err(vk::Result::*)`.

Documenting what `VkResult`s result in these `bool` return values should make it easier for the user to relate to the upstream documentation, as the naming of the Vulkan function never clearly reflects the expected output (i.e. `get_*status()` could have been named `is_*_signaled()` which is a more typical Rust convention), or whether `true` means "optimal" or "suboptimal" for the `Swapchain` interfaces.

### DIFF
--- a/ash/src/device.rs
+++ b/ash/src/device.rs
@@ -1051,9 +1051,11 @@ impl Device {
         .assume_init_on_success(event)
     }
 
-    /// Returns [`true`] if the event was set, and [`false`] if the event was reset, otherwise it will
-    /// return the error code.
     /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetEventStatus.html>
+    ///
+    /// # Returns
+    /// Returns [`true`] if the event is _signaled_ ([`vk::Result::EVENT_SET`]), [`false`] if the
+    /// event is _unsignaled_ ([`vk::Result::EVENT_RESET`]), or [`Err`] on failure.
     #[inline]
     pub unsafe fn get_event_status(&self, event: vk::Event) -> VkResult<bool> {
         let err_code = (self.device_fn_1_0.get_event_status)(self.handle(), event);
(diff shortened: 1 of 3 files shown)


Ralith (approved):

MarijnS95:
Changed `_not optimal_` wording to use `_suboptimal_` to match Vulkan wording. OTOH it doesn't **feel** like it stands out as much against `_optimal_` as `_not optimal_` used to do…

Ralith:
I actually almost suggested that because "not optimal: true" is kind of confusing due to the negation, even regardless of matching the wording.

PR #931, opened 2024-08-07
Work around `unused_qualifications` lint for Rust 1.80 prelude extension

MarijnS95:
`size_of(_val)()` was added to the prelude in Rust 1.80, causing `unused_qualifications` warnings whenever we import `mem` in scope and call it with `mem::` prefix.

The easiest workaround is to remove the prefix and explicitly import the function(s) in scope, which was already done in a few places.  We annotate the import with a `TODO` to remove it once bumping our MSRV on or past 1.80.

### DIFF
--- a/ash-examples/src/bin/texture.rs
+++ b/ash-examples/src/bin/texture.rs
@@ -5,6 +5,7 @@ use std::error::Error;
 use std::ffi;
 use std::io::Cursor;
 use std::mem;
+use std::mem::{size_of, size_of_val}; // TODO: Remove when bumping MSRV to 1.80
 use std::os::raw::c_void;
 
 use ash::util::*;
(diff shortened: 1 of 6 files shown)


MarijnS95:
Never mind, `size_of(_val)()` seems to have only been added to the prelude in 1.80: https://doc.rust-lang.org/src/std/prelude/common.rs.html#17-19. It explains my confusion as I don't recall accessing these functions without qualification before.

The documentation is to blame here. https://doc.rust-lang.org/std/prelude/v1/index.html specifically mentions:
> The first version of the prelude of The Rust Standard Library.
>
> See the [module-level documentation](https://doc.rust-lang.org/std/prelude/index.html) for more.

Followed by a list _including `size_of(_val)()`_.

And the module-level docs say:

> The first version of the prelude is used in Rust 2015 and Rust 2018, and lives in [std::prelude::v1](https://doc.rust-lang.org/std/prelude/v1/index.html). [std::prelude::rust_2015](https://doc.rust-lang.org/std/prelude/rust_2015/index.html) and [std::prelude::rust_2018](https://doc.rust-lang.org/std/prelude/rust_2018/index.html) re-export this prelude. It re-exports the following:

Falsely giving the sense that this has always been in the prelude.

PR #941, opened 2024-09-17
Fix sync validation errors

### LINKED ISSUE #917: ERROR: VALIDATION on examples
After getting setup with the VulkanSDK on macOS, I'm getting a lot of these errors at runtime when running either of the examples.

```
ERROR:
VALIDATION [VUID-vkAcquireNextImageKHR-semaphore-01779 (1461184347)] : Validation Error: [ VUID-vkAcquireNextImageKHR-semaphore-01779 ] Object 0: handle = 0xd5b26f0000000010, type = VK_OBJECT_TYPE_SEMAPHORE; | MessageID = 0x5717e75b | vkAcquireNextImageKHR():  Semaphore must not have any pending operations. The Vulkan spec states: If semaphore is not VK_NULL_HANDLE it must not have any uncompleted signal or wait operations pending (https://vulkan.lunarg.com/doc/view/1.3.280.0/mac/1.3-extensions/vkspec.html#VUID-vkAcquireNextImageKHR-semaphore-01779)
```

I haven't tested this on Linux so I can't say if it's my environment or something in ash itself, but I thought I'd open this issue since I don't see what else I should change without changing source code.

koplas:
Simplify the fence usage to avoid errors with synchronization.

Closes #917

### DIFF
--- a/ash-examples/src/bin/texture.rs
+++ b/ash-examples/src/bin/texture.rs
@@ -354,7 +354,7 @@ fn main() -> Result<(), Box<dyn Error>> {
         record_submit_commandbuffer(
             &base.device,
             base.setup_command_buffer,
-            base.setup_commands_reuse_fence,
+            base.submit_complete_fence,
             base.present_queue,
             &[],
             &[],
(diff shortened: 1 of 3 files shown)


koplas:
I had to remove the fence wait in `record_submit_commandbuffer`.
While inspecting the code I found that calling `record_submit_commandbuffer` is not safe, as there is a plethora of things that are not checked that could violate that valid usage (See for example https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkQueueSubmit.html#_description).

koplas:
Superseded by #1009

PR #943, opened 2024-09-19
extensions/amd: Add VK_AMD_anti_lag extension

MarijnS95:


Ralith (approved):

ash/src/extensions/amd/anti_lag.rs
```diff
@@ -0,0 +1,12 @@
+//! <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_AMD_anti_lag.html>
+
+use crate::vk;
+
+impl crate::amd::anti_lag::Device {
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkAntiLagUpdateAMD.html>
+    #[inline]
+    #[doc(alias = "vkAntiLagUpdateAMD")]
+    pub unsafe fn anti_lag_update(&self, data: &vk::AntiLagDataAMD<'_>) {
+        (self.fp.anti_lag_update_amd)(self.handle, data)
```
MarijnS95, line 10: https://registry.khronos.org/vulkan/specs/latest/man/html/vkAntiLagUpdateAMD.html docs are a bit vague, they say:

> If `pData` is not NULL...

Yet the respective `VUID-vkAntiLagUpdateAMD-pData-parameter` lacks the `` if `pData` is not `NULL`, `pData` must be a valid pointer to a valid... `` comment, as if it is not valid to leave this pointer `NULL` to begin with.

I'll leave it like this without `Option` and perhaps request a small spec update to remove the misleading `is not NULL` comment. CC @oddhack.

PR #944, opened 2024-09-19
extensions/khr: Add VK_KHR_pipeline_binary extension

MarijnS95:


MarijnS95:
Our `no_std` build complains:

```
error[E0412]: cannot find type `Vec` in this scope
  --> ash/src/extensions/khr/pipeline_binary.rs:66:19
   |
66 |     ) -> VkResult<Vec<u8>> {
   |                   ^^^ not found in this scope
   |
help: consider importing this struct
   |
3  | use alloc::vec::Vec;
   |
```

It's quite unfortunate that the `clippy:Lstd_instead_of_alloc` lint didn't warn that `Vec` from the `prelude` is in `std` (this lint only triggers if `use std::vec::Vec` is explicitly imported). CC @i509VCB who landed this feature, I haven't found an upstream tracking issue but maybe you know about this?

ash/src/extensions/khr/pipeline_binary.rs
```diff
@@ -0,0 +1,95 @@
...
+    #[inline]
+    #[doc(alias = "vkGetPipelineBinaryDataKHR")]
+    pub unsafe fn get_pipeline_binary_data(
+        &self,
+        info: &vk::PipelineBinaryDataInfoKHR<'_>,
+        pipeline_binary_key: &mut vk::PipelineBinaryKeyKHR<'_>,
+    ) -> VkResult<Vec<u8>> {
+        // TODO: This returns NOT_ENOUGH_SPACE_KHR, not INCOMPLETE (because nothing is written, instead of returning an impartial result)
+        read_into_uninitialized_vector(|count, data| {
+            (self.fp.get_pipeline_binary_data_khr)(
```
MarijnS95, line 70: Not sure what to do here. We could map `NOT_ENOUGH_SPACE_KHR` to `INCOMPLETE` to make `read_into_uninitialized_vector()` try again, though we could also be changing it at some point to return the incomplete but *partial* result to the user, which is invalid here.

If we don't, we can still use `read_into_uninitialized_vector()` to help us call `vkGetPipelineBinaryDataKHR()` twice. On `NOT_ENOUGH_SPACE_KHR` - which is unexpected - that error is bubbled up to the caller (and they wouldn't know how to handle it, since the wrapper is the one that queried the sizes initially).

MarijnS95: Mapped to `INCOMPLETE` and rewrote the documentation, also to reference the upstream issue (non-public repo).

@Ralith I think this is ready for merging unless you find anything glaring, and I'll follow up and clean up depending on how the upstream discussion pans out.

MarijnS95: I wrote a new variant of `read_into_uninitialized_binary_vector()` to deal with this.

ash/src/extensions/khr/pipeline_binary.rs
```diff
@@ -0,0 +1,93 @@
...
+        )
+    }
+
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPipelineKeyKHR.html>
+    #[inline]
+    #[doc(alias = "vkGetPipelineKeyKHR")]
+    pub unsafe fn get_pipeline_key(
+        &self,
+        pipeline_create_info: Option<&vk::PipelineCreateInfoKHR<'_>>,
+    ) -> VkResult<vk::PipelineBinaryKeyKHR<'_>> {
```
Ralith, line 49: Shouldn't `vk::PipelineBinaryKeyKHR` be an out parameter, to allow for extension? If this is for some reason undesired, then the lifetime should be `'static`.

ash/src/prelude.rs
```diff
@@ -64,6 +65,40 @@ where
     }
 }
 
+/// Calls `f` twice until it does not return [`vk::Result::ERROR_NOT_ENOUGH_SPACE_KHR`], ensuring all
+/// available binary data has been read into the vector.
```
Ralith, line 69: "twice until" is weird word choice. Maybe "at most twice"?

ash/src/prelude.rs
```diff
@@ -64,6 +65,40 @@ where
...
+) -> VkResult<Vec<u8>> {
+    let mut count = 4096;
+    let mut data = Vec::<u8>::with_capacity(count);
+    let mut err_code = f(&mut count, data.as_mut_ptr().cast());
+    if err_code == vk::Result::ERROR_NOT_ENOUGH_SPACE_KHR {
+        debug_assert!(
+            count > 4096,
+            "Implementation should have updated the value to be higher than the initial request"
+        );
+        err_code = f(&mut count, data.as_mut_ptr().cast());
```
Ralith, line 92: This seems to be missing the code to actually expand the `Vec`.

```suggestion
        data.reserve(count);
        err_code = f(&mut count, data.as_mut_ptr().cast());
```

ash/src/prelude.rs
```diff
@@ -64,6 +65,40 @@ where
...
+/// [`read_into_uninitialized_vector()`] instead.
+///
+/// See for example [`vkGetPipelineBinaryDataKHR()`], where the new return code was first introduced.
+///
+/// [`vkGetPipelineBinaryDataKHR()`]: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPipelineBinaryDataKHR.html
+pub(crate) unsafe fn read_into_uninitialized_binary_vector(
+    mut f: impl FnMut(&mut usize, *mut ffi::c_void) -> vk::Result,
+) -> VkResult<Vec<u8>> {
+    let mut count = 4096;
+    let mut data = Vec::<u8>::with_capacity(count);
```
Ralith, line 85: nit: `with_capacity` may allocate extra space, and we might as well use it
```suggestion
    let mut data = Vec::<u8>::with_capacity(4096);
    let mut count = data.capacity();
```
(and adjust the `debug_assert` accordingly)

ash/src/prelude.rs
```diff
@@ -64,6 +65,40 @@ where
...
+/// This function is _not_ designed to be used with [`vk::Result::INCOMPLETE`], see
+/// [`read_into_uninitialized_vector()`] instead.
+///
+/// See for example [`vkGetPipelineBinaryDataKHR()`], where the new return code was first introduced.
+///
+/// [`vkGetPipelineBinaryDataKHR()`]: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPipelineBinaryDataKHR.html
+pub(crate) unsafe fn read_into_uninitialized_binary_vector(
+    mut f: impl FnMut(&mut usize, *mut ffi::c_void) -> vk::Result,
+) -> VkResult<Vec<u8>> {
+    let mut count = 4096;
```
Ralith, line 84: Do we know that binaries are, in practice, usually smaller than this?

Ralith (changes requested):

ash/src/extensions/khr/pipeline_binary.rs
Ralith: > These calls could benefit from the same "maximum of 2 calls" strategy

What's the benefit here? Just hypothetical improved detection of implementation bugs? Seems like unnecessary churn.

ash/src/prelude.rs
MarijnS95: doing this, should we also handle `let mut count = capacity()` again? Either way shouldn't matter as the size is now adequate...

ash/src/lib.rs
```diff
@@ -269,6 +270,41 @@ where
     }
 }
 
+/// Calls `f` at most twice until it does not return [`vk::Result::ERROR_NOT_ENOUGH_SPACE_KHR`],
+/// ensuring all available binary data has been read into the vector.
+///
+/// The first call happens with a [`Vec`] of capacity `4096`.  If this is not adequate, `f` is
```
Ralith, line 276: Not really a blocker, but I'd like to see *some* sort of reference for why we think 4KiB in particular makes sense here rather than, say, 128B, or 1MiB.

ash/src/lib.rs
```diff
@@ -269,6 +270,41 @@ where
...
+) -> VkResult<Vec<u8>> {
+    let mut data = Vec::<u8>::with_capacity(4096);
+    let mut count = data.capacity();
+    let mut err_code = f(&mut count, data.as_mut_ptr().cast());
+    if err_code == vk::Result::ERROR_NOT_ENOUGH_SPACE_KHR {
+        debug_assert!(
+            count > data.capacity(),
+            "Implementation should have updated the value to be higher than the initial request"
+        );
+        data.reserve(count);
```
Ralith, line 297: `Vec::reserve` takes the number of *additional* bytes, so this will over-allocate. We should pass `count - data.capacity()` (probably via a temporary to keep borrowck happy).

Ralith, line 297: nit: We don't expect to reallocate this any further, so `Vec::reserve_exact` would be a little better.

ash/src/prelude.rs
Ralith: Yeah, no need.

Ralith (changes requested):

PR #946, opened 2024-09-26
extensions/ext: Add VK_EXT_device_generated_commands

MarijnS95:


Ralith (approved):

voxelbee:
I've been looking to use this extension and would like to have this merged into the Ash for usage it looks like we should be able to merge this without 1.4 support? [Should only need 1.3?](https://www.phoronix.com/news/Vulkan-1.3.296-Released)

Issue #343, opened 2020-12-12
Don't generate provisional extensions by default


MaikKlein:
Some extensions are marked as `provisional="true"`, and they are allowed to contain breaking changes. 

Because ash adheres follows semver like most rust crates, this isn't great.

We should either filter out those extensions or move them into an experimental folder where we don't adhere to semver.

Jasper-Bekkers:
It would be nice to be able to opt in to them at least, maybe through a cfg

MaikKlein:
> It would be nice to be able to opt in to them at least, maybe through a cfg

I mentioned this here https://github.com/MaikKlein/ash/issues/344 under feature flags

MarijnS95:
This is coming up again when `VK_AMDX_shader_enqueue` was eagerly wrapped in https://github.com/ash-rs/ash/pull/776, and is causing compatibility issues now that _allowed_ breaking changes are made to the extension, because it's marked provisional: https://github.com/KhronosGroup/Vulkan-Docs/pull/2442#issuecomment-2394979245

We should probably wrap the generated code as well as the relevant extension behind a feature flag or cfg flag, one that might allow us to make breaking changes to the code in nonbreaking releases or at the very least disable compiling these bits in i.e. Khronos CI.

PR #951, opened 2024-10-04
Update Vulkan-Headers to 1.4.350

MarijnS95:


MarijnS95:
With the `1.3.298` upgrade we now incorporated breaking changes from https://github.com/KhronosGroup/Vulkan-Docs/pull/2442 to a provisional extension.

I don't think we can address this in the current release, but I do think we can solve https://github.com/ash-rs/ash/issues/343 by adding a new [`provisional` or `experimental` feature flag](https://github.com/ash-rs/ash/pull/952) and document that if people opt-in to that feature, we might push breaking changes for extensions behind it - driven by upstream `vk.xml` changes in the same fashion.

(This is not "entirely" valid in Rust terms as I don't think feature flags "allow" you to opt out of semver compatibility even if you name them as such, and I'm also not sure how often we have to rely on it, but the current "just generate/wrap it anyway" approach does lock us out of making non-breaking releases with `Vulkan-Headers` updates)

ash/src/extensions_generated.rs
```diff
@@ -16198,8 +16242,6 @@ pub mod khr {
             pub cmd_pipeline_barrier2_khr: PFN_vkCmdPipelineBarrier2,
             pub cmd_write_timestamp2_khr: PFN_vkCmdWriteTimestamp2,
             pub queue_submit2_khr: PFN_vkQueueSubmit2,
-            pub cmd_write_buffer_marker2_amd: PFN_vkCmdWriteBufferMarker2AMD,
-            pub get_queue_checkpoint_data2_nv: PFN_vkGetQueueCheckpointData2NV,
```
MarijnS95, line 16202: Technically a semver-breaking change too. These functions were provided by `VK_AMD_buffer_marker` or `VK_NV_device_diagnostic_checkpoints` respectively, but `vk.xml` previously exposed them under the `VK_KHR_synchronization2` extension, and so did our generator.

ash/src/device.rs
```diff
@@ -64,6 +68,274 @@ impl Device {
...
+
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkUnmapMemory2.html>
+    #[inline]
+    #[doc(alias = "vkUnmapMemory2")]
+    pub unsafe fn unmap_memory2(
+        &self,
+        memory_unmap_info: &vk::MemoryUnmapInfo<'_>,
+    ) -> VkResult<()> {
+        (self.device_fn_1_4.unmap_memory2)(self.handle, memory_unmap_info).result()
+    }
```
MarijnS95, line 114: Of all the stabilized extensions, only `VK_KHR_map_memory2` was missing a wrapper implementation in Ash. I'm considering to retroactively add it.

MarijnS95:
Marking as draft to set expectations: this will be rebased/updated and merged after we've gone through all generator churn, which is important to get through first.

generator/src/lib.rs
```diff
@@ -976,6 +974,10 @@ fn generate_function_pointers<'a>(
             // must only emit a single definition.
             let define_pfn = fn_cache.insert(name.as_str());
 
+            let deprecated = deprecated_commands
+                .get(name.as_str())
+                .map(deprecated_annotation);
```
MarijnS95, line 979: I'm a bit on the fence about immediately marking commands deprecated on the Rust side following `vk.xml` annotations. They seem to be too harsh for what is intended to be a nudge towards newer Vulkan APIs that may not even be widely adopted and available yet.

0xbeefd1ed:
```
warning: use of deprecated method `ash::Instance::get_physical_device_queue_family_properties`: <https://docs.vulkan.org/spec/latest/appendices/legacy.html#legacy-gpdp2>                                                                                      
   --> src\physical_device.rs:150:18
    |
150 |                 .get_physical_device_queue_family_properties(self.0.physical_device)
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: use of deprecated method `ash::Instance::get_physical_device_memory_properties`: <https://docs.vulkan.org/spec/latest/appendices/legacy.html#legacy-gpdp2>
   --> src\physical_device.rs:177:51
    |
177 |         let memory_properties = unsafe { instance.get_physical_device_memory_properties(pdevice) };
    |                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: use of deprecated method `ash::Instance::get_physical_device_properties`: <https://docs.vulkan.org/spec/latest/appendices/legacy.html#legacy-gpdp2>
   --> src\physical_device.rs:178:52
    |
178 |         let pdevice_properties = unsafe { instance.get_physical_device_properties(pdevice) };
```

This is indeed causing a lot of noise for us. I don't think a deprecated API in Vulkan is equivalent to a deprecated API in Rust. A deprecated API in Rust typically implies that the API will be removed in a future version; a deprecated API in Vulkan still carries a lifetime warranty and will be supported forever.

PR #952, opened 2024-10-11
ash: Hide provisional extensions behind new `provisional` feature flag

### LINKED ISSUE #343: Don't generate provisional extensions by default

MarijnS95:
Fixes #343

Provisional extensions are susceptible to **breaking** API changes at any point (with accompanying `SPEC_VERSION` bump).  By hiding them behind a non-default feature flag, we can document that opting into this feature allows us to do breaking changes to these modules in non-breaking `ash` releases when driven by upstream `Vulkan-Headers` changes to `vk.xml`.

Moreover, upstream Khronos repositories that build `ash` in their CI can continue to build-test their changes without compile-testing any extension behind a `provisional` flag, which is especially useful if we've already wrapped a `provisonal` extension that is seeing breaking changes.

---

One important feature that is missing in this PR and the reason for its draft state is that we don't yet wrap any types `<require>`d by provisional changes.  While these types are unlikely to be useful without enabling the extension, non-breaking releases with breaking changes to such types should be considered invalid.  And there may be provisional extensions that solely consist of new types (extending existing types via `pNext`).

Secondly, feature flags may not technically exempt us from semver compatibility in releases, even if the feature flag is documented as such.  We could "do this anyway" for simplicity, or move to something more aggressive like `cfg` definitions.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -26,7 +26,7 @@ use quote::*;
 use regex::Regex;
 use std::{
     borrow::Cow,
-    collections::{BTreeMap, HashMap, HashSet},
+    collections::{hash_map::Entry, BTreeMap, HashMap, HashSet},
     fmt::Display,
     ops::Not,
     path::Path,
@@ -49,6 +49,7 @@ macro_rules! get_variant {
     };
     ($variant:path { $($member:ident),+ }) => {
         |enum_| match enum_ {
+            #[allow(clippy::double_parens, reason = "Occurs when only one member is queried")]
             $variant { $($member),+, .. } => Some(( $($member),+ )),
             _ => None,
         }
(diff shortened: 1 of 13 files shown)


oddhack:
BTW, I don't know that it impacts anything you're doing here, but just to note that in the C bindings, the provisional types and commands are kept in a separate 'beta' header, but the provisional enums added to existing enumerated types have to be visible in the core API header as there is no way to define them separately from the type they are part of.

MarijnS95:
They could remain in the same file - it's only a handful - and individual enum variants or constants can also be hidden behind a `#[cfg]` (somewhat similar to a `#if` preprocessor directive).

MarijnS95:
Working on this PR is really showing that the generators' lack of building definitions etc. top-down while being aware of their enabling feature(s)/extension(s) is such a hassle to build back in. It takes quite a while to push that information back in and is messy while at it.

MarijnS95:
@Ralith it looks to me like all items provided by these provisional extensions are accurately guarded now, if you agree we could merge.

generator/src/lib.rs
```diff
@@ -3218,17 +3445,25 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
     let vk_parse_definitions: Vec<_> = vk_parse_types
         .iter()
         .filter_map(|def| {
-            generate_definition_vk_parse(def, &required_types, &mut identifier_renames)
+            generate_definition_vk_parse(
+                def,
+                &ProvidedBy {
+                    names: vec!["TODO"],
```
Ralith: Are we missing some logic here? This TODO is showing up in definitions.rs doc comments. Fine to just not generate those comments, if that's easier, I think.

Ralith (approved):
LGTM aside from one doc issue. I don't see any gaps, but even if there were some this is a major strict improvement and shouldn't be blocked.

generator/src/lib.rs
MarijnS95: Yup yup, that was a pretty simple oversight.

However the new documentation additions on enum `impl` blocks (those defining variants on newtype enums) are still wrong, they're both incomplete and some of those blocks need to be split to account for different requirements between different variants. I'm inclined to just delete the docs again and address that in a separate PR as it's a bit more complicated to solve on the generator side and mostly unrelated to `provisional` features.

Ralith: Sounds like the right priorities to me.

MarijnS95:
Quite a few `Generated from` docs are wrong or at the very least pretty incomplete thanks to this terrible `const_cache` hashmap, but while this PR leans into it with the `ProvidedBy` it's not the right place to fix; will postpone.

PR #958, opened 2024-10-24
examples: Only use `image`'s `png` feature to speed up compile

Elabajaba:
The examples only needs the png feature, so disable default features and only add the png feature.

This saves ~4s (16s->12s) and reduces the built dependencies from 180 to 113 on Linux with a 5900x.


Ralith (approved):
Nice catch, thanks!

PR #960, opened 2024-11-13
generator: Demote `vk-parse` error-asserts to a warning message

### LINKED ISSUE #959: Ignoring unrecognized attributes (redux)
We are getting this ash-generate CI error in the internal vulkan spec repo again (see e.g. https://gitlab.khronos.org/vulkan/vulkan/-/jobs/658141), which spreads out to affect more branches as people sync up their working branches to main. They are just warnings, but unfortunate when they result in every CI job coming up with the warning label.

@MarijnS95 could you remind me where you are WRT ignoring unrecognized attributes? If not addressed this is going to keep happening as we do come up with new attributes reasonably often to convey more semantic information about API entities.

MarijnS95:
Fixes #959

When `vk-parse` encounters XML parsing errors (e.g. on unknown attributes or elements) it skips the element and emits an error in a list of errors that are returned to the caller.  We were originally never checking these leading to hard-to-debug issues, but eventually started asserting on this list being empty or otherwise panicking in #930.

Since `ash`'s generator is used in Khronos' upstream CI (in a fallible yet warning way), any new attribute would cause the generator to fail and subsequently require updates to `vk-parse` and `ash` before their CI is no longer flagged as "succeeded with warnings".  "Solve" this by once again allowing errors to exist, while still at least printing them to `stderr` for insights (that were previously missing...) to anyone running the generator.

Note @oddhack that this once again allows the generator to fail further in the chain when expected elements are missing, in which case an update is required either way.  If this happens often we can adjust `vk-parse` to be more lenient in still emitting errors but continuing to parse said element, if the new/unrecognized elements or attributes are considered harmless to the parsing process.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -3079,10 +3079,9 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
     use std::fs::File;
     use std::io::Write;
     let (spec2, errors) = vk_parse::parse_file(&vk_xml).expect("Invalid xml file");
-    assert!(
-        errors.is_empty(),
-        "vk_parse encountered one or more errors while parsing: {errors:?}"
-    );
+    if !errors.is_empty() {
+        eprintln!("vk_parse encountered one or more errors while parsing: {errors:?}")
+    }
     let extensions: Vec<&vk_parse::Extension> = spec2
         .0
         .iter()


oddhack:
Thanks. It's very rare that new attributes are not harmless to existing uses, exactly because there are downstream consumers that can take a very long time to update their XML use to account for it. Mostly they are additional metainformation for our own generators used to create spec artifacts and bits of CTS and validation layer code. Sometimes there are new tags allowed as well, same thing - for example \<feature> tags inside \<require> blocks are a recent addition.

MarijnS95:
Exactly, **if I remember correctly** that might have been one of the parser errors where a `<feature>` element inside a `<require>` element _may_ have skipped the entire `<require>` element, leading to a broken generator further down the line.

If that happens we should make `vk-parse` more lenient by skipping the unknown `<feature>` element but still parsing the rest of the content.

Ralith (approved):

PR #916, opened 2024-05-09
Implement vulkan video extensions

Cach30verfl0w:
Closes #722

Added support for the video extensions.

Issue #722, opened 2023-03-24
Safe wrapper for the video extension


pingpongcat:
Hello everyone, as a member of vulkan video working group under Khronos I would like contribute to this project by providing safe wrapper for the video extension. Spec for the encode is still in flux but I volunteer to ensure compatibility in future releases.

For I while Im have ash fork with added extensions:
https://github.com/neurotok/ash.git
And a basic encode example:
https://github.com/neurotok/ash-video.git

But the problem which I cannot overcome is a panic when calling even basic functions from the encode:
https://docs.rs/ash/latest/ash/vk/struct.KhrVideoQueueFn.html#
(using my ash-video example)

I had some success when I wrote this wrapper as an instance extension, but I'm running out of ideas on how the wrapper should be designed as device one. 

@MarijnS95 , @MaikKlein ? Could you please take a look at my ash fork and provided example ?

Ralith:
What is the panic message?

neurotok:
Hi @Ralith Im getting "thread 'main' panicked at 'Unable to load get_physical_device_video_capabilities_khr'"

This is caused by calling get_physical_device_video_capabilities in ash-video\src\main.rs:224

This function is implemented here ash/src/extensions/khr/video_encode_queue.rs:25

To run the examples you'll need nvidia beta drivers with video support

MarijnS95:
@neurotok Do you want to use this as tracking issue or can we close it now that the initial problem has been resolved (nothing to do with Ash it seems, just general Vulkan usage)?

Also keep in mind that the Video bindings are currently generated from the headers because they landed in an era where `video.xml` with bitfield struct members did not yet exist. We might change that in the future, though most likely with the generator rewrite and not in the current setup.

neurotok:
Hi @MarijnS95, of course you were right,  I already update m fork but let's keep this as a issue tracker for now please. At least until I have a working sample project with decoding.

Do you want some help with the generator to pull video extension from xml?

MarijnS95:
We already generate the full video extension bindings, don't need help for that but it'll likely change with the generator rewrite. In the event that you do need to make changes, use this tracking issue to discuss so first please.

MarijnS95:
@pingpongcat another user seems to have beat you to submitting the Video bindings in #916, though in a less mergeable state than your fork would have been (a year ago when it was proposed; we've made significant breaking/simplification changes to hand-written extension wrappers now).

[...]

PR #965, opened 2024-11-30
Add wrappers for `VK_KHR_video{,_encode,_decode}_queue`

colinmarc:
I've been using these via [internal wrappers](https://github.com/colinmarc/magic-mirror/blob/17e632ccbee15132e2420a5fc162c94171d4a34a/mm-server/src/vulkan/video.rs) for a good bit.

One downside (or is it a feature?) of the new `impl crate::khr::video_decode_queue::Instance` pattern is that I can't keep the wrappers in my project if I want to upgrade :)


MarijnS95:
Before looking, how is this supposed to compare to https://github.com/ash-rs/ash/pull/916? That PR has a not-insignificant amount of basic review comments, and has stalled since.

When a PR is opened, and is supposed to supersede another PR, I expect a little attribution documenting what this PR _does better_ than the old PR. I.e. more specifically, can I assume that all of the review comments there have been resolved in this PR, and won't need to be commented on again (e.g. because it was developed in isolation from that PR).

ash/src/extensions/khr/video_decode_queue.rs
```diff
@@ -0,0 +1,15 @@
...
+
+impl crate::khr::video_decode_queue::Device {
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdDecodeVideoKHR.html>
+    pub unsafe fn cmd_decode_video(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        decode_info: &vk::VideoDecodeInfoKHR<'_>,
+    ) {
+        (self.fp.cmd_decode_video_khr)(command_buffer, decode_info);
```
MarijnS95, line 13:
```suggestion
        (self.fp.cmd_decode_video_khr)(command_buffer, decode_info)
```

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,55 @@
...
+
+impl crate::khr::video_encode_queue::Device {
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdEncodeVideoKHR.html>
+    pub unsafe fn cmd_encode_video(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        encode_info: &vk::VideoEncodeInfoKHR<'_>,
+    ) {
+        (self.fp.cmd_encode_video_khr)(command_buffer, encode_info);
```
MarijnS95, line 14:
```suggestion
        (self.fp.cmd_encode_video_khr)(command_buffer, encode_info)
```

ash/src/extensions/khr/video_decode_queue.rs
```diff
@@ -0,0 +1,15 @@
+//! <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_video_decode_queue.html>
+
+use crate::vk;
+
+impl crate::khr::video_decode_queue::Device {
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdDecodeVideoKHR.html>
+    pub unsafe fn cmd_decode_video(
```
MarijnS95, line 8: Inline goes below the doc-comment, that makes it more uniform across the entire repo. And below `#[inline]`, I am trying to add `#[doc(alias = "vkCmdDecodeVideoKHR")]` too.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceVideoFormatPropertiesKHR.html>
+    pub unsafe fn get_physical_device_video_format_properties(
+        &self,
+        physical_device: vk::PhysicalDevice,
+        video_format_info: &vk::PhysicalDeviceVideoFormatInfoKHR<'_>,
+    ) -> VkResult<Vec<vk::VideoFormatPropertiesKHR<'_>>> {
+        read_into_defaulted_vector(|count, data| {
```
MarijnS95, line 173: Since the user could technically request more structures to be filled by use of the `pNext` chain in `VkVideoFormatPropertiesKHR`, this should be split between a `_len()` function and the actual getter that takes `&mut [vk::VideoFormatPropertiesKHR<'_>]`. There are various examples of this around the code-base to copy.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCreateVideoSessionKHR.html>
+    pub unsafe fn create_video_session(
+        &self,
+        create_info: &vk::VideoSessionCreateInfoKHR<'_>,
+        allocation_callbacks: Option<&vk::AllocationCallbacks<'_>>,
+    ) -> VkResult<vk::VideoSessionKHR> {
+        let mut video_session = mem::zeroed();
```
MarijnS95, line 62: We've always used `MaybeUninit` in these cases.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdBeginVideoCodingKHR.html>
+    pub unsafe fn cmd_begin_video_coding(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        begin_info: &vk::VideoBeginCodingInfoKHR<'_>,
+    ) {
+        (self.fp.cmd_begin_video_coding_khr)(command_buffer, begin_info);
```
MarijnS95, line 32:
```suggestion
        (self.fp.cmd_begin_video_coding_khr)(command_buffer, begin_info)
```

Let's drop all semicolons, please. That gives us a clear error when the return types don't match, in case a return type isn't annotated with `#[must_use]` already.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCreateVideoSessionParametersKHR.html>
+    pub unsafe fn create_video_session_parameters(
+        &self,
+        create_info: &vk::VideoSessionParametersCreateInfoKHR<'_>,
+        allocation_callbacks: Option<&vk::AllocationCallbacks<'_>>,
+    ) -> VkResult<vk::VideoSessionParametersKHR> {
+        let mut video_session_parameters = mem::zeroed();
```
MarijnS95, line 79: again `MaybeUninit`. Don't forget the `assume_init_on_success()` helper in place of `.result_with_success()`.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,182 @@
...
+        );
+    }
+
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetVideoSessionMemoryRequirementsKHR.html>
+    pub unsafe fn get_video_session_memory_requirements(
+        &self,
+        video_session: vk::VideoSessionKHR,
+    ) -> VkResult<Vec<vk::VideoSessionMemoryRequirementsKHR<'_>>> {
+        read_into_defaulted_vector(|count, data| {
```
MarijnS95, line 123: Same comment regarding `pNext` chains in "returned / written" fields.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,182 @@
+//! <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_video_queue.html>
+
+use crate::prelude::*;
+use crate::vk;
+use crate::RawPtr as _;
```
MarijnS95, line 5: I like importing trait _functionality_ with `as _`, but we should apply this pattern globally across the repo if we commit to it.

Is there a `clippy` lint that could help us with such a migration?

MarijnS95 (changes requested):
This PR is unfortunately forward in some places, but backwards in other places compared to !916. Please stick to the style used in other parts of the `ash` code-base, and take extra care when dealing with "returned" slices of structures containing `pNext` chains: the user should be able to set those before calling into the helper.

MarijnS95:
And don't forget a changelog entry :cold_sweat:

MarijnS95:
> One downside (or is it a feature?) of the new `impl crate::khr::video_decode_queue::Instance` pattern is that I can't keep the wrappers in my project if I want to upgrade :)

What is this supposed to mean? I don't think anything is "new" here besides the `struct Instance` being autogenerated, and having a restructured `mod`ule hierarchy. `impl` blocks for types were always allowed in arbitrary modules, as long as those `impl` blocks live within the same crate.

No clue what kind of upgrade you are anticipating: if these wrappers are merged to `ash`, and make it into a release, you can remove them from your project?

MarijnS95:
> I'm just pointing to the fact that the wrappers have to live inside `ash` if they want to follow that pattern - it's not possible to `impl` the definition outside in some other crate, as you correctly point out. My linked in-tree wrapper was using the old pattern, or at least I thought it was, and the change in pattern prompted me to PR this instead of keeping it in-tree. Please don't read too much into it, it was just an observation.

The pattern is identical. You could still define your own `struct VideoQueueExt`, using the same members _or using `ash::khr::video_queue::Device` as its sole member_ with a `Deref` to be able to call the wrapper functions inside `ash`.

Defining `impl` on foreign types has never been supported by Rust, and is not something that has been changed by whatever "new' extension pattern you are referring to.

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    #[inline]
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetEncodedVideoSessionParametersKHR.html>
+    pub unsafe fn get_encoded_video_session_parameters(
+        &self,
+        session_parameters_info: &vk::VideoEncodeSessionParametersGetInfoKHR<'_>,
+        info: &mut vk::VideoEncodeSessionParametersFeedbackInfoKHR<'_>,
+    ) -> VkResult<Vec<u8>> {
+        let ptr: *mut _ = info;
```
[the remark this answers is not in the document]
MarijnS95: Found it, I was doing the same in https://github.com/ash-rs/ash/pull/944.

ash/src/extensions/khr/video_decode_queue.rs
colinmarc: You put the alias above the `inline` in the only example I found, but it should probably be below the docstring, no?

ash/src/extensions/khr/video_queue.rs
colinmarc: I'll just note here that `VideoSessionMemoryRequirementsKHR` doesn't have any valid pNext members (but neither did `VideoFormatProperties` when I originally wrote this code).

ash/src/extensions/khr/video_decode_queue.rs
MarijnS95: "should" -> it "really doesn't matter", as long as there's one consistent ordering across the code-base to make (for example) `regex`es easier to work with.

It'd be great if `rustfmt` could deal with this for us.

ash/src/extensions/khr/video_queue.rs
MarijnS95: `VkVideoSessionMemoryRequirementsKHR` does have a `pNext` "member"/"field". It may not have any valid `structextends` structures _yet_, but that's the entire point of this API design: allowing extensibility in the _near_ future.

By not dealing with that now, we'll shoot ourselves in the foot and have to provide alternative functions or a semver-breaking release when extensions structures are added. We've had such "broken" abstractions (accidentally) in the past and I'd rather not have them again.

MarijnS95: Also note that this push-back is annoying and unnecessary.

ash/src/extensions/khr/video_encode_queue.rs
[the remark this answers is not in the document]
colinmarc: Should that be a separate PR?

ash/src/extensions/khr/video_queue.rs
colinmarc: Should I remove this or leave it? Or make a separate PR with the lint, and then rebase once that's merged?

ash/src/extensions/khr/video_queue.rs
colinmarc: Yup, I understand that! Not objecting to the change, just adding some context.

ash/src/extensions/khr/video_encode_queue.rs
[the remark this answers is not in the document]
MarijnS95: Feel free to include it here. I try to merge PRs in submission order (when there are no stalls) so #944 should provide this and it should naturally rebase out of this PR when I squash-merge it.

I hope you noticed an uptick in activity here in the `ash` crate: Vulkan 1.4 is on the brink of release (see public Vulkan repos) and I'd like to push out a new (unfortunately semver-breaking) `ash` release with it too, including all these open PRs.

ash/src/extensions/khr/video_queue.rs
colinmarc: Removed the `as _`, based on https://github.com/ash-rs/ash/pull/965#discussion_r1865861437.

ash/src/extensions/khr/video_decode_queue.rs
colinmarc: Adjusted the order to put the `#[doc(alias = "...")]` last.

colinmarc:
Ugh, I forgot that I split this into multiple commits last night - I've just been amending the last. I will clean it up, one sec.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// Call [`get_video_session_memory_requirements_len()`][Self::get_video_session_memory_requirements_len] to query the number of elements to pass to `out.`
+    /// Be sure to [`Default::default()`]-initialize these elements and optionally set their `p_next` pointer.
+    #[inline]
+    #[doc(alias = "vkGetVideoSessionMemoryRequirementsKHR")]
+    pub unsafe fn get_video_session_memory_requirements(
+        &self,
+        video_session: vk::VideoSessionKHR,
+        out: &mut [vk::VideoSessionMemoryRequirementsKHR<'_>],
+    ) -> VkResult<()> {
+        let mut req_count = out.len() as u32;
```
MarijnS95, line 154: Nit: I've been trying hard to name these all `count` across the `ash` code-base (for easy copy-pastability). Can we change it here and below?

MarijnS95 (approved):
> > Defining impl on foreign types has never been supported by Rust, and is not something that has been changed by whatever "new' extension pattern you are referring to.
> 
> I'm sorry if I stepped on your toes somehow - please consider the observation retracted.

Okay. If there's something wrong with the `ash` API, note that I do need to understand the issue and the scope of it in case there is something we are doing wrong or could do better. I still don't understand your request/observation, so I cannot help (other than refuting statements in hopes of better understanding the point of contention).

---

> To be completely honest, I glanced at #916, but it didn't seem far enough along to work off of.

That's the kind of information I'd like to see in a PR description. That way I know I don't have to directly correlate the review comments and confirm they were addressed as requested, for example. On the other hand, knowing that you would have looked at those review comments on a stale PR, and tried hard to preclude them from your PR in the first place, is a huge bonus.

> This code came from a wrapper internal to my project, which I wrote before #916 existed.

All fine if you wrote the code before #916 existed, just wanted to know if you updated/modified the code before making this PR which happened after #916 existed (and was hard to miss given that it was mentioned in #722).

> I understand that you don't want to review the ~same code twice, but if that's the case then feel free to just close this.

At least we actually made progress in this PR nice and quickly, and this looks all set to be merged before the next `ash` Release :tada: 

> Ugh, I forgot that I split this into multiple commits last night - I've just been amending the last. I will clean it up, one sec.

Don't worry, we'll probably end up squash-merging this as we do with any PR.

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,57 @@
...
+        (self.fp.cmd_encode_video_khr)(command_buffer, encode_info)
+    }
+
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetEncodedVideoSessionParametersKHR.html>
+    #[doc(alias = "vkGetEncodedVideoSessionParametersKHR")]
+    #[inline]
+    pub unsafe fn get_encoded_video_session_parameters(
+        &self,
+        session_parameters_info: &vk::VideoEncodeSessionParametersGetInfoKHR<'_>,
+        info: &mut vk::VideoEncodeSessionParametersFeedbackInfoKHR<'_>,
```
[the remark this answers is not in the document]
colinmarc: I pushed a version with a separate `_len` method for determining the length of the output slice, and optional `feedback_info` on both methods. It required a new `RawMutPtr` helper. (Happy to adjust if you want something different.)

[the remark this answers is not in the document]
MarijnS95: We discussed this function in chat. It is used to query three "distinct" things:

1. `pFeedbackInfo`;
2. `pDataSize`;
3. `pData`.

Of those, we identified two patterns that appear to be useful:

1. Query everything. This is already provided by the current implementation of this PR, and conveniently makes use of `read_into_uninitialized_vector()`. My only "optimization request" would be to pass `pFeedbackInfo` only during the first or second call, so that the driver doesn't have to (walk the optional `pNext` chain and) fill the struct(s) repeatedly;
2. Query `pFeedbackInfo` only, and optionally `pData`. This implies that the first call should provide `pDataSize` too, so that there is no third call needed for that (i.e. instead of adding a `pFeedbackInfo=NULL` variant based on `read_into_uninitialized_vector()`).

I didn't list this, but there is a third pattern that I am not sure is useful: query `pData` exclusively. This is identical to the ail of pattern 2., where a `read_into_uninitialized_vector()`-based variant would let `ash` perform the `pDataSize` query oblivious to the user.

---

Since this either spawns a lot of variants of the same function, or hides possible optimizations from the user (insofar returning preallocated `Vec`s isn't considered suboptimal...), we think having a `_len()` variant of this function - despite not needing to initialize the data - is the  solution. If both getters carry `feedback_info: Option<&mut ...>`, the caller can decide when and where to query this.

However, this implies that the array getter takes `&mut [MaybeUninit<u8>]` to not unnecessarily force the callers to (zero-)initialize a `&mut [u8]` when it is going to be overwritten anyway.

ash/src/lib.rs
```diff
@@ -91,6 +91,19 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
     }
 }
 
+pub trait RawMutPtr<T> {
+    unsafe fn as_raw_mut_ptr(self) -> *mut T;
+}
+
+impl<'r, T> RawMutPtr<T> for Option<&'r mut T> {
+    unsafe fn as_raw_mut_ptr(self) -> *mut T {
```
colinmarc, line 99: I just want to call out that this doesn't work if you don't consume `self`, because `&mut T` is not `Copy` (that would be aliasing). However, that makes this slightly inconsistent with regards to `RawPtr` (just above).

colinmarc:
> Okay. If there's something wrong with the ash API, note that I do need to understand the issue and the scope of it in case there is something we are doing wrong or could do better. I still don't understand your request/observation, so I cannot help (other than refuting statements in hopes of better understanding the point of contention).

Nothing is wrong with the `ash` API. I was being lazy and keeping the `Video*QueueExt` wrapper internal to my project rather than upstreaming it. The confusion stems from this:

> The pattern is identical. You could still define your own struct VideoQueueExt, using the same members or using ash::khr::video_queue::Device as its sole member with a Deref to be able to call the wrapper functions inside ash.

I didn't know that - I didn't have a deep understanding of what bits are codegen and what are manual and how they interact, and I had previously built my internal wrapper by copy/pasting ash code and tweaking the signatures. :) 

Because the struct definition moved into codegen and implementing foreign structs is illegal, it didn't work anymore to just copy/paste and tweak the signatures. Thus, the observation that the change made it impossible to write external wrappers. In actuality, it is still very possible, you just have to actually know how it works

Changelog.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_KHR_get_display_properties2` instance extension (#932)
 - Added `VK_EXT_metal_objects` device extension (#942)
 - Added `VK_AMD_anti_lag` device extension (#943)
+- Added `VK_KHR_video_queue`, `VK_KHR_video_encode_queue`, and `VK_KHR_video_decode_queue` device and instance extensions (#965)
```
MarijnS95, line 16: All of these extensions are still regarded as "device extensions" by the documentation (https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_video_queue.html, https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_video_encode_queue.html). The fact that the `PhysicalDevice` query exists on the `::Instance` is wrapper (using an instance function, rather than a device-specific function) doesn't make them instance extensions (but it does blur the lines/meaning somewhat).

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,84 @@
...
+            feedback_info.as_raw_mut_ptr(),
+            len.as_mut_ptr(),
+            ptr::null_mut(),
+        )
+        .assume_init_on_success(len)
+    }
+
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetEncodedVideoSessionParametersKHR.html>
+    ///
+    /// Call [`get_encoded_video_session_parameters_len`][Self::get_encoded_video_session_parameters] to determine the correct length of the output slice.
```
MarijnS95, line 41: To keep this documentation similar with the rest, and fix the link:

```suggestion
    /// Call [`get_encoded_video_session_parameters_len()`][Self::get_encoded_video_session_parameters_len()] to query the length of the byte slice to pass to `out`.
```

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[doc(alias = "vkCmdEncodeVideoKHR")]
+    pub unsafe fn cmd_encode_video(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        encode_info: &vk::VideoEncodeInfoKHR<'_>,
+    ) {
+        (self.fp.cmd_encode_video_khr)(command_buffer, encode_info)
+    }
+
+    /// Retrieves the length of the byte slice to pass to [`get_encoded_video_session_parameters`][Self::get_encoded_video_session_parameters].
```
MarijnS95, line 22: The Rust documentation seems to be written in a different form; if we wish to change that, we should do so consistently across the `ash` codebase otherwise.

```suggestion
    /// Retrieve the length of the byte slice to pass to [`get_encoded_video_session_parameters()`][Self::get_encoded_video_session_parameters()].
```

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetEncodedVideoSessionParametersKHR.html>
+    ///
+    /// Call [`get_encoded_video_session_parameters_len`][Self::get_encoded_video_session_parameters] to determine the correct length of the output slice.
+    #[inline]
+    #[doc(alias = "vkGetEncodedVideoSessionParametersKHR")]
+    pub unsafe fn get_encoded_video_session_parameters(
+        &self,
+        session_parameters_info: &vk::VideoEncodeSessionParametersGetInfoKHR<'_>,
+        feedback_info: Option<&mut vk::VideoEncodeSessionParametersFeedbackInfoKHR<'_>>,
+        out: &mut [u8],
```
MarijnS95, line 48: My gripe with this API was forcing the caller to initialize the byte-slice. `&mut [MaybeUninit<u8>]` seems more fitting, and it allows callers to pass [`Vec::spare_capacity_mut()`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.spare_capacity_mut) directly after reserving enough capacity.

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,84 @@
+//! <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_video_encode_queue.html>
+
+use core::mem;
+use core::ptr;
+
```
MarijnS95, line 5:
```suggestion
```

We've always let `rustfmt` sort these together. If we wish to diverge, we should do so across the codebase (at the very least your `video_queue.rs` :grimacing:).

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,84 @@
...
+        let mut len = out.len();
+        (self.fp.get_encoded_video_session_parameters_khr)(
+            self.handle,
+            session_parameters_info,
+            feedback_info.as_raw_mut_ptr(),
+            &mut len,
+            out.as_mut_ptr() as _,
+        )
+        .result()?;
+
```
MarijnS95, line 59:
```suggestion
```

ash/src/lib.rs
MarijnS95: I wonder if `trait RawPtr` should take `self` by value too, given that `Option<&T>` is `Copy` (and we don't have `&Option<&T>`).

Locally this violates `clippy::wrong_self_convention` though.

ash/src/lib.rs
```diff
@@ -91,6 +91,19 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+        match self {
+            Some(val) => val,
+            None => core::ptr::null_mut(),
```
MarijnS95, line 102: Let's make the above match arm `None =>` too instead of `_ =>`, and match the `::core` -> `core` convention too? We barely use that anywhere outside of generated code.

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,84 @@
...
+        (self.fp.cmd_encode_video_khr)(command_buffer, encode_info)
+    }
+
+    /// Retrieves the length of the byte slice to pass to [`get_encoded_video_session_parameters`][Self::get_encoded_video_session_parameters].
+    pub unsafe fn get_encoded_video_session_parameters_len(
+        &self,
+        session_parameters_info: &vk::VideoEncodeSessionParametersGetInfoKHR<'_>,
+        feedback_info: Option<&mut vk::VideoEncodeSessionParametersFeedbackInfoKHR<'_>>,
+    ) -> VkResult<usize> {
+        let mut len = mem::MaybeUninit::uninit();
```
MarijnS95, line 28: Perhaps `size` is more fitting. That is the same distinction the Vulkan spec uses (vs `count` for "number of elements").

ash/src/extensions/khr/video_encode_queue.rs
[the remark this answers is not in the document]
MarijnS95: @Ralith what is your thought on this, and the provided implementation/resolution?

ash/src/extensions/khr/video_encode_queue.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub unsafe fn cmd_encode_video(
+        &self,
+        command_buffer: vk::CommandBuffer,
+        encode_info: &vk::VideoEncodeInfoKHR<'_>,
+    ) {
+        (self.fp.cmd_encode_video_khr)(command_buffer, encode_info)
+    }
+
+    /// Retrieves the length of the byte slice to pass to [`get_encoded_video_session_parameters`][Self::get_encoded_video_session_parameters].
+    pub unsafe fn get_encoded_video_session_parameters_len(
```
MarijnS95, line 23: Looking at other implementations, we may have coincidentally forgotten `#[inline]` on some `_len()` functions?

ash/src/extensions/khr/video_encode_queue.rs
colinmarc: Looks like there's also a new helper for that: [MaybeUninit::uninit_array](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.uninit_array)

ash/src/extensions/khr/video_encode_queue.rs
colinmarc: Presumably wherever I copy/pasted from :)

ash/src/extensions/khr/video_encode_queue.rs
MarijnS95: I wouldn't call it new: I've been waiting _for years_ for those helpers to stabilize 😅. Likewise for `slice_assume_init_*()`.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+    ) -> VkResult<()> {
+        let mut count = out.len() as u32;
+        (self.fp.get_video_session_memory_requirements_khr)(
+            self.handle,
+            video_session,
+            &mut count,
+            out.as_mut_ptr(),
+        )
+        .result()?;
+
```
MarijnS95, line 162: Nit: as in the other file, let's drop this whitespace and below too.
```suggestion
```

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+    ) -> VkResult<()> {
+        let mut format_count = out.len() as u32;
+        (self.fp.get_physical_device_video_format_properties_khr)(
+            physical_device,
+            video_format_info,
+            &mut format_count,
+            out.as_mut_ptr(),
+        )
+        .result()?;
+
```
MarijnS95, line 240:
```suggestion
```

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// Be sure to [`Default::default()`]-initialize these elements and optionally set their `p_next` pointer.
+    #[inline]
+    #[doc(alias = "vkGetPhysicalDeviceVideoFormatPropertiesKHR")]
+    pub unsafe fn get_physical_device_video_format_properties(
+        &self,
+        physical_device: vk::PhysicalDevice,
+        video_format_info: &vk::PhysicalDeviceVideoFormatInfoKHR<'_>,
+        out: &mut [vk::VideoFormatPropertiesKHR<'_>],
+    ) -> VkResult<()> {
+        let mut format_count = out.len() as u32;
```
MarijnS95, line 232:
```suggestion
        let mut count = out.len() as u32;
```

ash/src/lib.rs
```diff
@@ -86,7 +86,20 @@ impl<'r, T> RawPtr<T> for Option<&'r T> {
...
+}
+
+pub trait RawMutPtr<T> {
+    unsafe fn as_raw_mut_ptr(self) -> *mut T;
+}
+
+impl<'r, T> RawMutPtr<T> for Option<&'r mut T> {
+    unsafe fn as_raw_mut_ptr(self) -> *mut T {
+        match self {
+            Some(val) => val,
```
MarijnS95, line 101: Nit: above we call this `inner`.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+            count.as_mut_ptr(),
+            ptr::null_mut(),
+        )
+        .assume_init_on_success(count)
+        .map(|c| c as usize)
+    }
+
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceVideoFormatPropertiesKHR.html>
+    ///
+    /// Call [`get_physical_device_video_format_properties_len()`][Self::get_physical_device_video_format_properties_len] to query the number of elements to pass to `out.`
```
MarijnS95, line 222:
```suggestion
    /// Call [`get_physical_device_video_format_properties_len()`][Self::get_physical_device_video_format_properties_len()] to query the number of elements to pass to `out`.
```

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+    ) -> VkResult<()> {
+        (self.fp.get_physical_device_video_capabilities_khr)(
+            physical_device,
+            video_profile,
+            capabilities,
+        )
+        .result()
+    }
+
+    /// Retrieve the number of elements to pass to [`get_physical_device_video_format_properties()`][Self::get_physical_device_video_format_properties],
```
MarijnS95, line 202:
```suggestion
    /// Retrieve the number of elements to pass to [`get_physical_device_video_format_properties()`][Self::get_physical_device_video_format_properties()].
```

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+        allocation_callbacks: Option<&vk::AllocationCallbacks<'_>>,
+    ) {
+        (self.fp.destroy_video_session_parameters_khr)(
+            self.handle,
+            video_session_parameters,
+            allocation_callbacks.as_raw_ptr(),
+        )
+    }
+
+    /// Retrieve the number of elements to pass to [`get_video_session_memory_requirements()`][Self::get_video_session_memory_requirements],
```
MarijnS95, line 126:
```suggestion
    /// Retrieve the number of elements to pass to [`get_video_session_memory_requirements()`][Self::get_video_session_memory_requirements()].
```

Though let's double-check other functions - I think none terminate this sentence.

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+            count.as_mut_ptr(),
+            ptr::null_mut(),
+        )
+        .assume_init_on_success(count)
+        .map(|c| c as usize)
+    }
+
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetVideoSessionMemoryRequirementsKHR.html>
+    ///
+    /// Call [`get_video_session_memory_requirements_len()`][Self::get_video_session_memory_requirements_len] to query the number of elements to pass to `out.`
```
MarijnS95, line 145:
```suggestion
    /// Call [`get_video_session_memory_requirements_len()`][Self::get_video_session_memory_requirements_len()] to query the number of elements to pass to `out`.
```

MarijnS95 (changes requested):
I hate giving repeated "changes requested" reviews like this (it's a common source of editor/submitter fatigue), but I just keep spotting new things that I either missed, or weren't there before the last pushes...

Happy to accept-changes / resolve them myself though! Seems to be mostly copy-paste typos while attempting to replicate constructs from other files.

ash/src/lib.rs
MarijnS95:
```suggestion
            Some(inner) => inner,
```

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+        let mut format_count = out.len() as u32;
+        (self.fp.get_physical_device_video_format_properties_khr)(
+            physical_device,
+            video_format_info,
+            &mut format_count,
+            out.as_mut_ptr(),
+        )
+        .result()?;
+
+        assert_eq!(format_count as usize, out.len());
```
MarijnS95, line 241:
```suggestion
        assert_eq!(count as usize, out.len());
```

ash/src/extensions/khr/video_queue.rs
```diff
@@ -0,0 +1,244 @@
...
+        &self,
+        physical_device: vk::PhysicalDevice,
+        video_format_info: &vk::PhysicalDeviceVideoFormatInfoKHR<'_>,
+        out: &mut [vk::VideoFormatPropertiesKHR<'_>],
+    ) -> VkResult<()> {
+        let mut format_count = out.len() as u32;
+        (self.fp.get_physical_device_video_format_properties_khr)(
+            physical_device,
+            video_format_info,
+            &mut format_count,
```
MarijnS95, line 236:
```suggestion
            &mut count,
```

MarijnS95 (approved):
I am happy with this, just need some more time to think about `get_encoded_video_session_parameters_len()`.

I guess that's where the `len` naming comes from. And as we discussed, this is the variant that could also be used to query _just `feedback_info`_ and optionally have the length around. That function name (in hindsight) sits a bit weird with me.

colinmarc:
> I hate giving repeated "changes requested" reviews like this (it's a common source of editor/submitter fatigue), but I just keep spotting new things that I either missed, or weren't there before the last pushes...
> 
> Happy to accept-changes / resolve them myself though! Seems to be mostly copy-paste typos while attempting to replicate constructs from other files.

I get the desire to have the codebase consistent, no worries. Sorry about all the copy pasta.

Are there plans to codegen more of this stuff? That would obviously help with the consistency.

MarijnS95:
> Are there plans to codegen more of this stuff? That would obviously help with the consistency.

There is (see open issues) but as you've hopefully noticed by the abundance of (requested) consistency yet _subtle differences_ all over the place, you probably understand that this is _hard_.

ash/src/extensions/khr/video_encode_queue.rs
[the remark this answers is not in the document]
MarijnS95: @Friz64 I believe your input is also valuable on the API of this function. See also my additional concerns in https://github.com/ash-rs/ash/pull/965#pullrequestreview-2485790116.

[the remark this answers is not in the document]
Ralith: Discussed this some more in chat. I don't have a ton of context for this API, or really a very strong opinion, but it seems like the current draft is close to a sweet spot, preserving consistency with two-call patterns elsewhere, not constraining callers, and not doing redundant work. I concur with @MarijnS95 that tweaking the non-`_len` version to return an `&mut [u8]` containing the initialized (sub)set of the input slice on success would be a nice convenient to add.

ash/src/lib.rs
colinmarc: Ah yeah, I missed that lint. I could change both to `to_...` or just do `RawMutPtr::to_raw_mut_ptr` and leave `RawPtr` as it is.

MarijnS95: Since both are private helpers, I'm not opposed to unifying both into a `to_`.

hgaiser:
I'm curious what the state is of this PR? Is it ready to be merged?

gitautas:
Also waiting for this to be merged.

Ralith:
@MarijnS95 has been out of touch on ash concerns for some months, and I don't personally have much time for reviewing nontrivial changes, so unfortunately this is probably blocked until the maintainership situation changes.

ash/src/extensions/khr/video_encode_queue.rs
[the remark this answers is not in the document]
MarijnS95: Re-looking into this, it seems we're covering most possible cases by always querying/passing `pSize`, and letting the user pick whether to query `pFeedbackInfo` (by setting the `Option`) and/or `pData` (by calling the non-`_len()` variant). Marking as resolved.

MarijnS95 (approved):

Ralith (approved):

PR #966, opened 2024-12-02
extensions: Remove `read_into_defaulted_vector()` to let caller pass `pNext`

MarijnS95:
Depends on #967

We've often discussed `pNext` chains in output arrays (i.e. #465, #588,  #744), and I always wondered why `read_into_defaulted_vector()` still existed: turns out this helper function was only hiding a few more remaining cases where the caller was previously _not_ able to manually extend the `pNext` chain to request arbitrary more structures to be filled with information.

Replace these remaining cases with a separate `_len()` getter, and have the main function take a caller-allocated `&mut [T]` slice where they can initialize the target struct including `pNext` pointer chains.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `VK_EXT_metal_objects` device extension (#942)
 - Added `VK_AMD_anti_lag` device extension (#943)
 
+### Changed
+
+- Enable passing mutable arrays of `pNext`-initialized structs for these remaining extension functions: (#966)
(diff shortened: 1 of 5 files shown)


MarijnS95:
Still need to do some code archaeology why this didn't get fixed in the past. We seem to have discussed it in #465 yet this helper made it in anyway (on a struct that has `pNext`...) and future PRs followed to reuse it in other extensions...

Ralith (approved):

MarijnS95:
Looks like the Rust 1.83 update finally hit us. I already started looking into omitting these lifetimes but it's "hard" because the builder pattern in the generator doesn't trivially know if it's going to be referencing one from the `impl` block, if i.e. we have a `pNext` chain without `structextends` referencing it for example. Let's see what can be done.

Issue #906, opened 2024-04-14
Extension structs have lifetimes even when they don't need one


Rua:
Extension structs do contain pointers, but their lifetime cannot ever be used if the only pointer they contain is `p_next`. Structs like this don't actually borrow anything, all structures in a `p_next` chain are borrowed by the parent. They have no `push_next` method, and there are no other methods that return `Self` with a different lifetime than the input struct. Technically you can set the lifetime using the `PhantomData` member, but that seems a bit silly and pointless.

Take `[ExternalMemoryImageCreateInfo](https://docs.rs/ash/latest/ash/vk/struct.ExternalMemoryImageCreateInfo.html)` for example. Its only pointer is `p_next`, it's not the parent struct, it doesn't borrow any data itself.

MarijnS95:
Cross-posting from https://github.com/ash-rs/ash/pull/969#issuecomment-2514806681: I don't think this issue is actionable if we don't want a breaking release any time such a struct starts being referenced in a new `structextends`. This is much less likely for `ExternalMemoryImageCreateInfo` given that it is already part of a chain defined by a parent struct, but very possible for "new" parent structs that are defined with `pNext` up-front with no extensions available right now, but possibility for extensions in future Vulkan releases.

Ralith:
Could we classify structs into "always root" or "always extension" and omit lifetimes from the latter? Not sure Khronos would respect that...

MarijnS95:
I doubt they would want to commit to that, but we could try.

I have no reason to pursue this (kept it out of the `0.39` milestone) as it's very easy to represent with `'static` lifetime in user code.

PR #969, opened 2024-12-03
generator: Elide lifetimes in `impl` blocks whenever they are redundant

MarijnS95:
Depends on (and is a partial alternative for) #967, depends on #968

Rust 1.83 suggests that whenever a named lifetime `'a` in an `impl<'a> (Trait for) Struct<'a>` is specified, but not used anywhere else within that `impl` block, the naming can be removed and replaced with `'_`.

While we can unconditionally remove this from blocks like `impl TaggedStructure`, more generator logic is necessary to optionally omit them on builder blocks with special care around optional presence of `push_next()`.

---

It is this complicated / error-prone generator logic that I am not sure warrants this feature.  We should measure if there's any compile-time effect before committing to this.  If there's not, I suggest we only handle trivial cases like removing it from `impl TaggedStructure for Struct<'_>`, perhaps via #967.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -920,7 +920,7 @@ fn generate_function_pointers<'a>(
     commands: &[&'a vk_parse::CommandDefinition],
     rename_commands: &HashMap<&'a str, &'a str>,
     fn_cache: &mut HashSet<&'a str>,
-    has_lifetimes: &HashSet<Ident>,
+    has_lifetimes: &HashMap<Ident, LifetimeInfo>,
     doc: &str,
 ) -> (TokenStream, TokenStream) {
     // Commands can have duplicates inside them because they are declared per features. But we only
@@ -966,7 +966,7 @@ fn generate_function_pointers<'a>(
                 .map(|param| {
                     let name = param.param_ident();
                     let type_lifetime = has_lifetimes
-                        .contains(&name_to_tokens(
+                        .contains_key(&name_to_tokens(
                             param.definition.type_name.as_ref().unwrap(),
                         ))
                         .then(|| quote!(<'_>));
@@ -1273,7 +1273,7 @@ pub fn generate_extension_commands<'a>(
     cmd_map: &CommandMap<'a>,
     cmd_aliases: &HashMap<&'a str, &'a str>,
     fn_cache: &mut HashSet<&'a str>,
-    has_lifetimes: &HashSet<Ident>,
+    has_lifetimes: &HashMap<Ident, LifetimeInfo>,
 ) -> ExtensionCommands<'a> {
     let byte_name_ident = Literal::byte_string(format!("{full_extension_name}\0").as_bytes());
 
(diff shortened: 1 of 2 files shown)


MarijnS95:
@Rua I've been considering pulling this further to solve your issue #906, but one major drawback is that non-breaking Vulkan releases _may_ add more structures that would reference `ExternalMemoryImageCreateInfo` in their `structextends` for example. In this case a `fn push_next()` would be retroactively added, but would also require a breaking change to add back the omitted lifetime to `ExternalMemoryImageCreateInfo`.

Ralith:
> It is this complicated / error-prone generator logic that I am not sure warrants this feature. We should measure if there's any compile-time effect before committing to this.

I am certain that rustc desugars the elided lifetimes to the explicit version immediately, so there should be no difference.

What's the motivation for this change? Normally elision is there to make reading and maintaining code easier in simple cases, but that doesn't seem very relevant to generated code.

MarijnS95:
Having shorter checked-in code is marginal: if you say it is desugared there's no runtime benefit at all.

I prefer fixing clippy linte rather than allowing them (in case future violations show up that are relevant to clean up, i.e. when generated code is unnecessarily verbose like `impl StructureType`). As mentioned, this added complexity doesn't seem to outweigh that benefit.

Ralith:
It's not terribly unusual for clippy lints to be overly opinionated or make assumptions that aren't justified in all cases. I think the cost of any additional complexity in the generator is quite high, so if there's no benefit to users I'd be inclined to stick with disabling the lint.

MarijnS95:
Agreed to definitely keep this complexity out of the generator, but I do tend to appreciate clippy pointing out shorter, more concise ways to write the same thing. It's just that we can't use it most of the time here.

generator/src/lib.rs
```diff
@@ -2620,7 +2632,11 @@ pub fn generate_handle(handle: &vkxml::Handle) -> Option<TokenStream> {
     };
     Some(tokens)
 }
-fn generate_funcptr(fnptr: &vkxml::FunctionPointer, has_lifetimes: &HashSet<Ident>) -> TokenStream {
+fn generate_funcptr(
+    fnptr: &vkxml::FunctionPointer,
+
```
filnet, line 2637: is the empty line on purpose ?

MarijnS95: It is not something that is relevant when discussing a proof-of-concept that already had only a 20% chance of being successful.

generator/src/lib.rs
```diff
@@ -2387,7 +2399,7 @@ fn derive_getters_and_setters(
 
         let value = variant_ident("VkStructureType", value);
         quote! {
-            unsafe impl #lifetime TaggedStructure for #name #lifetime {
+            unsafe impl TaggedStructure for #name #opaque_lifetime {
```
MarijnS95, line 2402: This is the only lifetime that we could reasonably and unconditionally elide, but I've opted not to.

PR #971, opened 2024-12-06
generator: Replace `Extends{Root}` with `Extends<Root>` trait

MarijnS95:
Works towards #879, yet without swapping the generic and the struct to implement it for.

---

Instead of emitting a new `trait` for every `Root` struct that is being implemented by one or more "child" structs (those that have `Root` in their `structextends`), create one trait that takes the root struct as a generic parameter, and implement that directly instead.

This not only saves on having to define the `trait` for every `Root` struct but also paves the way towards providing default trait implementations for any pair of root and child struct, such as the `p_next` builder methods.

### DIFF
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -2330,8 +2330,6 @@ fn derive_getters_and_setters(
         })
     });
 
-    let extends_name = format_ident!("Extends{}", name);
-
     // The `p_next` field should only be considered if this struct is also a root struct
     let root_struct_next_field = next_field.filter(|_| root_structs.contains(&name));
 
@@ -2349,7 +2347,7 @@ fn derive_getters_and_setters(
             /// # Panics
             /// If `next` contains a pointer chain of its own, this function will panic.  Call
             /// `unsafe` [`Self::extend()`] to insert this chain instead.
-            pub fn push<T: #extends_name + ?Sized>(mut self, next: &'a mut T) -> Self {
+            pub fn push<T: Extends<Self> + ?Sized>(mut self, next: &'a mut T) -> Self {
                 // SAFETY: All implementors of T are required to have the `BaseOutStructure` layout
                 let next_base = unsafe { &mut *<*mut T>::cast::<BaseOutStructure<'a>>(next) };
                 // `next` here can contain a pointer chain.  This function refuses to insert the struct,
@@ -2374,7 +2372,7 @@ fn derive_getters_and_setters(
             /// The last struct in this chain (i.e. the one where `p_next` is `NULL`) must
             /// be writable memory, as its `p_next` field will be updated with the value of
             /// `self.p_next`.
-            pub unsafe fn extend<T: #extends_name + ?Sized>(mut self, next: &'a mut T) -> Self {
+            pub unsafe fn extend<T: Extends<Self> + ?Sized>(mut self, next: &'a mut T) -> Self {
                 // `next` here can contain a pointer chain. This means that we must correctly
                 // attach he head to the root and the tail to the rest of the chain
                 // For example:
@@ -2392,18 +2390,10 @@ fn derive_getters_and_setters(
 
             #[doc(hidden)]
             #[deprecated = "Migrate to `push()` if `next` does not have an existing chain (i.e. `p_next` is `NULL`), `extend()` otherwise"]
-            pub unsafe fn push_next<T: #extends_name + ?Sized>(self, next: &'a mut T) -> Self {
+            pub unsafe fn push_next<T: Extends<Self> + ?Sized>(self, next: &'a mut T) -> Self {
                 self.extend(next)
             }
         }
-    } else {
-        quote!()
-    };
-
-    // Root structs come with their own trait that structs that extend
-    // this struct will implement
-    let next_trait = if root_struct_next_field.is_some() {
-        quote!(pub unsafe trait #extends_name {})
     } else {
         quote!()
     };
@@ -2415,10 +2405,10 @@ fn derive_getters_and_setters(
         .extends
         .iter()
         .flat_map(|extends| extends.split(','))
-        .map(|extends| format_ident!("Extends{}", name_to_tokens(extends)))
         .map(|extends| {
+            let base = name_to_tokens(extends);
             // Extension structs always have a pNext, and therefore always have a lifetime.
-            quote!(unsafe impl #extends for #name<'_> {})
+            quote!(unsafe impl Extends<#base<'_>> for #name<'_> {})
         });
 
     let impl_structure_type_trait = structure_type_field.map(|member| {
@@ -2441,7 +2431,6 @@ fn derive_getters_and_setters(
     let q = quote! {
         #impl_structure_type_trait
         #(#impl_extend_trait)*
-        #next_trait
 
         impl #lifetime #name #lifetime {
             #(#setters)*
@@ -3451,7 +3440,7 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
         use super::platform_types::*;
         use super::{
             ptr_chain_iter, wrap_c_str_slice_until_nul, write_c_str_slice_with_nul,
-            CStrTooLargeForStaticArray, Handle, Packed24_8, TaggedStructure,
+            CStrTooLargeForStaticArray, Extends, Handle, Packed24_8, TaggedStructure,
         };
         use core::ffi::*;
         use core::fmt;
--- a/Changelog.md
+++ b/Changelog.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The expected length of this array can be queried with the respective `*_len()` variant of these functions.
 - `push_next()` has been renamed to `extend()` and marked as `unsafe`. Users are encouraged to call `push()` for singular structs instead. (#909)
 - Changed and renamed `RawPtr::as_raw_ptr(&self)` trait function to a by-value `RawPtr::to_raw_ptr(self)` function. (#965)
+- All `Extends{Root}` traits have been replaced with a single `Extends<Root>` trait using generics. (#971)
 
 ### Removed
 
--- a/ash/Cargo.toml
+++ b/ash/Cargo.toml
@@ -24,6 +24,9 @@ rust-version = "1.69.0"
 [dependencies]
 libloading = { version = "0.8", optional = true }
 
+[dev-dependencies]
+trybuild = "1.0"
+
 [features]
 default = ["loaded", "debug", "std"]
 # Link the Vulkan loader at compile time.
--- a/ash/src/extensions/khr/acceleration_structure.rs
+++ b/ash/src/extensions/khr/acceleration_structure.rs
@@ -263,7 +263,10 @@ impl crate::khr::acceleration_structure::Device {
         max_primitive_counts: &[u32],
         size_info: &mut vk::AccelerationStructureBuildSizesInfoKHR<'_>,
     ) {
-        assert_eq!(max_primitive_counts.len(), build_info.geometry_count as _);
+        assert_eq!(
+            max_primitive_counts.len(),
+            build_info.geometry_count as usize
+        );
 
         (self.fp.get_acceleration_structure_build_sizes_khr)(
             self.handle,
--- a/ash/src/lib.rs
+++ b/ash/src/lib.rs
@@ -268,3 +268,9 @@ where
         }
     }
 }
+
+#[test]
+fn trybuild() {
+    let t = trybuild::TestCases::new();
+    t.compile_fail("tests/fail/long_lived_root_struct_borrow.rs");
+}
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -50,11 +50,20 @@ pub trait Handle: Sized {
 }
 
 /// Structures implementing this trait are layout-compatible with [`BaseInStructure`] and
-/// [`BaseOutStructure`]. Such structures have an `s_type` field indicating its type, which
-/// must always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
+/// [`BaseOutStructure`]. Such structures have an `s_type` field indicating its type, which must
+/// always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
 pub unsafe trait TaggedStructure {
     const STRUCTURE_TYPE: StructureType;
 }
+
+/// Implemented for every structure that extends base structure `B`. Concretely that means struct
+/// `B` is listed in its array of [`structextends` in the Vulkan registry][1].
+///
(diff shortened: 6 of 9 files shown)


ash/src/vk/prelude.rs
```diff
@@ -63,6 +63,15 @@ pub unsafe trait TaggedStructure {
...
 }
 
+/// Implemented for every structure that extends base structure `B`. Concretely that means struct
+/// `B` is listed in its array of [`structextends` in the Vulkan registry][1].
+///
+/// Similar to [`TaggedStructure`] , all `unsafe` implementers of this trait must guarantee that
+/// their structure is layout-compatible [`vk::BaseInStructure`] and [`vk::BaseOutStructure`].
+///
+/// [1]: https://registry.khronos.org/vulkan/specs/latest/styleguide.html#extensions-interactions
+pub unsafe trait StructExtends<B> {}
```
MarijnS95, line 73: I would've liked to `: TaggedStructure` to imply these docs, but that makes the trait no longer object-safe meaning that folks cannot pass `&mut dyn StructExtends<B>` into any of our `.push*()`/`.extend()` functions.

If we go that route we can also drop `+ ?Sized` from the trait bounds and put `: Sized` on `StructExtends` directly :)

Ralith: We can achieve the best of both worlds by splitting `TaggedStructure` into two traits: one that's object-safe and implies the contract (maybe `AnyTaggedStructure`), and one that inherits that trait (maybe `TaggedStructure: AnyTaggedStructure`), and adds `const STRUCTURE_TYPE`.

Bit of ceremony for a really obscure use case, but it doesn't seem like it'd hurt.

generator/src/lib.rs
```diff
@@ -2373,33 +2371,25 @@ fn derive_getters_and_setters(
...
     let impl_extend_trait = struct_
         .extends
         .iter()
         .flat_map(|extends| extends.split(','))
-        .map(|extends| format_ident!("Extends{}", name_to_tokens(extends)))
         .map(|extends| {
+            let base = name_to_tokens(extends);
             // Extension structs always have a pNext, and therefore always have a lifetime.
-            quote!(unsafe impl #extends for #name<'_> {})
+            quote!(unsafe impl StructExtends<#base<'_>> for #name<'_> {})
```
Ralith, line 2392: Does coupling these two lifetimes have any weird effects? I'm not sure if that's a change in the semantics of `push` impls or not. I guess the examples build fine...

ash/src/vk/prelude.rs
Ralith, line 73: bikeshed: what about just `Extends`?

generator/src/lib.rs
MarijnS95: Good catch. I think there should be _some relation_, such that anything already borrowed by `#name<'a>` gets forwarded into the borrow of `#base<'b>` which I assume is what we end up getting in the `push` functions for `T: Extends<Self>`?

I came up with this little test to validate that:

```rust
let mut layers = vec![];
let mut api =
    vk::PhysicalDeviceLayeredApiPropertiesListKHR::default().layered_apis(&mut layers);
let pdev_props = vk::PhysicalDeviceProperties2::default().push(&mut api);

// Access to either variable is disallowed because both are mutably borrowed in pdev_props
dbg!(&api);
dbg!(&layers);

dbg!(pdev_props); // Holds a borrow on api and layers
```

Specifically `layers` should be inaccessible as long as `pdev_props` is live. This appears to be satisfied (by failed compilation) for each of these:

```rust
unsafe impl Extends<PhysicalDeviceProperties2<'_>> for PhysicalDeviceLayeredApiPropertiesListKHR<'_> {}
unsafe impl<'root, 'child> Extends<PhysicalDeviceProperties2<'root>> for PhysicalDeviceLayeredApiPropertiesListKHR<'child> {}
unsafe impl<'a> Extends<PhysicalDeviceProperties2<'a>> for PhysicalDeviceLayeredApiPropertiesListKHR<'a> {}
```

And I'm not exactly sure what is most correct. Perhaps the second case might lack a `'child: 'root`, i.e. the scope of `'child` is at least as large, or bigger than `'root` (`'child` can live longer than `'root` but not vice-versa. Likewise, as long as `'root` is `mut`ably borrowed, `'child` is `mut`ably borrowed as well).

:shrug: as usual, I could use a little bit of help/validation on lifetimes :)

ash/src/vk/prelude.rs
MarijnS95: Let's do that when we get there? I'm not exactly set on keeping the ability to pass `&dyn` around, but folks like #953 seem to be wanting that and even added a `ASH_DYNAMIC` to get around this, which I think we should avert by splitting the trait like you described, and not having a `const` available for "generic" unknown objects.

Having to implement two (hierarchical) traits just to share the layout requirement documentation sounds like overdoing it.

MarijnS95: When I suggested to add our `push()` functions to this not-object-safe `TaggedStructure` in https://github.com/ash-rs/ash/issues/879, I think we'd actually break that `&dyn` case already without the `AnyTaggedStructure` (or `RuntimeTaggedStructure`)? I think we'll add it in the next PR then.

ash/src/vk/prelude.rs
```diff
@@ -63,6 +63,15 @@ pub unsafe trait TaggedStructure {
...
 }
 
+/// Implemented for every structure that extends base structure `B`. Concretely that means struct
+/// `B` is listed in its array of [`structextends` in the Vulkan registry][1].
+///
+/// Similar to [`TaggedStructure`] , all `unsafe` implementers of this trait must guarantee that
+/// their structure is layout-compatible [`vk::BaseInStructure`] and [`vk::BaseOutStructure`].
+///
+/// [1]: https://registry.khronos.org/vulkan/specs/latest/styleguide.html#extensions-interactions
+pub unsafe trait Extends<B> {}
```
MarijnS95, line 73: Fwiw I've not found a use/need for it yet, but we might want to consider the base struct "unsized" / object-safe too?

```suggestion
pub unsafe trait Extends<B: ?Sized> {}
```

boshra36 (approved):

MarijnS95:
It looks like this PR comes with a ±150ms build speedup:

## `master` (92a0ee6)

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.964 s ±  0.109 s    [User: 7.768 s, System: 0.902 s]
  Range (min … max):    7.797 s …  8.133 s    10 runs
```

## This PR (1cf5047d)

```console
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.817 s ±  0.092 s    [User: 7.669 s, System: 0.845 s]
  Range (min … max):    7.707 s …  8.008 s    10 runs
```

ash/src/vk/prelude.rs
Ralith: I'd be inclined not to for simplicity, since it's difficult to imagine `Extends<dyn Trait>` or `Extends<[T]>` being useful. Note that `dyn Extends<T>` is already valid.

Ralith:
Awesome! Glad we finally stumbled on *something* that helps there, and great that it's a readability/concision win too.

MarijnS95:
Sneak-preview: moving the functions to `AnyTaggedStructure` is going to give us another ±90ms :)

ash/src/vk/prelude.rs
Ralith: > Having to implement two (hierarchical) traits

We could flip that around with `impl<T: TaggedStructure> AnyTaggedStructure for T`; then it's completely automatic.

ash/src/vk/prelude.rs
MarijnS95: Yeah, `dyn Extends<T>` is valid, as checked by the new tests (and requested by users), but `Extends<dyn T>` is not though we don't have any trait bound `T` that would be useful in this place.

ash/src/vk/prelude.rs
MarijnS95: That's great, I didn't expect that to work because of a circular `TaggedStructure: AnyTaggedStructure` dependency, but it does :tada:

generator/src/lib.rs
Ralith: The `'_` version is equivalent to the single-variable `'a` version. I think the two-variable version is directly analogous to what we had before; the missing constraint is funny looking, but I believe exactly that is implied whenever you actually call `push` because `&'root Child<'child>` is illegal unless `'child: 'root`, so everything works out.

I think it does make sense that the single-variable version works too: we can always coerce `Child<'child>` to `Extension<'root>` if `'child: 'root`, so `fn push(&mut self, child: &'root Child<'root>)` is well-typed and has no adverse effects.

So, both approaches are at least superficially valid. I lean towards the two-variable approach:
- This trait describes a structural relationship between types, not a dynamic one, so saying things about lifetimes doesn't seem relevant
- There's no benefit in introducing a constraint that the signature of `push` already enforces
- It minimizes the change wrt. the previous semantics

Not a strongly held opinion, though. It's not exactly unintuitive to say that an extension definitionally outlives the thing it extends.

Ralith: We should have a `trybuild` test to validate that short-lived extension structs with long-lived roots are rejected.

MarijnS95: I added `trybuild`. But whenever having this snippet in `ash/src/lib.rs`:

```rust
#[test]
fn trybuild() {
    let t = trybuild::TestCases::new();
    // t.compile_fail("tests/fail/long_lived_root_struct_borrow.rs");
}
```

```console
❯ cargo t
   Compiling ash-window v0.13.0 (/newdata/Code/TraverseResearch/ash/ash-window)
   Compiling ash v0.38.0+1.3.296 (/newdata/Code/TraverseResearch/ash/ash)
   Compiling ash-examples v0.1.0 (/newdata/Code/TraverseResearch/ash/ash-examples)
error[E0282]: type annotations needed
   --> ash/src/extensions/khr/acceleration_structure.rs:266:77
    |
266 |         assert_eq!(max_primitive_counts.len(), build_info.geometry_count as _);
    |                                                                             ^ cannot infer type

For more information about this error, try `rustc --explain E0282`.
error: could not compile `ash` (lib test) due to 1 previous error
```

@Ralith any clue?

MarijnS95: Also, I love this crate and suggestion already. As mentioned I immediately moved the `push()` functions to `trait AnyTaggedStructure`, and had a `// TODO: update lifetimes` because the `fn push<T: Extends<Self> + ?Sized>(mut self, next: &mut T) -> Self` no longer captures the borrow of `T` in the returned `Self`, and the test caught it :+1:

Ralith: > Obviously the fix is easy - I just wish to understand why this wasn't a problem in the past. Does this crate affect assert_eq!()?

That is pretty weird. Maybe it's getting a very old rustc from your environment somewhere, or the build configuration is otherwise changing to add/remove some `PartialEq` impls in ways that introduce ambiguity? Not sure.

Ralith: It's not your environment, repros here too... must be a change in the visible impls somehow.

MarijnS95: And it occurs while compiling `ash (lib test)` before `trybuild` is executed. Unless `trybuild` uses some magic to inject itself whenever it's compiled/linked into our crate, but that should happen with it being present in `Cargo.toml` regardless.

I don't see any difference in `cargo t -v -j1` leading up to the error that stands out.

MarijnS95:
Performance optimization is now less significant, ±105ms, perhaps because of the expanded lifetimes?

(Retested `master` too and its performance measurement was within margin of error on my machine)

## This PR (4305ec7d):

```console
❯ hyperfine -p 'cargo clean -p ash' 'cargo b -p ash'
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.858 s ±  0.103 s    [User: 7.702 s, System: 0.876 s]
  Range (min … max):    7.745 s …  8.026 s    10 runs
```

Ralith (approved):
We're done here, right? I'm happy to discuss changing the lifetime semantics to try to recapture some of that buildtime if you want, but it's a net win regardless.

MarijnS95:
### `master` @ 98e5db8

```console
❯ hyperfine -p 'cargo clean -p ash' 'cargo b -p ash'
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.118 s ±  0.072 s    [User: 7.144 s, System: 0.827 s]
  Range (min … max):    7.041 s …  7.233 s    10 runs
```

### This PR @ cbdf201

```console
❯ hyperfine -p 'cargo clean -p ash' 'cargo b -p ash'
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.149 s ±  0.093 s    [User: 7.168 s, System: 0.854 s]
  Range (min … max):    7.054 s …  7.352 s    10 runs
```

### This PR @ cbdf201 with unnamed (`'_`) lifetime

```rust
quote!(unsafe impl Extends<#base<'_>> for #name<'_> {})
```
Yields:
```console
❯ hyperfine -p 'cargo clean -p ash' 'cargo b -p ash'
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.141 s ±  0.086 s    [User: 7.163 s, System: 0.832 s]
  Range (min … max):    7.011 s …  7.271 s    10 runs
```

### `'a`

Just in case I named the parameter with:

```rust
quote!(unsafe impl<'a> Extends<#base<'a>> for #name<'a> {})
```
Within margin of error:
```console
❯ hyperfine -p 'cargo clean -p ash' 'cargo b -p ash'
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.155 s ±  0.086 s    [User: 7.187 s, System: 0.834 s]
  Range (min … max):    7.046 s …  7.352 s    10 runs
```

### Explicitly specifying the `'child: 'root` bound:

```rust
quote!(unsafe impl<'root, 'child: 'root> Extends<#base<'root>> for #name<'child> {})
```

```console
❯ hyperfine -p 'cargo clean -p ash' 'cargo b -p ash'
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.143 s ±  0.090 s    [User: 7.157 s, System: 0.847 s]
  Range (min … max):    6.977 s …  7.261 s    10 runs
```

MarijnS95:
Just in case I prefer the nameless `'_` variant, because it keeps a lot more `impl`s on the same line. We have tests ensuring that the lifetime bounds work as we expect (allowing certain cases, while failing others).

```console
❯ git diff --stat
 ash/src/vk/definitions.rs | 3991 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-------------------------------------------------------------------------------------------------------------------------------------------------
 generator/src/lib.rs      |    2 +-
 2 files changed, 1215 insertions(+), 2778 deletions(-)
```

Ralith:
So in summary, no obvious impact from anything? Weird, I wonder what we measured before. Regardless, I'm happy to proceed with the terse anonymous lifetime variant.

PR #972, opened 2024-12-07
Inline `predule` module contents into their parent modules

MarijnS95:
We've never used `prelude` in a sensible way, both modules inside the crate root and `vk` namespace have always abused these modules to store a random combination of private utilities next to a few extra hand-written type/trait definitions and implementations, as outlined before in https://github.com/ash-rs/ash/issues/879#issuecomment-2524289739.

Neither of these need to glob-imported by users to make easy use of `ash` functionality besides perhaps ensuring traits like `Handle` and `TaggedStructure` are in scope.  Their glob imports were only useful to `ash` itself, and even then the `vk::prelude` was already fully glob-reexported in `vk`.

Besides inlining its contents, I've shuffled some private helpers like `debug_flags()` closer to where they are being used.


ash/src/vk/definitions.rs
```diff
@@ -1,12 +1,10 @@
...
-use crate::vk::native::*;
-use crate::vk::platform_types::*;
-use crate::vk::prelude::*;
-use crate::vk::{ptr_chain_iter, Handle};
+use super::aliases::*;
+use super::constants::*;
+use super::enums::*;
+use super::native::*;
+use super::platform_types::*;
+use super::*;
```
MarijnS95, line 7: XXX: Might want to import these explicitly now.

ash/src/vk.rs
```diff
@@ -46,21 +72,153 @@ pub(crate) unsafe fn ptr_chain_iter<T: ?Sized>(
...
-    ///
-    /// ```
-    /// # use ash::vk::{Handle, Instance};
-    /// let instance = Instance::null();
-    /// assert!(instance.is_null());
-    /// ```
-    fn is_null(self) -> bool {
-        self.as_raw() == 0
+#[cfg(test)]
+mod tests {
```
Ralith, line 77: nit: place at the bottom of the module?

generator/src/lib.rs
```diff
@@ -3406,26 +3406,28 @@ pub fn write_source_code<P: AsRef<Path>>(vk_headers_dir: &Path, src_dir: P) {
 
     let feature_code = quote! {
         use core::ffi::*;
-        use crate::vk::bitflags::*;
-        use crate::vk::definitions::*;
-        use crate::vk::enums::*;
+        use super::bitflags::*;
+        use super::definitions::*;
+        use super::enums::*;
```
Ralith, line 3411: What's the motivation for this change? Outside of inline modules, I usually prefer to avoid `super` since it makes it more difficult to move code around.

ash/src/vk.rs
MarijnS95: Done. Wanted to keep it close to `fn ptr_chain_iter()` which is the only thing that is being tested; for that we could've dropped the module entirely though because there's nothing else to test.

generator/src/lib.rs
MarijnS95: I don't particularly plan to move this code around very often, and favor locality/shortest-path. When reading `crate::vk::` it confuses me, because I _thought_ I was already somewhere inside the `vk` yet going into the module hierarchy from the root causes a disconnect.

It also feels like it helps identify the remaining cases where things, i.e. specifically the types and helpers from `prelude.rs` are defined in the wrong spot, if they're referred across our top-level modules. A good example is `fn debug_flags()` that seems more fitting inside the `vk` module.

ash/src/vk.rs
MarijnS95: Actually we also have a very minimal `ash/tests/display.rs` that we could inline into `vk.rs` where the `fn debug_flags()` at test is defined?

Ralith: Sure, if you like.

generator/src/lib.rs
Ralith: Works for me.

Ralith (approved):
Thanks for tidying this up! Been irritating me for years.

ash/src/vk.rs
MarijnS95: Decided to keep the `mod` as it seems more fitting with multiple `#[test]` functions (unless you think those individual test functions should move below the actual implementations), and because I was extending the `push()` tests with more missed validations at #971.

MarijnS95:
> Thanks for tidying this up! Been irritating me for years.

Any time :tada:

ash/src/vk/definitions.rs
MarijnS95: Didn't do this because it was way too messy.

PR #983, opened 2025-02-07
fix(ohos): OpenHarmony should link libvulkan.so

richerfu:
For OpenHarmony/HarmonyNext, we need to link `libvulkan.so` not `libvulkan.so.1`. See detail with [doc](https://developer.huawei.com/consumer/en/doc/harmonyos-references-V5/vulkan-guidelines-V5#how-to-develop)

### DIFF
--- a/ash/src/entry.rs
+++ b/ash/src/entry.rs
@@ -69,12 +69,13 @@ impl Entry {
                 target_os = "macos",
                 target_os = "ios",
                 target_os = "android",
-                target_os = "fuchsia"
+                target_os = "fuchsia",
+                target_env = "ohos"
             ))
         ))]
         const LIB_PATH: &str = "libvulkan.so.1";
 
-        #[cfg(any(target_os = "android", target_os = "fuchsia"))]
+        #[cfg(any(target_os = "android", target_os = "fuchsia", target_env = "ohos"))]
         const LIB_PATH: &str = "libvulkan.so";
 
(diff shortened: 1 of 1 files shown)


Ralith:
> See detail with [doc](https://developer.huawei.com/consumer/en/doc/harmonyos-references-V5/vulkan-guidelines-V5#how-to-develop)

This documentation isn't definitive. Usually `libvulkan.so` is just a symlink to `libvulkan.so.1`, and those symlinks are intended for use at compile time, resolved to the versioned path before being recorded in the binary's `RUNPATH`. Can you confirm whether libvulkan.so.1 exists on this platform?

richerfu:
> > See detail with [doc](https://developer.huawei.com/consumer/en/doc/harmonyos-references-V5/vulkan-guidelines-V5#how-to-develop)
> 
> This documentation isn't definitive. Usually `libvulkan.so` is just a symlink to `libvulkan.so.1`, and those symlinks are intended for use at compile time, resolved to the versioned path before being recorded in the binary's `RUNPATH`. Can you confirm whether libvulkan.so.1 exists on this platform?

I can promise that `libvulkan.so.1` doesn't exist in OpenHarmony.

- I try to use wgpu which use ash as vulkan render and get some error(dlopen failed).
- NDK only provide `libvulkan.so` to link.

image

Ralith (approved):
Thanks!

PR #987, opened 2025-03-11
generator: Bump nom to 8.0.0

linkmauve:
The main change is that we now have to call `.parse(i)` instead of the previous closure-based `(i)` call.

Supersedes #981.


Ralith (approved):

Ralith:
Thanks for drafting this! Looks like it's mostly good, but needs a `cargo fmt`.

linkmauve:
Here, force-pushed with `cargo fmt`. Sorry about that!

Ralith:
Thank you!

PR #989, opened 2025-03-24
generator: Replace `super::super::*` with narrower imports.

TechPizzaDev:
A `mod` importing itself hits a slow path in rustc at `compute_effective_visibilities` since cyclic imports are handled specially. This improvement also happens to greatly reduce the size of `.rmeta` files.

Self-profile of release build for `ash` on `AMD Ryzen 7 7700X`:
| Version   | `rmeta` Size | CPU Time | Real Time |
| --------- | ------------ | -------- | --------- |
| Before PR | ~68 MiB      | 15.3s    | 23.3s     |
| After PR  | ~40 MiB      | 12.6s    | 23.3s     |
| Diff      | -29 MiB      | -2.6s    | 0s        |

Elapsed real time didn't change likely because of build pipelining. I imagine weaker machines could benefit more from this. This should in general help many projects since `ash` is used in `wgpu`.

### DIFF
--- a/ash/src/extensions_generated.rs
+++ b/ash/src/extensions_generated.rs
@@ -1,11 +1,7 @@
-#![allow(unused_imports)]
-use crate::vk::*;
-use core::ffi::*;
 #[doc = "Extensions tagged AMD"]
 pub mod amd {
     #[doc = "VK_AMD_rasterization_order"]
     pub mod rasterization_order {
-        use super::super::*;
         pub use {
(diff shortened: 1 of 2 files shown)


Ralith (approved):
Nice, thanks! Even if wall time doesn't change, reduced CPU time should mean more CPU for other parts of downstream builds, I think. I wonder if using fully qualified names would help even more.

Ralith (approved):

PR #994, opened 2025-05-08
Deduplicate `push()`/`extend()` helpers as a default impl on `TaggedStructure`

MarijnS95:
Having these functions generated for every root struct in our definitions file contributes to significant bloat, in part because of their massive documentation comments and not containing any generator-dynamic code or identifiers.

Now that `trait Extends{Root}` was generalized into a single `trait Extends<Root>` with the root structure as generic argument, it becomes possible to build on top of that and define a default trait function on every Vulkan structure.  These functions require the "pushed" type to derive `Extends<Self>`, hence limiting the function to be called without ever having to track if the root structure is being extended.

### DIFF
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -49,25 +49,9 @@ pub trait Handle: Sized {
     }
 }
 
-/// Structures implementing this trait are layout-compatible with [`BaseInStructure`] and
-/// [`BaseOutStructure`]. Such structures have an `s_type` field indicating its type, which must
-/// always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
-pub unsafe trait TaggedStructure {
-    const STRUCTURE_TYPE: StructureType;
-}
-
-/// Implemented for every structure that extends base structure `B`. Concretely that means struct
-/// `B` is listed in its array of [`structextends` in the Vulkan registry][1].
-///
-/// Similar to [`TaggedStructure`], all `unsafe` implementers of this trait must guarantee that
-/// their structure is layout-compatible [`BaseInStructure`] and [`BaseOutStructure`].
-///
-/// [1]: https://registry.khronos.org/vulkan/specs/latest/styleguide.html#extensions-interactions
-pub unsafe trait Extends<B> {}
-
 /// Iterates through the pointer chain. Includes the item that is passed into the function.
 /// Stops at the last [`BaseOutStructure`] that has a null [`BaseOutStructure::p_next`] field.
-pub(crate) unsafe fn ptr_chain_iter<T: ?Sized>(
+pub(crate) unsafe fn ptr_chain_iter<'a, T: AnyTaggedStructure<'a> + ?Sized>(
     ptr: &mut T,
 ) -> impl Iterator<Item = *mut BaseOutStructure<'_>> {
     let ptr = <*mut T>::cast::<BaseOutStructure<'_>>(ptr);
@@ -81,6 +65,103 @@ pub(crate) unsafe fn ptr_chain_iter<T: ?Sized>(
         Some(old)
     })
 }
+
+/// Structures implementing this trait are layout-compatible with [`BaseInStructure`] and
+/// [`BaseOutStructure`]. Such structures have an `s_type` field indicating its type, which must
+/// always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
+pub unsafe trait AnyTaggedStructure<'a> {
+    /// Prepends the given extension struct between the root and the first pointer. This
+    /// method only exists on structs that can be passed to a function directly. Only
+    /// valid extension structs can be pushed into the chain.
+    /// If the chain looks like `A -> B -> C`, and you call `A.push(&mut D)`, then the
+    /// chain will look like `A -> D -> B -> C`.
+    ///
+    /// # Panics
+    /// If `next` contains a pointer chain of its own, this function will panic.  Call
+    /// `unsafe` [`Self::extend()`] to insert this chain instead.
+    fn push<T: Extends<'a, Self> + ?Sized>(mut self, next: &'a mut T) -> Self
+    where
+        Self: Sized,
+    {
+        // SAFETY: All implementors of `AnyTaggedStructure` are required to have the `BaseOutStructure` layout
+        let slf_base = unsafe { &mut *<*mut _>::cast::<BaseOutStructure<'_>>(&mut self) };
+        // SAFETY: All implementors of `T: Extends<'a, >: AnyTaggedStructure` are required to have the `BaseOutStructure` layout
+        let next_base = unsafe { &mut *<*mut T>::cast::<BaseOutStructure<'_>>(next) };
+        // `next` here can contain a pointer chain.  This function refuses to insert the struct,
+        // in favour of calling unsafe extend().
+        assert!(
+            next_base.p_next.is_null(),
+            "push() expects a struct without an existing p_next pointer chain (equal to NULL)"
+        );
+        next_base.p_next = slf_base.p_next;
+        slf_base.p_next = next_base;
+        self
+    }
+
+    /// Prepends the given extension struct between the root and the first pointer. This
+    /// method only exists on structs that can be passed to a function directly. Only
+    /// valid extension structs can be pushed into the chain.
+    /// If the chain looks like `A -> B -> C` and `D -> E`, and you call `A.extend(&mut D)`,
+    /// then the chain will look like `A -> D -> E -> B -> C`.
+    ///
+    /// # Safety
+    /// This function will walk the [`BaseOutStructure::p_next`] chain of `next`, requiring
+    /// all non-`NULL` pointers to point to a valid Vulkan structure starting with the
+    /// [`BaseOutStructure`] layout.
+    ///
+    /// The last struct in this chain (i.e. the one where `p_next` is `NULL`) must
+    /// be writable memory, as its `p_next` field will be updated with the value of
+    /// `self.p_next`.
+    unsafe fn extend<T: Extends<'a, Self> + ?Sized>(mut self, next: &'a mut T) -> Self
+    where
+        Self: Sized,
+    {
+        // `next` here can contain a pointer chain. This means that we must correctly
+        // attach he head to the root and the tail to the rest of the chain
+        // For example:
+        //
+        // next = A -> B
+        // Before: `Root -> C -> D -> E`
+        // After: `Root -> A -> B -> C -> D -> E`
+        //                 ^^^^^^
+        //                 next chain
+        let slf_base = unsafe { &mut *<*mut _>::cast::<BaseOutStructure<'_>>(&mut self) };
+        let next_base = <*mut T>::cast::<BaseOutStructure<'_>>(next);
+        let last_next = ptr_chain_iter(next).last().unwrap();
+        (*last_next).p_next = slf_base.p_next;
+        slf_base.p_next = next_base;
+        self
+    }
+
+    #[doc(hidden)]
+    #[deprecated = "Migrate to `push()` if `next` does not have an existing chain (i.e. `p_next` is `NULL`), `extend()` otherwise"]
+    unsafe fn push_next<T: Extends<'a, Self> + ?Sized>(self, next: &'a mut T) -> Self
+    where
+        Self: Sized,
+    {
+        self.extend(next)
+    }
+}
+
+/// Non-object-safe variant of [`AnyTaggedStructure`], meaning that a `dyn TaggedStructure` cannot
+/// exist but as a consequence the [`TaggedStructure::STRUCTURE_TYPE`] associated constant is
+/// available.
+///
+/// [`AnyTaggedStructure`]s have a [`BaseInStructure::s_type`] field indicating its type, which must
+/// always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
+pub unsafe trait TaggedStructure<'a>: AnyTaggedStructure<'a> {
+    const STRUCTURE_TYPE: StructureType;
+}
+unsafe impl<'a, T: TaggedStructure<'a>> AnyTaggedStructure<'a> for T {}
+
+/// Implemented for every structure that extends base structure `B`. Concretely that means struct
+/// `B` is listed in its array of [`structextends` in the Vulkan registry][1].
+///
+/// Similar to [`TaggedStructure`], all `unsafe` implementers of this trait must guarantee that
+/// their structure is layout-compatible [`BaseInStructure`] and [`BaseOutStructure`].
+///
+/// [1]: https://registry.khronos.org/vulkan/specs/latest/styleguide.html#extensions-interactions
+pub unsafe trait Extends<'a, B: AnyTaggedStructure<'a>>: AnyTaggedStructure<'a> {}
 
 /// Holds 24 bits in the least significant bits of memory,
 /// and 8 bytes in the most significant bits of that memory,
@@ -211,6 +292,7 @@ pub(crate) fn debug_flags<Value: Into<u64> + Copy>(
 #[cfg(test)]
 mod tests {
     use crate::vk;
+    use crate::vk::AnyTaggedStructure as _;
     use alloc::vec::Vec;
     #[test]
     fn test_ptr_chains() {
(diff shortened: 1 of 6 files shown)


generator/src/lib.rs
```diff
@@ -2408,7 +2339,7 @@ fn derive_getters_and_setters(
         .map(|extends| {
             let base = name_to_tokens(extends);
             // Extension structs always have a pNext, and therefore always have a lifetime.
-            quote!(unsafe impl Extends<#base<'_>> for #name<'_> {})
+            quote!(unsafe impl<'a> Extends<'a, #base<'a>> for #name<'a> {})
```
MarijnS95, line 2342: Named lifetimes are back otherwise the `trybuild` test no longer fails as expected.

ash/src/vk.rs
```diff
@@ -82,6 +66,103 @@ pub(crate) unsafe fn ptr_chain_iter<T: ?Sized>(
...
+    /// Prepends the given extension struct between the root and the first pointer. This
+    /// method only exists on structs that can be passed to a function directly. Only
+    /// valid extension structs can be pushed into the chain.
+    /// If the chain looks like `A -> B -> C`, and you call `A.push(&mut D)`, then the
+    /// chain will look like `A -> D -> B -> C`.
+    ///
+    /// # Panics
+    /// If `next` contains a pointer chain of its own, this function will panic.  Call
+    /// `unsafe` [`Self::extend()`] to insert this chain instead.
+    fn push<T: Extends<'a, Self> + ?Sized>(mut self, next: &'a mut T) -> Self
```
Ralith, line 82: If the goal if this trait is object safety, do we need to revise this not to take `self` by value? You can't pass unsized types by value, after all.

ash/src/vk.rs
```diff
@@ -82,6 +66,103 @@ pub(crate) unsafe fn ptr_chain_iter<T: ?Sized>(
...
+        let next_base = <*mut T>::cast::<BaseOutStructure<'_>>(next);
+        let last_next = ptr_chain_iter(next).last().unwrap();
+        (*last_next).p_next = slf_base.p_next;
+        slf_base.p_next = next_base;
+        self
+    }
+
+    #[doc(hidden)]
+    #[deprecated = "Migrate to `push()` if `next` does not have an existing chain (i.e. `p_next` is `NULL`), `extend()` otherwise"]
+    unsafe fn push_next<T: Extends<'a, Self> + ?Sized>(self, next: &'a mut T) -> Self
```
Ralith, line 138: Should we take the opportunity to kill this?

ash/src/vk.rs
```diff
@@ -82,6 +66,103 @@ pub(crate) unsafe fn ptr_chain_iter<T: ?Sized>(
     })
 }
 
+/// Structures implementing this trait are layout-compatible with [`BaseInStructure`] and
+/// [`BaseOutStructure`]. Such structures have an `s_type` field indicating its type, which must
+/// always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
+pub unsafe trait AnyTaggedStructure<'a> {
+    /// Prepends the given extension struct between the root and the first pointer. This
+    /// method only exists on structs that can be passed to a function directly. Only
```
Ralith, line 74: > This method only exists on structs that can be passed to a function directly

Is that still true?

ash/src/vk.rs
```diff
@@ -82,6 +66,103 @@ pub(crate) unsafe fn ptr_chain_iter<T: ?Sized>(
...
+unsafe impl<'a, T: TaggedStructure<'a>> AnyTaggedStructure<'a> for T {}
+
+/// Implemented for every structure that extends base structure `B`. Concretely that means struct
+/// `B` is listed in its array of [`structextends` in the Vulkan registry][1].
+///
+/// Similar to [`TaggedStructure`], all `unsafe` implementers of this trait must guarantee that
+/// their structure is layout-compatible [`BaseInStructure`] and [`BaseOutStructure`].
+///
+/// [1]: https://registry.khronos.org/vulkan/specs/latest/styleguide.html#extensions-interactions
+pub unsafe trait Extends<'a, B: AnyTaggedStructure<'a>>: AnyTaggedStructure<'a> {}
```
Ralith, line 164: Note to self: play with these lifetimes to see if we can simplify any further before we commit.

Ralith (changes requested):

ash/src/vk.rs
MarijnS95: No, I think the first goal was to provide object safety for the `next` parameter `T: Extends<Self>` first. We can take a look at reworking the builder pattern to support that later, perhaps?

ash/src/vk.rs
MarijnS95: Since we're planning on a breaking release, I'm more than happy to not keep legacy deprecated code around; folks should read the changelog and learn how/why to migrate their original `.push_next()` to either call `push()` or `extend()` instead.

ash/src/vk.rs
MarijnS95: If I'm not mistaken, there exists no `Extends<T>` (based on `vk.xml` `structextends="T"`) for a `T` that is not a "root structure" (~= structure that can be passed to a function). Hence there's no way to satisfy the `T: Extends<Self>` bound (enabling this `fn push()` function) unless `Self` was such a root structure?

Ralith: Ah, right. I think "exists" is slightly confusing word choice for this, since the method does literally exist on "any tagged structure", but for non-root structures there are not (normally) any legal *arguments* to it. Maybe:
```suggestion
    /// method is only useful on structs that can be passed to a function directly. Only
```

MarijnS95: Perhaps the wording `is only available` solves that? Perhaps `available` carries the right weight of "it exists, but you may not always be able to use it".

Ralith:
Alright, spent some time messing around with this.

First: It seems like a lifetime parameter is needed on whatever trait `push` abstracts over, because that's the only way we can cause inference to constrain the lifetime of the base struct being extended. Clauses like `Self: 'a` set a *minimum* bound on the lifetime of `Self`, which is the opposite of what we want.

Separately, I'm not sure the object safety achieved here is useful. IIRC (though I can't find the discussion) the motivation was to support stuff like `Vec<Box<dyn Extends<T>>>`. However, with this change, we need to specify lifetimes as part of the trait object. A `Vec<Box<dyn Extends<'a, T>>>` is much less useful, since it's not `'static`; for example, it can't be safely passed through a `mpsc`. Outright passing `'static` for the lifetime means the trait objects can't be pushed as an extension, since that would require narrowing their lifetime. If the caller unsafely transmutes `'static` to `'a`, then they're not getting useful lifetime checking, which is most of the benefit of `push`.

If we remove the object safety constraint, then it's no longer necessary to have `Extends` be a subtrait of `AnyTaggedStructure`, so `Extends` no longer needs a lifetime at all. We can get the lifetime we need in `push` with independent bounds like `T: Extends<Self> + TaggedStructure<'a>` which gets us back to the nice simple interface prior to this PR, provided that we `impl TaggedStructure<'a> for Foo<'a>`. The `impl TaggedStructure<'_> for Foo<'_>` currently in this PR is actually `impl<'a, 'b> TaggedStructure<'a> for Foo<'b>` which is too weak.

In short, I think the following diff, on top of this PR, plus a generator rerun, is a sweet spot:

<details>

```commit c3e2205ab2eaa20e7e4373a64325e6bf9ec6b1e4
diff --git a/ash/src/vk.rs b/ash/src/vk.rs
index 71ae233..02210c3 100644
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -51,7 +51,7 @@ pub trait Handle: Sized {
 
 /// Iterates through the pointer chain. Includes the item that is passed into the function. Stops at
 /// the last [`BaseOutStructure`] that has a null [`BaseOutStructure::p_next`] field.
-pub(crate) unsafe fn ptr_chain_iter<'a, T: AnyTaggedStructure<'a> + ?Sized>(
+pub(crate) unsafe fn ptr_chain_iter<'a, T: TaggedStructure<'a> + ?Sized>(
     ptr: &mut T,
 ) -> impl Iterator<Item = *mut BaseOutStructure<'_>> {
     let ptr = <*mut T>::cast::<BaseOutStructure<'_>>(ptr);
@@ -69,7 +69,9 @@ pub(crate) unsafe fn ptr_chain_iter<'a, T: AnyTaggedStructure<'a> + ?Sized>(
 /// Structures implementing this trait are layout-compatible with [`BaseInStructure`] and
 /// [`BaseOutStructure`]. Such structures have an `s_type` field indicating its type, which must
 /// always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
-pub unsafe trait AnyTaggedStructure<'a> {
+pub unsafe trait TaggedStructure<'a> {
+    const STRUCTURE_TYPE: StructureType;
+
     /// Prepends the given extension struct between the root and the first pointer. This method is
     /// only available on structs that can be passed to a function directly. Only valid extension
     /// structs can be pushed into the chain.
@@ -79,13 +81,13 @@ pub unsafe trait AnyTaggedStructure<'a> {
     /// # Panics
     /// If `next` contains a pointer chain of its own, this function will panic.  Call `unsafe`
     /// [`Self::extend()`] to insert this chain instead.
-    fn push<T: Extends<'a, Self> + ?Sized>(mut self, next: &'a mut T) -> Self
+    fn push<T: Extends<Self> + TaggedStructure<'a> + ?Sized>(mut self, next: &'a mut T) -> Self
     where
         Self: Sized,
     {
-        // SAFETY: All implementors of `AnyTaggedStructure` are required to have the `BaseOutStructure` layout
+        // SAFETY: All implementors of `TaggedStructure` are required to have the `BaseOutStructure` layout
         let slf_base = unsafe { &mut *<*mut _>::cast::<BaseOutStructure<'_>>(&mut self) };
-        // SAFETY: All implementors of `T: Extends<'a, >: AnyTaggedStructure` are required to have the `BaseOutStructure` layout
+        // SAFETY: All implementors of `TaggedStructure` are required to have the `BaseOutStructure` layout
         let next_base = unsafe { &mut *<*mut T>::cast::<BaseOutStructure<'_>>(next) };
         // `next` here can contain a pointer chain.  This function refuses to insert the struct,
         // in favour of calling unsafe extend().
@@ -111,7 +113,10 @@ pub unsafe trait AnyTaggedStructure<'a> {
     ///
     /// The last struct in this chain (i.e. the one where `p_next` is `NULL`) must be writable
     /// memory, as its `p_next` field will be updated with the value of `self.p_next`.
-    unsafe fn extend<T: Extends<'a, Self> + ?Sized>(mut self, next: &'a mut T) -> Self
+    unsafe fn extend<T: Extends<Self> + TaggedStructure<'a> + ?Sized>(
+        mut self,
+        next: &'a mut T,
+    ) -> Self
     where
         Self: Sized,
     {
@@ -132,17 +137,6 @@ pub unsafe trait AnyTaggedStructure<'a> {
     }
 }
 
-/// Non-object-safe variant of [`AnyTaggedStructure`], meaning that a `dyn TaggedStructure` cannot
-/// exist but as a consequence the [`TaggedStructure::STRUCTURE_TYPE`] associated constant is
-/// available.
-///
-/// [`AnyTaggedStructure`]s have a [`BaseInStructure::s_type`] field indicating its type, which must
-/// always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
-pub unsafe trait TaggedStructure<'a>: AnyTaggedStructure<'a> {
-    const STRUCTURE_TYPE: StructureType;
-}
-unsafe impl<'a, T: TaggedStructure<'a>> AnyTaggedStructure<'a> for T {}
-
 /// Implemented for every structure that extends base structure `B`. Concretely that means struct
 /// `B` is listed in its array of [`structextends` in the Vulkan registry][1].
 ///
@@ -150,7 +144,7 @@ unsafe impl<'a, T: TaggedStructure<'a>> AnyTaggedStructure<'a> for T {}
 /// their structure is layout-compatible [`BaseInStructure`] and [`BaseOutStructure`].
 ///
 /// [1]: https://registry.khronos.org/vulkan/specs/latest/styleguide.html#extensions-interactions
-pub unsafe trait Extends<'a, B: AnyTaggedStructure<'a>>: AnyTaggedStructure<'a> {}
+pub unsafe trait Extends<B> {}
 
 /// Holds 24 bits in the least significant bits of memory,
 /// and 8 bytes in the most significant bits of that memory,
@@ -281,7 +275,7 @@ pub(crate) fn debug_flags<Value: Into<u64> + Copy>(
 #[cfg(test)]
 mod tests {
     use crate::vk;
-    use crate::vk::AnyTaggedStructure as _;
+    use crate::vk::TaggedStructure as _;
     use alloc::vec::Vec;
     #[test]
     fn test_ptr_chains() {
@@ -343,21 +337,6 @@ mod tests {
         assert_eq!(chain, chain2);
     }
 
-    #[test]
-    fn test_dynamic_add_to_ptr_chain() {
-        let mut variable_pointers = vk::PhysicalDeviceVariablePointerFeatures::default();
-        let variable_pointers: &mut dyn vk::Extends<'_, vk::DeviceCreateInfo<'_>> =
-            &mut variable_pointers;
-        let chain = alloc::vec![<*mut _>::cast(variable_pointers)];
-        let mut device_create_info = vk::DeviceCreateInfo::default().push(variable_pointers);
-        let chain2: Vec<*mut vk::BaseOutStructure<'_>> = unsafe {
-            vk::ptr_chain_iter(&mut device_create_info)
-                .skip(1)
-                .collect()
-        };
-        assert_eq!(chain, chain2);
-    }
-
     #[test]
     fn test_debug_flags() {
         assert_eq!(
diff --git a/ash/tests/fail/long_lived_root_struct_borrow.rs b/ash/tests/fail/long_lived_root_struct_borrow.rs
index a73837a..79df8d7 100644
--- a/ash/tests/fail/long_lived_root_struct_borrow.rs
+++ b/ash/tests/fail/long_lived_root_struct_borrow.rs
@@ -1,5 +1,5 @@
 use ash::vk;
-use vk::AnyTaggedStructure as _;
+use vk::TaggedStructure as _;
 
 fn main() {
     let mut layers = vec![];
diff --git a/generator/src/lib.rs b/generator/src/lib.rs
index 1f2d391..e021f8e 100644
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -2339,7 +2339,7 @@ fn derive_getters_and_setters(
         .map(|extends| {
             let base = name_to_tokens(extends);
             // Extension structs always have a pNext, and therefore always have a lifetime.
-            quote!(unsafe impl<'a> Extends<'a, #base<'a>> for #name<'a> {})
+            quote!(unsafe impl Extends<#base<'_>> for #name<'_> {})
         });
 
     let impl_structure_type_trait = structure_type_field.map(|member| {
@@ -2353,7 +2353,7 @@ fn derive_getters_and_setters(
 
         let value = variant_ident("VkStructureType", value);
         quote! {
-            unsafe impl TaggedStructure<'_> for #name<'_> {
+            unsafe impl<'a> TaggedStructure<'a> for #name<'a> {
                 const STRUCTURE_TYPE: StructureType = StructureType::#value;
             }
         }

```
</details>

Ralith:
I also experimented briefly with something like

```rs
trait TaggedStructure {
    type Abstract<'a>;
    fn narrow<'a>(self) -> Self::Abstract<'a>
        where Self: 'a;
}
```
but that would require a `fn narrow` implementation on every type, which I think undermines much of the benefit.

Ralith:
> Does it make sense to have a + ?Sized bound in the functions still?

It's harmless, but I think we should remove it for simplicity's sake. It'd be safe to restore in the future if we came up with a need.

ash/src/vk.rs
```diff
@@ -49,11 +49,83 @@ pub trait Handle: Sized {
...
+        *p_ptr = n_ptr;
+        Some(old)
+    })
+}
+
 /// Structures implementing this trait are layout-compatible with [`BaseInStructure`] and
 /// [`BaseOutStructure`]. Such structures have an `s_type` field indicating its type, which must
 /// always match the value of [`TaggedStructure::STRUCTURE_TYPE`].
-pub unsafe trait TaggedStructure {
+pub unsafe trait TaggedStructure<'a>: Sized {
```
Ralith, line 72: Is `: Sized` necessary here? I think it's more conventional to only state this explicitly if removing it would break downstream code.

Ralith (approved):

ash/src/vk.rs
MarijnS95: Not sure if it was missed, but I posed this as a question (to you) in my message above:

> The presence of `const STRUCTURE_TYPE` on `TaggedStructure` removes object safety, but Rust only complains about the `+ ?Sized` bound being useless when adding an explicit `TaggedStructure: Sized` requirement...

In my eyes these are similar. Is the linter simply not strong enough to detect this case, including:

> It says nothing about the `where Self: Sized` bound being redundant either way.

I can remove it either way.

PR #1000, opened 2025-07-18
Add helper wrappers for Vulkan core 1.4 `Device` functions

MarijnS95:


ash/src/device.rs
```diff
@@ -65,6 +69,274 @@ impl Device {
...
+            set,
+            descriptor_writes.len() as u32,
+            descriptor_writes.as_ptr(),
+        )
+    }
+
+    /// <https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdPushDescriptorSetWithTemplate.html>
+    #[inline]
+    #[doc(alias = "vkCmdPushDescriptorSetWithTemplate")]
+    pub unsafe fn cmd_push_descriptor_set_with_template(
```
Ralith, line 203: I bet there's a nice type-safe way to wrap this family of functions, though it should be built by someone who actually needs that. Not a blocker.

ash/src/device.rs
```diff
@@ -582,7 +854,7 @@ impl Device {
             count_buffer_offset,
             max_draw_count,
             stride,
-        );
```
Ralith, line 585: I don't think it's good style to omit trailing semicolons in bodies of functions that return `()`, since propagating the value of the inner expression isn't intended or useful, and it's a footgun for future refactoring efforts that might add further statements below.

Ralith (approved):

ash/src/device.rs
MarijnS95: Perhaps yeah. At least we already had this in an extension.

PR #1003, opened 2025-08-07
generator: Replace once_cell with std::sync::LazyLock

linkmauve:
Now that this feature has been integrated into std back in Rust 1.80.0, we can avoid the dependency on the `once_cell` crate.

There are many other crates depending on it in our dependency tree, but hopefully they’ll phase it out at some point.


Ralith (approved):
Cleaning up generator bitrot is always welcome.

PR #1005, opened 2025-08-12
Decouple `next` lifetime from itself in `TaggedStructure::push()` and `::extend()`

SteveIntensifies:
Right now `TaggedStructure::push()` is defined by #994 as:

```rs
fn push<T: Extends<Self> + TaggedStructure<'a>>(mut self, next: &'a mut T) -> Self
```

This requires that the extending structure `next` has the same lifetime as the contents of the base structure `self: TaggedStructure<'a>` because of `next: &'a mut T`, _and `next` itself_ because of `T: TaggedStructure<'a>` which is unnecessarily restricting. Specifically, even after `self` is no longer live (i.e. it is no longer mutably borrowing `next`) `next` cannot be accessed because it is mutably borrowed within itself through the fixed/shared lifetime of `'a` as demonstrated by a new test in commit 4942dc3. The same restriction exists for `TaggedStructure::extend()`.

This PR decouples lifetime `'a` for the contents of `self` and the borrow of `next` from a new lifetime `'b` for _the contents of `next`_ in `push()` and `extend()` to require that the _contents of `next`_ outlive the borrow of `next` rather than considering `next` to remain mutably borrowed within itself.

### DIFF
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -81,7 +81,7 @@ pub unsafe trait TaggedStructure<'a>: Sized {
     /// # Panics
     /// If `next` contains a pointer chain of its own, this function will panic.  Call `unsafe`
     /// [`Self::extend()`] to insert this chain instead.
-    fn push<T: Extends<Self> + TaggedStructure<'a>>(mut self, next: &'a mut T) -> Self {
+    fn push<'b: 'a, T: Extends<Self> + TaggedStructure<'b>>(mut self, next: &'a mut T) -> Self {
         // SAFETY: All implementers of `TaggedStructure` are required to have the `BaseOutStructure` layout
         let slf_base = unsafe { &mut *<*mut _>::cast::<BaseOutStructure<'_>>(&mut self) };
         // SAFETY: All implementers of `T: TaggedStructure` are required to have the `BaseOutStructure` layout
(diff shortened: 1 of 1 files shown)


Ralith (approved):
I can't think of any reason why this would be wrong, though I'm not super great at reasoning about variance.

SteveIntensifies:
That's a fair request. Updated the test.

> Also, welcome back ;)

I'm totally new here 😅

Ralith:
Thanks for the test. I agree that this case should work, and I confirm that it fails without this change.

There is at least one subtlety here: `x.push(&mut y)` may update `y.p_next` to point at an extension struct previously stored in `x.p_next`, which might have a shorter lifetime than `y` despite being equally or longer-lived than `x`. This could lead to `y.p_next` becoming a dangling pointer, but not before the lifetime of `x` ends.

*However*, I think that's okay: if `y` is passed to a future `push` call, then the dangling pointer is replaced, and since `y` is an extension struct it can't be passed to a Vulkan API call on its own. Code which traverses existing extension chains (e.g. `extend`) must be unsafe with an explicit invariant on their validity regardless, so I think this is still both technically sound and consistent with the practical usefulness of the safe `push` API: if you only build extension chains via `push`, then you'll never create a root struct with a dangling `p_next` in its extension chain.

tl;dr I'm reasonably confident this change is correct and desirable.

Ralith (approved):

krakow10:
As an example, I ran into this using the VideoCapabilitiesKHR, VideoDecodeCapabilitiesKHR pointer chain.  I cannot access the decode capabilities because the lifetime of `&mut video_decode_capabilities` is fixed by `video_capabilities.extend(&mut video_decode_capabilities)`.
image
See source code [here](https://github.com/ralfbiedert/vulkan_video/commit/45aad3e8b7ad4b87cbe5027381da9c27fa653a54)
This patch allows the lifetimes to be different and fixes the compilation error.

MarijnS95 (approved):

MarijnS95 (approved):
Thanks a lot, I've rebased this to make it compatible with `master` and demonstrate that your suggestion fixes the test as described, as well as extended the title/description to explain what this is fixing _and how_.

Not yet sure how to better describe that `next` is still borrowed for the lifetime of `'a` though while _the contents of `next` via `TaggedStructure<'b>`_ are now decoupled instead; will need some more tuning of the wording.

Ralith (approved):
> will need some more tuning of the wording.

Which wording, specifically? The comments in the vicinity of the change still seem correct to me.

PR #1009, opened 2025-10-11
examples: Prevent invalid semaphore reuse across frames

### LINKED ISSUE #917: ERROR: VALIDATION on examples

MarijnS95:
Fixes #917
Closes #941
Closes #979

> [!CAUTION]
> The example teaches bad practices when it comes to performance. I have a change in flight that switches them to triple-buffering which inevitably replaces the second commit. We may prefer going that route instead.

### examples: Have a unique semaphore per swapchain image

This resolves `VUID-vkQueueSubmit-pSignalSemaphores-00067` as explained at https://docs.vulkan.org/guide/latest/swapchain_semaphore_reuse.html.

A unique semaphore should be used per swapchain image, because the presentation engine may still reference and use it before the image index it was used with is returned out of `acquire_next_image()` again.

Alternatives include waiting for the `Fence` from presentation in `VkSwapchainPresentFenceInfoKHR`, added by the `VK_EXT_swapchain_maintenance1`/`VK_KHR_swapchain_maintenance1` extension but this is less performant than just using a unique semaphore per swapchain image.

### examples: Wait earlier for command-buffer fence to prevent semaphore reuse

This resolves `VUID-vkAcquireNextImageKHR-semaphore-01779` where `acquire_next_image()` is asked to signal a semaphore that may still be waited on by the currently-running submit.

Since the next submit is reusing the existing command buffer and resources as well it must inevitably wait on a fence from the previous frame/submit, pulling out that fence to wait on it is the most practical approach despite teaching bad/slow practices.

Note that this fence was also used to "reuse" the setup command buffer in the `texture` sample; this was instead generalized to use another command buffer without waiting for and signaling any fence.

### examples: Implement triple-buffering

This removes a per-frame stall and demonstrates how users should utilize Vulkan to keep the GPU fed with rendering work, as long as that work (and corresponding presentation requests) complete before we're getting 3 frames ahead.

Note that this does not implement proper frame throttling, and may render lots of discarded frames.

It is an alternative solution to solving `VUID-vkAcquireNextImageKHR-semaphore-01779` by also using a unique `Semaphore` that we know is no longer waited on via previously waiting on the relevant `Fence`; although similar in nature to the previous solution.

### DIFF
--- a/ash-examples/src/lib.rs
+++ b/ash-examples/src/lib.rs
@@ -167,7 +167,7 @@ pub struct ExampleBase {
     pub depth_image_memory: vk::DeviceMemory,
 
     pub present_complete_semaphore: vk::Semaphore,
-    pub rendering_complete_semaphore: vk::Semaphore,
+    pub rendering_complete_semaphores: Vec<vk::Semaphore>,
 
     pub draw_commands_reuse_fence: vk::Fence,
     pub setup_commands_reuse_fence: vk::Fence,
@@ -533,9 +533,13 @@ impl ExampleBase {
             let present_complete_semaphore = device
                 .create_semaphore(&semaphore_create_info, None)
                 .unwrap();
-            let rendering_complete_semaphore = device
-                .create_semaphore(&semaphore_create_info, None)
-                .unwrap();
+            let rendering_complete_semaphores = (0..present_images.len())
+                .map(|_| {
+                    device
+                        .create_semaphore(&semaphore_create_info, None)
+                        .unwrap()
+                })
+                .collect();
 
             Ok(Self {
                 event_loop: RefCell::new(event_loop),
@@ -560,7 +564,7 @@ impl ExampleBase {
                 depth_image,
                 depth_image_view,
                 present_complete_semaphore,
-                rendering_complete_semaphore,
+                rendering_complete_semaphores,
                 draw_commands_reuse_fence,
                 setup_commands_reuse_fence,
                 surface,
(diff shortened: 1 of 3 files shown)


Ralith (approved):
Triple-buffering sounds nice!

ash-examples/src/lib.rs
```diff
@@ -400,15 +411,16 @@ impl ExampleBase {
             let pool = device.create_command_pool(&pool_create_info, None).unwrap();
 
             let command_buffer_allocate_info = vk::CommandBufferAllocateInfo::default()
-                .command_buffer_count(2)
+                .command_buffer_count(5)
```
Ralith:
```suggestion
                .command_buffer_count(2 + MAX_FRAME_LATENCY)
```
For clarity.

Ralith (approved):

PR #1018, opened 2025-11-15
Generate aliases for global constants

MarijnS95:
Generating aliases will help `ash` with `Vulkan-Headers` updates without causing breaking changes when extensions and their global constants are promoted (from vendor to `EXT` to `KHR` to a Vulkan feature version) where the original constant suffix is renamed or removed while a fallback **alias** is added with the old name; which our generator was ignoring thus far.

One such example is `MAX_GLOBAL_PRIORITY_SIZE_KHR` which was stabilized in Vulkan 1.4 and renamed to `MAX_GLOBAL_PRIORITY_SIZE`; if we don't generate the old `_KHR`-suffixed constant as an alias to the stabilized constant that is technically a breaking change in `ash`'s release cycle.

By generating these aliases we can remove one unnecessary hand-written define for `SHADER_UNUSED_NV` (which seems to have been specifically added to retain backwards compatibility).


generator/src/lib.rs
```diff
@@ -84,15 +84,20 @@ impl quote::ToTokens for CType {
 }
 
 fn parse_ctype(i: &str) -> IResult<&str, CType> {
-    (alt((value(CType::U64, tag("ULL")), value(CType::U32, tag("U"))))).parse(i)
+    (alt((
+        value(CType::U64, tag("ULL")),
+        value(CType::U32, tag("U")),
+        // success(CType::U32), // Only for `0x1234` hexadecimals without type-suffix
```
MarijnS95: Was mainly "experimenting" with removing the `Number` and `Hex` variants from `enum Constant` because this code already handles both, but they'd end up weirdly formatted.

We have logic that doesn't append the type (`ctype` parsing does it by default) and interleaves the number with `_`.

EDIT: Though it wouldn't look out of order to just delete the `0x...u32` suffix.

Ralith: I don't think the formatting of generated literals is very important. I guess it'd be nice to easily eyeball against the spec, but it doesn't affect downstream.

generator/src/lib.rs
```diff
@@ -2805,20 +2817,25 @@ pub fn constant_name(name: &str) -> &str {
...
-    constant: &'a vkxml::Constant,
-    cache: &mut HashSet<&'a str>,
+    constant: &'a vk_parse::Enum,
+    global_const_cache: &mut HashMap<&'a str, CType>,
 ) -> TokenStream {
-    cache.insert(constant.name.as_str());
-    let c = Constant::from_constant(constant);
+    let (c, _, is_alias) = Constant::from_vk_parse_enum(constant, None, None).unwrap();
+    // This requires knowing the type of the alias, that is unavailable here.  Besides
+    // backwards-compatibility, stabilization aliases serve no purpose.
```
Ralith: I'm not sure what this comment is trying to express -- it reads like a value judgement of the spec, but I assume it's actually intended to say something about the behavior of the following code?

It's especially confusing that it says "[the type of the alias] is unavailable" immediately before, apparently, looking up the type of the alias.

generator/src/lib.rs
```diff
@@ -2805,20 +2817,25 @@ pub fn constant_name(name: &str) -> &str {
...
     let name = constant_name(&constant.name);
     let ident = format_ident!("{}", name);
     let notation = constant.doc_attribute();
-
-    let ty = if name == "TRUE" || name == "FALSE" {
+    let ty = if let Constant::Alias(a) = &c {
+        // TODO: Drop is_alias over Constant::Alias match
+        assert!(is_alias);
+        global_const_cache[a.to_string().as_str()]
```
Ralith: It looks like this code both populates and reads `global_const_cache`. How do we know the value we're trying to read has been populated already? Document.

Ralith (approved):
Some of these comments (both in code and on the PR) confuse me, but the improved output looks nice!

Ralith (approved):

PR #1019, opened 2025-11-21
ash-window: Keep loaded fns around via new `SurfaceFactory` object

### LINKED ISSUE #974: ash_window loses access to Surface Instance
Hi, I am new to Vulkan and was looking through the docs for `ash_window::create_instance`, the docs say that end-users should destroy the surface before destroying the parents, but this method explicitly drops the instance while orphaning the surface. How should I safely destroy a surface acquired this way?

MarijnS95:
Fixes #974

Even though `create_surface()` is typically off the hot path, we're repeatedly loading all platform-specific surface function pointers each time a new surface has to be created (sometimes for multiple windows, and on Android after every suspend-resume cycle when the app becomes invisible).  [In a subsequent patch](https://github.com/ash-rs/ash/pull/774) these loaded extensions can be used to call their respective `get_present_support()` without reloading the entire function-pointer struct too.

Additionally some users seem to be confused about object lifetimes when seeing this `Instance::new()` being dropped inside the implementation. This patch only marginally addresses that and users are more than free to just drop their `SurfaceFactory` object if no longer used (or recreate it later) without having any impact on the "parent/child relationship" mentioned in the `fn create_surface()` documentation.

For now the choice has been made to store the passed `DisplayHandle` field with the enum variant when used in `create_surface()` to reduce ambiguity by not requiring it to be passed again.  The user might still pass an incompatible `WindowHandle` in which case this returns `ERROR_EXTENSION_NOT_PRESENT`.

### DIFF
--- a/ash-window/examples/winit.rs
+++ b/ash-window/examples/winit.rs
@@ -30,6 +30,13 @@ fn main() -> Result<(), Box<dyn Error>> {
 
         let instance = entry.create_instance(&instance_desc, None)?;
 
+        let surface_factory = ash_window::SurfaceFactory::new(
+            &entry,
+            &instance,
+            event_loop.display_handle()?.as_raw(),
+        )
+        .unwrap();
+
         let window = WindowBuilder::new()
             .with_inner_size(PhysicalSize::<u32>::from((800, 600)))
             .build(&event_loop)?;
@@ -56,7 +63,8 @@ fn main() -> Result<(), Box<dyn Error>> {
             }
             Event::LoopExiting => {
                 // This will be the last event before the loop terminates.
-                // TODO: How does this play with Suspended?
+                // TODO: Winit should issue Suspended consistently on all platforms, matching
+                // Resumed and creating a single uniform place to handle surface destruction.
                 // https://github.com/rust-windowing/winit/issues/3206
                 if let Some(surface) = surface.take() {
                     surface_fn.destroy_surface(surface, None);
@@ -64,14 +72,9 @@ fn main() -> Result<(), Box<dyn Error>> {
             }
             Event::Resumed => {
                 // Create a surface from winit window.
-                let s = ash_window::create_surface(
-                    &entry,
-                    &instance,
-                    window.display_handle().unwrap().as_raw(),
-                    window.window_handle().unwrap().as_raw(),
-                    None,
-                )
-                .unwrap();
+                let s = surface_factory
+                    .create_surface(window.window_handle().unwrap().as_raw(), None)
+                    .unwrap();
                 println!("surface: {s:?}");
                 assert!(
                     surface.replace(s).is_none(),
--- a/ash-window/src/lib.rs
+++ b/ash-window/src/lib.rs
@@ -9,118 +9,178 @@ use ash::{
 };
 use raw_window_handle::{RawDisplayHandle, RawWindowHandle};
 
-/// Create a surface from a raw display and window handle.
+#[derive(Clone)]
+enum SurfaceExtension {
+    Windows(win32_surface::Instance),
+    Wayland(
+        raw_window_handle::WaylandDisplayHandle,
+        wayland_surface::Instance,
+    ),
+    Xlib(raw_window_handle::XlibDisplayHandle, xlib_surface::Instance),
+    Xcb(raw_window_handle::XcbDisplayHandle, xcb_surface::Instance),
+    Android(android_surface::Instance),
+    #[cfg(target_os = "macos")]
+    AppKit(metal_surface::Instance),
+    #[cfg(target_os = "ios")]
+    UiKit(metal_surface::Instance),
+}
+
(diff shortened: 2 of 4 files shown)


ash-window/src/lib.rs
```diff
@@ -9,118 +9,173 @@ use ash::{
...
-                Layer::Existing(layer) | Layer::Allocated(layer) => layer.cast(),
-            };
+impl SurfaceExtension {
+    /// Load the relevant surface extension for a given [`RawDisplayHandle`].
+    ///
+    // TODO: Safety docs about Entry/Instance lifetime?
+    ///
+    /// `instance` must have been created with platform specific surface extensions enabled, acquired
+    /// through [`enumerate_required_extensions()`].
+    pub fn load(
```
StarLederer: I would prefer this name to be the same as the name used for initialization in the underlying platform specific extension instances. If they don't end up getting renamed I think it is better to use `new` here

ash-window/examples/winit.rs
```diff
@@ -30,6 +30,13 @@ fn main() -> Result<(), Box<dyn Error>> {
 
         let instance = entry.create_instance(&instance_desc, None)?;
 
+        let surface_factory = ash_window::SurfaceExtension::load(
```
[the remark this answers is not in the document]
Ralith: That's a different extension, which does not overlap with those loaded here.

Ralith (approved):
Nice design!

ash-window/examples/winit.rs
```diff
@@ -56,22 +63,18 @@ fn main() -> Result<(), Box<dyn Error>> {
             }
             Event::LoopExiting => {
                 // This will be the last event before the loop terminates.
-                // TODO: How does this play with Suspended?
+                // TODO: Winit should issue Suspended consistently on all platforms, matching
+                // Resumed and creating a single uniform place to handle surface destruction.
```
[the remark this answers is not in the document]
MarijnS95: No, _I_ made the request in https://github.com/rust-windowing/winit/issues/3206 as it was simply forgotten in https://github.com/rust-windowing/winit/pull/2331. Other platforms don't have a true `Resumed` event either, yet it is consistently emitted.

At least we've renamed the events to `can_create_surfaces()` and `destroy_surfaces()`, but they're still stuck on the global `ApplicationHandler` rather than living on `WindowEvent`. This is an Android deficiency that I haven't found the motivation to fix for, for years.

---

Tying your "`Renderer` struct" to any `Surface` is likely just wrong, will be inconvenient when implementing multiple-window support, and won't work on Android where a `Surface`/`Swapchain` have a completely disjoint lifetime from the "`Renderer`".

ash-window/src/lib.rs
MarijnS95: I have just been increasingly dissatisfied with the limited number of hours that we receive with each day, which is why the **proposed** PR for renaming them all to `load()` wasn't opened yet. That change is strictly for another pull request.

---

> If we rename this away from `...Extension` as I propose elsewhere, `load` might be less clear even after the function pointer table constructors are renamed.

Would you recommend keeping it `fn new()` even after I rename the loader constructors... to `fn load()`?

Ralith: I think `load` makes sense if design the API to visibly be "some Vulkan extension", and `new` makes sense if we design the API to be a black box. I favor the latter for better encapsulation.

ash-window/src/lib.rs
```diff
@@ -9,118 +9,182 @@ use ash::{
...
+/// Also stores the platform-specific [`raw_window_handle::RawDisplayHandle`] variant if necessary
+/// to create surfaces, identifying the selected display server handle that was used to load the
+/// relevant extension when creating a [`SurfaceFactory`].
+#[derive(Clone)]
+pub struct SurfaceFactory(SurfaceExtension);
+
+impl SurfaceFactory {
+    /// Load the relevant surface extension for a given [`RawDisplayHandle`].
+    ///
+    // TODO: Safety docs about Entry/Instance lifetime?
```
Ralith: Because neither `new` nor `drop` traverse `display_handle`, I don't think we need a formal safety comment here. The safety comment on `create_surface` is sufficient.

ash-window/src/lib.rs
```diff
@@ -9,118 +9,182 @@ use ash::{
...
+    ///
+    /// `instance` must have been created with platform specific surface extensions enabled, acquired
+    /// through [`enumerate_required_extensions()`].
+    pub fn new(
+        entry: &Entry,
+        instance: &Instance,
+        // TODO: Return "lifetimed" on DisplayHandle?
```
Ralith: I like this idea, but it might give users a false sense of security if we don't also track the lifetime of the window handle, which would require bundling a lifetime with the `SurfaceKHR` somehow. I'd leave this for future work -- updating `ash-window` should be much less disruptive than updating `ash`, anyway.

Ralith (approved):

PR #1038, opened 2026-05-06
Generate deprecation docs for types like we do for commands

MarijnS95:
When handling the new `<deprecate>` element where new `<feature>`s and `<extension>`s can deprecate earlier functionality with an explanation link, these links were only forwarded to commands but not types (only defines).  Forward it so that it's also readable for the user.


MarijnS95:
On a side-note, someone mentioned to me to keep the `deprecated` attributes, but behind a `cfg_attr`. I think perhaps a `feature = "pedantic_deprecated"` could help users find where they're not using the utmost latest Vulkan features, if that's the sort of thing they're interested in and want some help with, without affecting other kinds of users?

Ralith (approved):

Ralith:
> I think perhaps a feature = "pedantic_deprecated" could help users find where they're not using the utmost latest Vulkan features, if that's the sort of thing they're interested in and want some help with, without affecting other kinds of users?

That makes sense to me! I'd personally find it useful, while we shouldn't be forcing it on everyone.

PR #1039, opened 2026-05-06
Handle extension-wide `deprecatedby`/`supersededby` attributes

MarijnS95:
After deciding to back out of marking individual commands and types as `#[deprecated]` because superseding functionality is unlikely to be widely available yet, there are however niche or vendor-specific extensions that have been replaced by more common EXT/KHR extensions or even stabilized features.  These would benefit from nudging the user towards the newer and/or cross-platform functionality which is more likely to be prevalent.

### DIFF
--- a/ash/src/vk/extensions.rs
+++ b/ash/src/vk/extensions.rs
@@ -5787,8 +5787,10 @@ pub type PFN_vkGetSwapchainGrallocUsage2ANDROID = unsafe extern "system" fn(
     gralloc_consumer_usage: *mut u64,
     gralloc_producer_usage: *mut u64,
 ) -> Result;
+#[deprecated = "Deprecated by VK_EXT_debug_utils"]
 pub const EXT_DEBUG_REPORT_NAME: &CStr =
     unsafe { CStr::from_bytes_with_nul_unchecked(b"VK_EXT_debug_report\0") };
+#[deprecated = "Deprecated by VK_EXT_debug_utils"]
 pub const EXT_DEBUG_REPORT_SPEC_VERSION: u32 = 10u32;
 #[allow(non_camel_case_types)]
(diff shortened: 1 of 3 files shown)


MarijnS95:
Semi-draft because:
- Unsure if this is enough weight for `#[deprecated]` versus a doc-comment;
- `promotedto` is not yet handled;
- Some of these attributes are empty on the `vk.xml` side, resulting in an "ugly" message here;
- Call-sites still need to be updated with `allow(deprecated)` noise in that case.

ash/src/vk/extensions.rs
```diff
@@ -5814,8 +5816,10 @@ pub type PFN_vkDebugReportMessageEXT = unsafe extern "system" fn(
     p_layer_prefix: *const c_char,
     p_message: *const c_char,
 );
+#[deprecated = "Deprecated by "]
 pub const NV_GLSL_SHADER_NAME: &CStr =
     unsafe { CStr::from_bytes_with_nul_unchecked(b"VK_NV_glsl_shader\0") };
+#[deprecated = "Deprecated by "]
```
Ralith, line 5822: Something not right here and just above.

e: oh, yeah, you noticed too

Ralith:
vulkan.hpp's use of this pattern gives me some confidence that it won't be hugely disruptive.

MarijnS95:
Alternatively this could be hidden behind the same `cfg_attr(feature = "pedantic_deprecated", deprecated = ...)` that I described in https://github.com/ash-rs/ash/pull/1038#issuecomment-4389510893.

Ralith:
I think extensions are usually replaced with other extensions (or core functionality) that's supported in the same places modulo driver updates, and references to extension names also aren't scattered broadly across a code base so it's easy for folks who really need to support old drivers to disable the lint, so a real deprecation annotation is less disruptive here than elsewhere.

PR #1041, opened 2026-05-08
extensions/khr/acceleration_structure: Make some arguments optional for `VK_KHR_opacity_micromap`

MarijnS95:
Depends on #1041

The new KHR Opacity Micromap extension leverages existing `vk::AccelerationStructureKHR` objects and related building functions (as opposed to defining bespoke types and commands in the EXT), but makes some input arguments to the existing Acceleration Structure commands optional.  These need to be adjusted in our high-level wrapper to accommodate for these requirements:

- `max_primitive_counts` in `get_acceleration_structure_build_sizes_khr()`;
- `build_range_infos` in `cmd_build_acceleration_structures()`.

Note that `build_range_infos` in `build_acceleration_structures()` is not affected, because Opacity Micromaps are limited to device-builds.

See the VUID changes in https://github.com/KhronosGroup/Vulkan-Headers/commit/0e9de566b7d4051c5cc1b762e242c46565956bdf for more details.


Jasper-Bekkers (approved):

Ralith (approved):
Another option would be to handle 0-length slices specially. I think I lean *slightly* towards the approach you've already taken, because (ignoring the assert) it probably compiles down to an unconditional load, which is cute.

PR #1051, opened 2026-07-27
Provide `const` struct constructor and builder (MSRV 1.75)

### LINKED ISSUE #1044: Const constructors and builders
Declaring a Vulkan struct in const context requires you to manually fill every member, including s_type and _marker, which looks a bit awkward. I worked this around with a free function, but it feels like unnecessary complication.

upd: maybe adding said free function in addition to constructors can be useful as well.
```
pub const fn zero_init<T : TaggedStructure>() -> T {
    unsafe {
        let mut val = ::core::mem::zeroed();
        (&raw mut val as *mut StructureType).write(T::STRUCTURE_TYPE);
        val
    }
}

```

MarijnS95:
Closes #1044 

Despite lack of `trait ~const Default` or whatever the syntax ends up being, we can already start providing `const fn const_default()` relying on `core::mem::zeroed()` and mark _most_ of the builder functions `const fn` afterwards, allowing full `const`-ized creation of Vulkan structures.

The milestone is tentative, but I think we want to buy into this MSRV-bump. There's another MSRV-bump up to `1.84` if we want the `_as_c_str()` getters to be marked `const fn` but that's almost completely useless given it's for reading back the result of queries _after Vulkan API calls_.

### (f4c2ca3) main

```console
# mount -t tmpfs -o size=5G ash_target ./target
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      8.109 s ±  0.019 s    [User: 8.312 s, System: 0.835 s]
  Range (min … max):    8.075 s …  8.146 s    10 runs
```

### (1cedf23) use `mem::zeroed()` for `Default` (and generic `TaggedStructure` `s_type` init)

```console
  Time (mean ± σ):      7.931 s ±  0.022 s    [User: 8.122 s, System: 0.837 s]
  Range (min … max):    7.892 s …  7.973 s    10 runs
```

### (3af6952) use free-function for initializing `TaggedStructure`

```console
  Time (mean ± σ):      7.881 s ±  0.031 s    [User: 8.090 s, System: 0.827 s]
  Range (min … max):    7.840 s …  7.948 s    10 runs
```

### (16476ae) Manual `derive(Default)` for the remainder

```console
  Time (mean ± σ):      7.884 s ±  0.032 s    [User: 8.088 s, System: 0.834 s]
  Range (min … max):    7.838 s …  7.933 s    10 runs
```

### (14eb203) Add `const_default()` and relay `impl Default` through it

```console
  Time (mean ± σ):      8.135 s ±  0.024 s    [User: 8.313 s, System: 0.859 s]
  Range (min … max):    8.104 s …  8.170 s    10 runs
```

### (5e4746e) Mark most builder functions `const`

```console
  Time (mean ± σ):      8.140 s ±  0.045 s    [User: 8.322 s, System: 0.865 s]
  Range (min … max):    8.016 s …  8.169 s    10 runs
```

### DIFF
--- a/ash/src/vk.rs
+++ b/ash/src/vk.rs
@@ -129,6 +129,18 @@ pub unsafe trait TaggedStructure<'a>: Sized {
         slf_base.p_next = next_base;
         self
     }
+
+    /// `const`-initializes a tagged structure with all-zeroes, filling only its
+    /// [`BaseOutStructure::s_type`] field.
+    ///
+    /// All Vulkan structures support zero-initialization.
+    fn init_tagged_structure() -> Self {
+        let mut slf = unsafe { core::mem::zeroed() };
+        // SAFETY: All implementers of `TaggedStructure` are required to have the `BaseOutStructure` layout
+        let slf_base = unsafe { &mut *<*mut _>::cast::<BaseOutStructure<'_>>(&mut slf) };
+        slf_base.s_type = Self::STRUCTURE_TYPE;
+        slf
+    }
 }
 
 /// Implemented for every structure that extends base structure `B`. Concretely that means struct
--- a/generator/src/lib.rs
+++ b/generator/src/lib.rs
@@ -1886,24 +1886,15 @@ fn derive_default(
     has_lifetime: bool,
     provisional: &Option<TokenStream>,
 ) -> Option<TokenStream> {
+    if struct_.name == "VkBaseInStructure" || struct_.name == "VkBaseOutStructure" {
+        // TODO: These don't really have a valid initializer, since they'll
+        // end up as vk::StructureType::APPLICATION_INFO (0).
+        // TODO: Returning None just triggers derive(Default)
+        return None;
+    }
     let name = name_to_tokens(&struct_.name);
     let is_structure_type = |field: &vkxml::Field| field.basetype == "VkStructureType";
 
-    // These are also pointers, and therefor also don't implement Default. The spec
-    // also doesn't mark them as pointers
-    let handles = [
-        "LPCWSTR",
-        "HANDLE",
-        "HINSTANCE",
-        "HWND",
-        "HMONITOR",
-        "IOSurfaceRef",
-        "MTLBuffer_id",
-        "MTLCommandQueue_id",
-        "MTLDevice_id",
-        "MTLSharedEvent_id",
-        "MTLTexture_id",
-    ];
     let contains_ptr = members
         .iter()
         .any(|member| member.vkxml_field.reference.is_some());
@@ -1913,48 +1904,26 @@ fn derive_default(
     let contains_static_array = members
         .iter()
         .any(|member| is_static_array(member.vkxml_field));
-    let contains_deprecated = members.iter().any(|member| member.deprecated.is_some());
-    let allow_deprecated = contains_deprecated.then(|| quote!(#[allow(deprecated)]));
     if !(contains_ptr || contains_structure_type || contains_static_array) {
         return None;
     };
-    let default_fields = members.iter().map(|member| {
-        let param_ident = member.vkxml_field.param_ident();
-        if is_structure_type(member.vkxml_field) {
-            if member.vkxml_field.type_enums.is_some() {
-                quote!(#param_ident: Self::STRUCTURE_TYPE)
-            } else {
-                quote!(#param_ident: unsafe { ::core::mem::zeroed() })
-            }
-        } else if member.vkxml_field.reference.is_some() {
-            if member.vkxml_field.is_const {
-                quote!(#param_ident: ::core::ptr::null())
-            } else {
-                quote!(#param_ident: ::core::ptr::null_mut())
-            }
-        } else if is_static_array(member.vkxml_field)
-            || handles.contains(&member.vkxml_field.basetype.as_str())
-        {
-            quote!(#param_ident: unsafe { ::core::mem::zeroed() })
-        } else {
-            let ty = member.vkxml_field.type_tokens(false, None);
-            quote!(#param_ident: #ty::default())
-        }
-    });
     let lifetime = has_lifetime.then(|| quote!(<'_>));
-    let marker = has_lifetime.then(|| quote!(_marker: PhantomData,));
+
+    let initializer = if contains_structure_type {
+        quote!(Self::init_tagged_structure())
+    } else {
+        quote! {
+            // SAFETY: All Vulkan structures support zero-initialization
+            unsafe { ::core::mem::zeroed() }
+        }
+    };
+
     Some(quote! {
         #provisional
         impl ::core::default::Default for #name #lifetime {
             #[inline]
             fn default() -> Self {
-                #allow_deprecated
-                Self {
-                    #(
-                        #default_fields,
-                    )*
-                    #marker
-                }
+                #initializer
             }
         }
     })
--- a/ash/src/vk/definitions.rs
+++ b/ash/src/vk/definitions.rs
@@ -696,7 +696,7 @@ pub type PFN_vkGetInstanceProcAddrLUNARG = Option<
 >;
 #[repr(C)]
 #[cfg_attr(feature = "debug", derive(Debug))]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Default)]
 #[doc = "<https://docs.vulkan.org/refpages/latest/refpages/source/VkBaseOutStructure.html>"]
 #[must_use]
 pub struct BaseOutStructure<'a> {
@@ -706,19 +706,9 @@ pub struct BaseOutStructure<'a> {
 }
 unsafe impl Send for BaseOutStructure<'_> {}
 unsafe impl Sync for BaseOutStructure<'_> {}
-impl ::core::default::Default for BaseOutStructure<'_> {
-    #[inline]
-    fn default() -> Self {
-        Self {
-            s_type: unsafe { ::core::mem::zeroed() },
-            p_next: ::core::ptr::null_mut(),
-            _marker: PhantomData,
-        }
-    }
-}
 #[repr(C)]
 #[cfg_attr(feature = "debug", derive(Debug))]
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Default)]
 #[doc = "<https://docs.vulkan.org/refpages/latest/refpages/source/VkBaseInStructure.html>"]
 #[must_use]
 pub struct BaseInStructure<'a> {
@@ -728,16 +718,6 @@ pub struct BaseInStructure<'a> {
 }
 unsafe impl Send for BaseInStructure<'_> {}
 unsafe impl Sync for BaseInStructure<'_> {}
-impl ::core::default::Default for BaseInStructure<'_> {
-    #[inline]
-    fn default() -> Self {
-        Self {
-            s_type: unsafe { ::core::mem::zeroed() },
-            p_next: ::core::ptr::null(),
-            _marker: PhantomData,
-        }
-    }
-}
 #[repr(C)]
 #[cfg_attr(feature = "debug", derive(Debug))]
 #[derive(Copy, Clone, Default, PartialEq, Eq, Hash)]
@@ -994,17 +974,7 @@ impl fmt::Debug for PhysicalDeviceProperties {
 impl ::core::default::Default for PhysicalDeviceProperties {
     #[inline]
     fn default() -> Self {
-        Self {
-            api_version: u32::default(),
-            driver_version: u32::default(),
-            vendor_id: u32::default(),
-            device_id: u32::default(),
-            device_type: PhysicalDeviceType::default(),
-            device_name: unsafe { ::core::mem::zeroed() },
-            pipeline_cache_uuid: unsafe { ::core::mem::zeroed() },
-            limits: PhysicalDeviceLimits::default(),
-            sparse_properties: PhysicalDeviceSparseProperties::default(),
-        }
+        unsafe { ::core::mem::zeroed() }
     }
 }
 impl PhysicalDeviceProperties {
@@ -1080,10 +1050,7 @@ impl fmt::Debug for ExtensionProperties {
 impl ::core::default::Default for ExtensionProperties {
     #[inline]
     fn default() -> Self {
-        Self {
-            extension_name: unsafe { ::core::mem::zeroed() },
-            spec_version: u32::default(),
-        }
+        unsafe { ::core::mem::zeroed() }
     }
 }
 impl ExtensionProperties {
@@ -1128,12 +1095,7 @@ impl fmt::Debug for LayerProperties {
 impl ::core::default::Default for LayerProperties {
     #[inline]
     fn default() -> Self {
-        Self {
-            layer_name: unsafe { ::core::mem::zeroed() },
-            spec_version: u32::default(),
-            implementation_version: u32::default(),
-            description: unsafe { ::core::mem::zeroed() },
-        }
+        unsafe { ::core::mem::zeroed() }
     }
 }
 impl LayerProperties {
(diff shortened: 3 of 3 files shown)


generator/src/lib.rs
```diff
@@ -1886,75 +1886,41 @@ fn derive_default(
     has_lifetime: bool,
     provisional: &Option<TokenStream>,
 ) -> Option<TokenStream> {
+    if struct_.name == "VkBaseInStructure" || struct_.name == "VkBaseOutStructure" {
+        // These don't really have a valid initializer, since they'll end up as
+        // vk::StructureType::APPLICATION_INFO (0).
+        return None;
+    }
```
MarijnS95, line 1893: Together with another change, this'll drop the "accidental" `Default for Base*Structure` which would be pretty useless/invalid to begin with?

generator/src/lib.rs
```diff
@@ -1886,75 +1886,41 @@ fn derive_default(
...
             #[inline]
             fn default() -> Self {
-                #allow_deprecated
-                Self {
-                    #(
-                        #default_fields,
-                    )*
-                    #marker
-                }
+                Self::const_default()
```
MarijnS95, line 1923: Unsure if this is really worth it, if `derive(Default)` is also available.

generator/src/lib.rs
```diff
@@ -2699,16 +2670,11 @@ pub fn generate_struct(
...
-        quote!(Default,)
-    } else {
-        quote!()
-    };
     quote! {
         #provisional
         #[repr(C)]
         #derive_debug
-        #[derive(Copy, Clone, #derive_default #manual_derive_tokens)]
+        #[derive(Copy, Clone, #manual_derive_tokens)]
```
MarijnS95, line 2677: Followup on the comments above: this could stay as long as they don't get applied to `Base*Structure`, and if it's faster/more convenient than forcing everything through `Self::const_default()`?

generator/src/lib.rs
```diff
@@ -1886,75 +1886,41 @@ fn derive_default(
...
+            pub const fn const_default() -> Self {
+                #initializer
             }
-        } else if is_static_array(member.vkxml_field)
-            || handles.contains(&member.vkxml_field.basetype.as_str())
-        {
-            quote!(#param_ident: unsafe { ::core::mem::zeroed() })
-        } else {
-            let ty = member.vkxml_field.type_tokens(false, None);
-            quote!(#param_ident: #ty::default())
```
MarijnS95, line 1941: The most contentious of the bunch: were/are we using any types currently where `mem::zeroed() != Default::default()`? Need to emit a code `#[test]` to keep that in check.

(That'd falsify my claim that all `Vulkan` structures are safely zero-initializable in the doc-comment for at least `TaggedStructure`s above)

ash/src/vk.rs
```diff
@@ -131,6 +131,18 @@ pub unsafe trait TaggedStructure<'a>: Sized {
     }
 }
 
+/// `const`-initializes a tagged structure with all-zeroes, filling only its
+/// [`BaseOutStructure::s_type`] field.
+///
+/// All Vulkan structures support zero-initialization.
```
MarijnS95, line 137: Also, be wary of nested `TaggedStructure`s owned by one another? Those might slip in and would previously get resolved via the `Default::default()` call; but would be caught by an autogenerated test at least.

ash/src/vk.rs
```diff
@@ -129,6 +129,18 @@ pub unsafe trait TaggedStructure<'a>: Sized {
         slf_base.p_next = next_base;
         self
     }
+
+    /// `const`-initializes a tagged structure with all-zeroes, filling only its
+    /// [`BaseOutStructure::s_type`] field.
+    ///
+    /// All Vulkan structures support zero-initialization.
+    fn init_tagged_structure() -> Self {
```
Ralith, line 137: nit: `new()`? `_tagged_structure` is redundant to the trait name, and we shouldn't usually name a constructor with a verb.

generator/src/lib.rs
Ralith: SGTM.

generator/src/lib.rs
Ralith: Even if we were, I think zero-initialization more closely matches Vulkan's design intent. Would be bad to silently change the default out from under folks though.

ash/src/vk.rs
Ralith: Mooted in the next commit, which I like better anyway.

generator/src/lib.rs
```diff
@@ -1909,11 +1909,18 @@ fn derive_default(
     };
 
     Some(quote! {
+        #provisional
+        impl #name #lifetime {
+            #[inline]
+            pub const fn const_default() -> Self {
```
Ralith, line 1915: Name this `new`? Then our API is more conventional, and it won't feel awkward and redundant when if/when true const `Default` manifests.

generator/src/lib.rs
Ralith: The derive seems a little neater, but global consistency is also nice. No strong feeling.

generator/src/lib.rs
Ralith: Is the compile-time cost due to delegating through `const_default`, or due to defining `const_default` at all? I lean slightly towards consistency if it's a wash, but if deriving is actually faster then I favor it.

ash/src/vk.rs
Ralith: I'd be very surprised if one was introduced. Assert against it, and treat it as a special case if that happens? We could introduce a visitor pattern to handle it down the line if we really have to.

Ralith (commented):
Mostly a fan of these changes. I'm very skeptical that `const_default` is worth having if it costs us a measurable amount of compile time, though. Has anyone presented a credible real-world use case?

generator/src/lib.rs
MarijnS95: Having a single source of truth is my preference, but it becomes moot when we insert the `Default == zeroed()` tests.

generator/src/lib.rs
MarijnS95: It appears to be due to defining `const_default` at all. The cost we got back by not manually deriving all fields was "paid back" by having `fn const_default()`.

Unfortunately my machine at home is a bit faster with the 3970X being on water as opposed to air, and the checkout being on a better NVMe drive. So, new numbers really quickly:

### (f4c2ca3) main

```console
# mount -t tmpfs -o size=5G ash_target ./target
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      7.927 s ±  0.096 s    [User: 8.168 s, System: 0.862 s]
  Range (min … max):    7.735 s …  8.081 s    10 runs
```

### (3b6e7b7) Cherry-picking the addition of `const_default()` to main

```
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      8.309 s ±  0.053 s    [User: 8.537 s, System: 0.875 s]
  Range (min … max):    8.215 s …  8.409 s    10 runs
```

### (5e4746e) Back on this PR

```
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      8.041 s ±  0.073 s    [User: 8.259 s, System: 0.883 s]
  Range (min … max):    7.955 s …  8.170 s    10 runs
```

Remains to be seen how much we actually pay when all bugs are fixed.

generator/src/lib.rs
```diff
@@ -1886,75 +1886,41 @@ fn derive_default(
...
-        .iter()
-        .any(|member| is_static_array(member.vkxml_field));
-    let contains_deprecated = members.iter().any(|member| member.deprecated.is_some());
-    let allow_deprecated = contains_deprecated.then(|| quote!(#[allow(deprecated)]));
-    if !(contains_ptr || contains_structure_type || contains_static_array) {
-        return None;
+    let lifetime = has_lifetime.then(|| quote!(<'_>));
+
+    let initializer = if contains_structure_type {
+        quote!(init_tagged_structure())
```
MarijnS95, line 1903: FYI changing this to FRU actually ended up being significantly slower:

```suggestion
        quote! {
            Self {
                s_type: Self::STRUCTURE_TYPE,
                // SAFETY: All Vulkan structures support zero-initialization
                ..unsafe { ::core::mem::zeroed() }
            }
        }
```
  

```
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      8.041 s ±  0.073 s    [User: 8.259 s, System: 0.883 s]
  Range (min … max):    7.955 s …  8.170 s    10 runs
```
->
```
Benchmark 1: cargo b -p ash
  Time (mean ± σ):      8.175 s ±  0.051 s    [User: 8.394 s, System: 0.883 s]
  Range (min … max):    8.114 s …  8.255 s    10 runs
```

ash/src/vk.rs
MarijnS95: Yeah, and it couldn't be made `const` in the `trait` anyway (yet).

generator/src/lib.rs
MarijnS95: There are quite a few structures that **contain `s_type` structures**:

```
failures:
    vk::definitions::eq_VkAccelerationStructureCreateInfoNV
    vk::definitions::eq_VkAttachmentSampleLocationsEXT
    vk::definitions::eq_VkBufferCollectionPropertiesFUCHSIA
    vk::definitions::eq_VkBufferConstraintsInfoFUCHSIA
    vk::definitions::eq_VkBuildPartitionedAccelerationStructureInfoNV
    vk::definitions::eq_VkClusterAccelerationStructureCommandsInfoNV
    vk::definitions::eq_VkComputePipelineCreateInfo
    vk::definitions::eq_VkDisplayPlanePropertiesKHR
    vk::definitions::eq_VkGeometryDataNV
    vk::definitions::eq_VkGeometryNV
    vk::definitions::eq_VkImageConstraintsInfoFUCHSIA
    vk::definitions::eq_VkImageFormatConstraintsInfoFUCHSIA
    vk::definitions::eq_VkMemoryHeap
    vk::definitions::eq_VkPhysicalDeviceLayeredApiVulkanPropertiesKHR
    vk::definitions::eq_VkPipelineSampleLocationsStateCreateInfoEXT
    vk::definitions::eq_VkSubpassSampleLocationsEXT
    vk::definitions::eq_VkVideoDecodeInfoKHR
    vk::definitions::eq_VkVideoEncodeInfoKHR
```

I think we should just build a list similar to `has_lifetimes: &HashSet<Ident>` for `has_struct_type`, and switch to FRU for those fields.

However, that means the public `init_tagged_structure()` added here may need to stop existing (or be made unsafe / not public) if there are `s_type` structs that contain other `s_type` structs. Then the caller should just redirect through `::new()`.

---

Note that my test was based on your earlier question about compile-time perf. I used that cherry-pick to compare `::default()` against `::const_default()`. But because of padding bytes, I opted to replace `::default()` with a `let mut slf = zeroed::<Self>(); #(slf.#field = #initializer); slf` sort of body.

For `eq_VkMemoryHeap` it seems that that didn't initialize the trailing padding bytes, causing its failure.

MarijnS95: The only curious interim case currently is `AccelerationStructureGeometryKHR`, which owns a `geometry_type` and `geometry` which is defined as `union AccelerationStructureGeometryKHR`: the three fields all have an `s_type`. I guess I'm treating this one as "pointer" since `Default::default()` also left the entire `union` at zeroed.

MarijnS95: Pending fixes and tests for this live in https://github.com/ash-rs/ash/compare/test-const-builder-eq. These curious stragglers remain:

```
---- vk::definitions::eq_VkComputePipelineCreateInfo stdout ----

thread 'vk::definitions::eq_VkComputePipelineCreateInfo' (1210561) panicked at ash/src/vk.rs:454:9:
assertion `left == right` failed
  left: [29, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 18, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 right: [29, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 41, 127, 0, 0, 18, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- vk::definitions::eq_VkDisplayPlanePropertiesKHR stdout ----

thread 'vk::definitions::eq_VkDisplayPlanePropertiesKHR' (1210728) panicked at ash/src/vk.rs:454:9:
assertion `left == right` failed
  left: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 242, 85, 0, 0]
 right: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

---- vk::definitions::eq_VkImageConstraintsInfoFUCHSIA stdout ----

thread 'vk::definitions::eq_VkImageConstraintsInfoFUCHSIA' (1210833) panicked at ash/src/vk.rs:454:9:
assertion `left == right` failed
  left: [182, 95, 160, 59, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 185, 95, 160, 59, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 right: [182, 95, 160, 59, 242, 85, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 185, 95, 160, 59, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

---- vk::definitions::eq_VkMemoryHeap stdout ----

thread 'vk::definitions::eq_VkMemoryHeap' (1210931) panicked at ash/src/vk.rs:454:9:
assertion `left == right` failed
  left: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 242, 85, 0, 0]
 right: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

---- vk::definitions::eq_VkVideoEncodeInfoKHR stdout ----

thread 'vk::definitions::eq_VkVideoEncodeInfoKHR' (1211778) panicked at ash/src/vk.rs:454:9:
assertion `left == right` failed
  left: [248, 89, 159, 59, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 218, 35, 155, 59, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 right: [248, 89, 159, 59, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 218, 35, 155, 59, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
```

